PR #58, opened 2021-09-16
fix(cacheing): Include namespace in formatting cache key

### LINKED ISSUE #57: Does not re-render when namespace changes
**Describe the bug**
Internationalized messages are not updated when the namespace provided in `useTranslations` changes.

**To Reproduce**

https://codesandbox.io/s/nice-shadow-ksq1z?file=/src/App.js

Klick 'NamespaceA' and 'NamespaceB' buttons.

**Expected behavior**
The title-message should be updated accordingly.

mismosmi:
Fixes #57

### DIFF
--- a/packages/use-intl/src/useTranslations.tsx
+++ b/packages/use-intl/src/useTranslations.tsx
@@ -154,9 +154,11 @@ export default function useTranslations(namespace?: string) {
       }
       const messages = messagesOrError;
 
+      const cacheKey = `${namespace}.${key}`
+
       let messageFormat;
-      if (cachedFormatsByLocale[locale]?.[key]) {
-        messageFormat = cachedFormatsByLocale[locale][key];
+      if (cachedFormatsByLocale[locale]?.[cacheKey]) {
+        messageFormat = cachedFormatsByLocale[locale][cacheKey];
       } else {
         let message;
         try {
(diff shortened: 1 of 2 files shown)


packages/use-intl/src/useTranslations.tsx
```diff
@@ -154,9 +154,11 @@ export default function useTranslations(namespace?: string) {
       }
       const messages = messagesOrError;
 
+      const cacheKey = `${namespace}.${key}`
```
amannn, line 157: The `namespace` is optional. Can you use `[namespace, key].filter(part => part != null).join('.')` here?

packages/use-intl/test/useTranslations.test.tsx
```diff
@@ -234,6 +234,35 @@ it('has a stable reference', () => {
   screen.getByText('2');
 });
 
+it('caches by locale, namespace and key', () => {
```
amannn, line 237: I think the caching should be a implementation detail here.

```suggestion
it('renders the correct message when the namespace changes', () => {
```

amannn (commented):
Thanks a lot for looking into this @mismosmi!

Your solution looks good to me, I'd just handle the case where we don't have a namespace a bit more explicit, so `undefined` doesn't end up in the cache key.

Can you also apply the suggestions from ESLint? Most of them should be auto-fixable.

mismosmi:
I applied all your suggestions.
Sorry for eslint, usually vscode bugs me with these things, I don't know what went wrong.

amannn (approved):
Thanks a lot, looks good to me!

PR #73, opened 2021-12-22
docs: Create docs site

### LINKED ISSUE #72: Documentation improvements
- [x] Make sure all used functions are imported in the code samples (focus on copy-pasting)
- [x] Add a table of contents for quick scanning
- [x] A standalone website
- [x] Potentially a separate example that shows just the basic hello world setup and another one showing more advanced features
- [x] Live examples
- [x] Clarify which polyfills are needed and under which conditions
- [x] Seems like use `new Date(dateString)` [is fine now](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date) in modern browsers, the docs should relfect that
- [x] Link to / use https://www.intl-explorer.com/DateTimeFormat?locale=en-GB?
- [x] Use TypeScript everywhere
- [x] Clarify what `locale` is
- [x] Upgrade code sandboxes.
- [ ] Inline examples with sandpack?
- [ ] Consider more examples or linking more to examples from the docs.
- [x] Avoid external links if not really necessary

amannn:
Contributes to #72


docs/usage.md
```diff
@@ -240,6 +247,7 @@ Similar to number formatting, you can format plain dates and times that are not
 
 ```js
 import {useIntl} from 'next-intl';
+import {parseISO} from 'data-fns';
```
andrevenancio, line 250: Do you mean `date-fns` ?

amannn: Yep 🤦. Fixed now!

Btw. I'm currently migrating the docs to a standalone website. There are no major changes yet, but I think it's already easier to browse in comparison to the markdown files. So if you're interested in an early preview: https://next-intl-amann.vercel.app/.

andrevenancio: YES! and I think you'll be able to include interactive snippets there right? via codesandbox or something. So it will be even better for users being able to read and play and see it in actions. I'm using radix-ui on my current project and their documentation website is really good if you're after a reference. https://www.radix-ui.com/ (this is their "sales pitch" site, but then you can jump straight into the documentation. Here's an example of their documentation with interactive examples https://www.radix-ui.com/docs/primitives/components/dialog

Its really nice to be able to test something without having to setup a project, install dependencies etc.

amannn: Haha, sounds promising 😁. Yes, I definitely remember when you wrote me on Twitter! I'll see if I can get the docs shaped up a bit and I'll let you know when I've got something to show! Thank you for your great inputs! 🙂

amannn:
A first version of the standalone website is now available at https://next-intl-docs.vercel.app. I'll look into adding and improving examples in a follow-up.

PR #86, opened 2022-02-05
feat: Default translation values

maxwoedl:
To achieve consistent usage and styling of rich text elements, I have included an option to define global defaults for rich text elements. The default will be overridden by a local definition of an element.

This approach is the same as the one from [react-intl](https://formatjs.io/docs/react-intl/components/#defaultrichtextelements).

### DIFF
--- a/packages/use-intl/src/useTranslations.tsx
+++ b/packages/use-intl/src/useTranslations.tsx
@@ -85,7 +85,7 @@ function prepareTranslationValues(values: RichTranslationValues) {
  */
 export default function useTranslations(namespace?: string) {
   const {
-    defaultRichTextElements,
+    defaultTranslationValues,
     formats: globalFormats,
     getMessageFallback,
     locale,
@@ -216,7 +216,7 @@ export default function useTranslations(namespace?: string) {
 
       try {
         const formattedMessage = messageFormat.format(
-          prepareTranslationValues({...defaultRichTextElements, ...values})
+          prepareTranslationValues({...defaultTranslationValues, ...values})
         );
 
         if (formattedMessage == null) {
@@ -306,7 +306,7 @@ export default function useTranslations(namespace?: string) {
     namespace,
     onError,
     timeZone,
-    defaultRichTextElements
+    defaultTranslationValues
   ]);
 
   return translate;
--- a/packages/use-intl/src/useTranslations.tsx
+++ b/packages/use-intl/src/useTranslations.tsx
@@ -48,7 +48,7 @@ function resolvePath(
 }
 
 function prepareTranslationValues(values: RichTranslationValues) {
-  if (Object.keys(values).length < 1) return undefined;
+  if (Object.keys(values).length === 0) return undefined;
 
   // Workaround for https://github.com/formatjs/formatjs/issues/1467
   const transformedValues: RichTranslationValues = {};
@@ -85,13 +85,13 @@ function prepareTranslationValues(values: RichTranslationValues) {
  */
 export default function useTranslations(namespace?: string) {
   const {
+    defaultRichTextElements,
     formats: globalFormats,
     getMessageFallback,
     locale,
     messages: allMessages,
     onError,
-    timeZone,
-    defaultRichTextElements
+    timeZone
   } = useIntlContext();
 
   const cachedFormatsByLocaleRef = useRef<
--- a/packages/use-intl/src/useTranslations.tsx
+++ b/packages/use-intl/src/useTranslations.tsx
@@ -47,8 +47,8 @@ function resolvePath(
   return message;
 }
 
-function prepareTranslationValues(values?: RichTranslationValues) {
-  if (!values) return values;
+function prepareTranslationValues(values: RichTranslationValues) {
+  if (Object.keys(values).length < 1) return undefined;
 
   // Workaround for https://github.com/formatjs/formatjs/issues/1467
   const transformedValues: RichTranslationValues = {};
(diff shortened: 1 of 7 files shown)


maxwoedl:
BTW is one of use-intl's tests broken due to a spelling mistake (plural and singular), I can also create a PR to fix it if you like.

packages/website/pages/docs/usage/configuration.mdx
```diff
@@ -71,6 +71,21 @@ t('ordered', {orderDate: parseISO('2020-11-20T10:36:01.516Z')});
...
 
+## Global Rich Text Elements
+To achieve consistent usage and styling of rich text elements, you can define a set of global elements and pass them to the provider.
+If you define an element locally it will override the global default.
+
+```js
+<NextIntlProvider
+  ...
+  defaultRichTextElements={{
+    b: (value) => <strong>{value}</strong>
```
amannn, line 82: I'm calling the argument "children" in [the rich text section](https://next-intl-git-fork-maxwoedl-main-amann.vercel.app/docs/usage/messages#rich-text). Can you adapt the wording here and possibly add brackets around the argument in the rich text section for consistency? Your version with brackets is better.

Can you also cross reference this section from the rich text docs?

packages/use-intl/src/useTranslations.tsx
```diff
@@ -215,7 +216,7 @@ export default function useTranslations(namespace?: string) {
 
       try {
         const formattedMessage = messageFormat.format(
-          prepareTranslationValues(values)
+          prepareTranslationValues({...defaultRichTextElements, ...values})
```
amannn, line 219: Can you please add some tests that cover this feature?

packages/website/pages/docs/usage/configuration.mdx
```diff
@@ -71,6 +71,21 @@ t('ordered', {orderDate: parseISO('2020-11-20T10:36:01.516Z')});
...
+  }}
+>
+  <App />
+</NextIntlProvider>
+```
+
```
amannn, line 88: > BTW is one of use-intl's tests broken due to a spelling mistake (plural and singular), I can also create a PR to fix it if you like.

Which is the test you're seeing problems with? Seems like CI passes. I'd be curious if it's a problem related to the environment. You might need node>=14, this is currently not enforced.

packages/website/pages/docs/usage/configuration.mdx
```diff
@@ -71,6 +71,21 @@ t('ordered', {orderDate: parseISO('2020-11-20T10:36:01.516Z')});
 t('latitude', {latitude: 47.414329182});
 ```
 
+## Global Rich Text Elements
+To achieve consistent usage and styling of rich text elements, you can define a set of global elements and pass them to the provider.
+If you define an element locally it will override the global default.
```
amannn, line 76: If we go for `defaultTranslationValues` the docs need to reflect that.

amannn (commented):
Thanks a lot for looking into this @maxwoedl! I hope you found your way around in the code base, your change is really well done! 👏 

I've added some minor comments, but I think we can get this in.

packages/use-intl/test/useIntl.test.tsx
```diff
@@ -272,7 +272,7 @@ describe('formatNumber', () => {
 
       const error: IntlError = onError.mock.calls[0][0];
       expect(error.message).toBe(
-        'FORMATTING_ERROR: Invalid currency codes : unknown'
+        'FORMATTING_ERROR: Invalid currency code : unknown'
```
amannn, line 275: Hmm, locally this is indeed "codes" for me with node v14.19.0. Seems like on CI there's the same result. Which node version are you using where you're seeing "code"?

packages/website/pages/docs/usage/configuration.mdx
maxwoedl: Interestingly enough now the CI pipeline failed... locally it failed the other way around. Can you maybe somehow undo/exclude the [34d9c05](https://github.com/amannn/next-intl/pull/86/commits/34d9c05817bbfdcccf66d3add66fc0828b2d715e) commit.

packages/website/pages/docs/usage/messages.mdx
```diff
@@ -149,6 +149,7 @@ t('selectordinal', {year: 11});
 ## Rich text
 
 You can format rich text with custom tags and map them to React components.
+If you want to use the same tag multiple times, you can configure it via the [default translation values](/docs/usage/configuration#default-translation-values).
```
amannn, line 152: Nit: Can we move this after the code snippet? I feel like it could be helpful to first finish introducing the concept and then providing supplemental information like the global config. It might also help a bit with discovery, I tend to skip directly to the code examples when reading a section and expect advanced information to follow afterwards.

packages/website/pages/docs/usage/configuration.mdx
amannn: I think if you run `git revert 34d9c05` in your branch that should do the trick.

packages/use-intl/test/useIntl.test.tsx
maxwoedl: I am using node v16.13.1, so you were correct about the node version, sorry for "fixing" the test, i will undo it :)

amannn: No problem! An upgrade to node v16 would definitely be interesting, but I'm not sure yet if there are other CI-related things to update then so that can be done at a later point for sure.

amannn (approved):
Really great PR, thank you so much @maxwoedl! 🙌

maxwoedl:
> Really great PR, thank you so much @maxwoedl! 🙌

Was a pleasure to work with you @amannn! ☺️

Issue #102, opened 2022-04-13
setup for tests not working - missing message


BernardA:
**Is your feature request related to a problem? Please describe.**
When setting up tests I am getting error:

> MISSING_MESSAGE: No messages were configured on the provider.

Tests are using `react testing library`.

**Describe the solution you'd like**
Working test set up.

**Describe alternatives you've considered**
I have tried solution based upon the[ `testing library` documentation](https://testing-library.com/docs/example-react-intl/) and also [this issue](https://testing-library.com/docs/example-react-intl/).

Like so:

/tests/utils/custom.tsx
      
       import React from 'react';
       import { render } from '@testing-library/react';
       import { NextIntlProvider } from 'next-intl';

     const customRender = (ui: React.ReactElement, options) => {
         return render(ui, {
             wrapper: (props: any) => (
                 <NextIntlProvider
                     locale="en"
                     {...props}
                     {...options?.NextIntlProviderProps}
                 />
             ),
             ...options?.testingLibraryOptions,
         });
     };

     // re-export everything
     export * from '@testing-library/react';

     // override render method
     export { customRender };

Then, on the actual test page:

/tests/pages/login.tsx
 
       import '@testing-library/jest-dom';
       import Login from 'pages/login';
       import { customRender } from '../utils/custom';

      describe('Login render Page', () => {
         it('renders the Login page', () => {
             const { getByText } = customRender(<Login />);
              expect(getByText(/login/i)).toBeInTheDocument();
         });
       ......

So, with above it is somehow unable to find the messages.

I am almost sure this is not a `next-intl` bug but rather that something is amiss on my setup. 

Note that translations **work fine on the actual component**. This is a test environment issue.

**Additional context**
The error is thrown on the `useTranslations()` hook on the component page as below:

<img width="930" alt="Screenshot 2022-04-13 at 16 44 24" src="https://user-images.githubusercontent.com/9449665/163206546-996152c1-8f88-4cd5-acb3-42790db52980.png">

amannn:
Hi @BernardA,

from what I can tell you didn't provide any messages on the provider. That's what the error message is trying to describe.

There's [a short section about Jest in the docs](https://next-intl-docs.vercel.app/docs/faq#can-next-intl-be-used-in-environments-like-jest-and-storybook). If you don't want to provide sample messages, you can also [configure the error handling](https://next-intl-docs.vercel.app/docs/usage/error-handling) to ignore errors for missing messages.

[...]

PR #113, opened 2022-05-20
fix: Add back ability to render without messages

amannn:
Relates to #102


packages/use-intl/src/useTranslations.tsx
```diff
@@ -103,16 +102,7 @@ export default function useTranslations<
...
-
   const messages = context.messages as IntlMessages;
-  if (!messages) {
-    const intlError = new IntlError(
-      IntlErrorCode.MISSING_MESSAGE,
-      __DEV__ ? `No messages were configured on the provider.` : undefined
-    );
-    context.onError(intlError);
-    throw intlError;
-  }
```
amannn, line 115: Seems like this was not necessary in the end when we annotated the return type explicitly.

PR #115, opened 2022-05-27
fix: #112 - default NestedKey to never

### LINKED ISSUE #112: TypeScript validation is broken when a single namespace is available in the messages
./types/next-intl.d.ts;

```ts
type Messages = typeof import('../messages/en.json');
declare interface IntlMessages extends Messages {}
```

./messages/en.json;

```json
{
  "Home": {
    "heroTitle1": "Discover website",
    "heroTitle2": "and freedom!",
    "heroSubtitle": "Vulputate erat cubilia praesent, dolor mattis condimentum eget tellus phasellus auctor elementum morbi, justo torquent egestas, nulla cursus tellus, per.",
    "heroButton": "See More"
  }
}
```

./pages/index.html;
bug

The example provided in the documentation is invalid.

P.S. My typescript version is 4.6.3

kelsny:
Fixes #112 

Modified files:

- `packages/use-intl/src/useTranslations.tsx`

It's just my comment with the ideas applied:

> And in the conditional, I just check if `NestedKey` is `never`: `[NestedKey] extends [never] ? ... : ...`. It seems to work fine.
> 
> You can find the fix [here](https://www.typescriptlang.org/play?#code/PTAEFpM0FcBcCWAbAzgWAFCbgTwA4CmoAcgSnAQCYDSBOA8gGYA89AfKALyj2gEAeFAHaUUoAPYAjAFYEAxnFAB+UAG9MoTaADa1UAiGgA1nXGMeAXQBcoAAYASVXoBkocgCcDAcwC+t0AA+do4ubnCeQr4AdI6k5FS0DCz0uhZsfgDcmD7aJjhmlqA2QgQAbgTuWVgYuISgAOIEcKwANKAACnyCBCJiHt4cnBodXcKiwaoGjBWgiT4xk0LT7qAASmRwfsOaKomjPeN5Bbwqjc0piRZt6+QcxWUV20UjAmNiR+YnPNrtFs8l5UqmGw+CIZ3oMnkcESKFas32vTCES8gzUwx++kM1GsDSacPaHFeBz64W8ylAAJmNnaVRy2KqmBAECg4H0AFs8EgCGyenAAIaIcRCYEYJkAIXc4hMwuqciF5AkMi4aIwmkkfPcNnUqq0fJsAHIABIEJBIcT64Y+bKgPliOVCcgMjCMGBCBQIIWgRh8hTidw4ACMzDiFBodAR43BkIUMOYtQIBSk0jaIYSdCYcdBiZkbFRlPcbAAFEI+TyUHgfQQlDZU2GcABKFWaJm1vYIMRStpwAAWRHtlAQgsM7dAkkl0qiw3cTRg7kM+sY4nNNrEhdb4aJiO0+b+KgARGOpT09889wB3P1Gbx7+u0kVMgDql+8IvtCqTACZldrNLbKIwtT-ADQD3IC9x8FphnVTUm11A1jVNc1LWtW1QDfOAnRdN0hy9H04D9HAP2DDY0xwCMxCjWQYzoWF42zaQPxTEi6wzOjzE-XMixLMsKzkKsa2YxJGx-TRpzgWd50XZdUMLSioVjNjFQYpj4hYlhFI4wlukRdcyP3Q9pRPGxz2fSIbzvaomRZVk5D5U10FleVFHcANlW9X1-QDQt9Wg-VGyZA9x2PTB0NAdwv24dz8P9D9vN8-ywBM9wrzMzAgA)

### DIFF
--- a/packages/use-intl/src/useTranslations.tsx
+++ b/packages/use-intl/src/useTranslations.tsx
@@ -17,7 +17,7 @@ import NestedValueOf from './utils/NestedValueOf';
  * (e.g. `namespace.Component`).
  */
 export default function useTranslations<
-  NestedKey extends NamespaceKeys<IntlMessages, NestedKeyOf<IntlMessages>>
+  NestedKey extends NamespaceKeys<IntlMessages, NestedKeyOf<IntlMessages>> = never
 >(
   namespace?: NestedKey
 ): // Explicitly defining the return type is necessary as TypeScript would get it wrong
@@ -27,17 +27,14 @@ export default function useTranslations<
     TargetKey extends MessageKeys<
       NestedValueOf<
         {'!': IntlMessages},
-        NamespaceKeys<IntlMessages, NestedKeyOf<IntlMessages>> extends NestedKey
+        [NestedKey] extends [never]
           ? '!'
           : `!.${NestedKey}`
       >,
       NestedKeyOf<
         NestedValueOf<
           {'!': IntlMessages},
-          NamespaceKeys<
-            IntlMessages,
-            NestedKeyOf<IntlMessages>
-          > extends NestedKey
+          [NestedKey] extends [never],
             ? '!'
             : `!.${NestedKey}`
         >
@@ -54,17 +51,14 @@ export default function useTranslations<
     TargetKey extends MessageKeys<
       NestedValueOf<
         {'!': IntlMessages},
-        NamespaceKeys<IntlMessages, NestedKeyOf<IntlMessages>> extends NestedKey
+        [NestedKey] extends [never]
           ? '!'
           : `!.${NestedKey}`
       >,
       NestedKeyOf<
         NestedValueOf<
           {'!': IntlMessages},
-          NamespaceKeys<
-            IntlMessages,
-            NestedKeyOf<IntlMessages>
-          > extends NestedKey
+          [NestedKey] extends [never]
             ? '!'
             : `!.${NestedKey}`
         >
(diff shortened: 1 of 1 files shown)


packages/use-intl/src/useTranslations.tsx
```diff
@@ -27,19 +30,12 @@ export default function useTranslations<
     TargetKey extends MessageKeys<
       NestedValueOf<
         {'!': IntlMessages},
-        NamespaceKeys<IntlMessages, NestedKeyOf<IntlMessages>> extends NestedKey
-          ? '!'
-          : `!.${NestedKey}`
+        [NestedKey] extends [never] ? '!' : `!.${NestedKey}`
```
amannn, line 33: From [your issue comment](https://github.com/amannn/next-intl/issues/112#issuecomment-1139945855):

> When using generics as types for optional parameters, and the parameter is not given, the generic is defaulted to its constraint.
So I solved this (in a terrible way I did not test) by testing if the constraint extends NestedKey . Obviously, now that I think about it, this will break for objects with only one top-level property...
I also just learnt this very recently, so now I can solve this problem. I've given NestedKey a default "value" (it's actually a type, of course, but at the type-level I refer to types as "values") of never.
And in the conditional, I just check if NestedKey is never: [NestedKey] extends [never] ? ... : .... It seems to work fine.

I think I understand. Ultimately we want to check if `namespace` was provided at the type-level and `never` would mean that it wasn't. Makes sense to me! It's still a bit strange to me though why have to wrap the types into arrays for comparing them with `extends`.

But it's really great to see that this works! 👏

amannn (commented):
Hey @kelsny, this is incredibly nice of you to directly submit a PR! Thank you so much for your help!

amannn:
I tested your fix in a larger codebase where I'm using `next-intl` and it seems to work flawlessly. Thank you very much for your help @kelsny!

PR #120, opened 2022-08-15
fix: Add ReactNode as allowed type of TranslationValue

omnomnomka:
Seems like ReactNode is acceptable as TranslationValue and it's useful if we want to use React Nodes as translation values.
Example:
```
<NextIntlProvider defaultTranslationValues={{ br: <br /> }}>
    {content}
</NextIntlProvider>
```

And then use it in JSON: `Line 1 {br} Line 2`

Adding it to types helps to omit unnecessary TypeScript transformations and `@ts-ignore` usages.

I'm not sure if it possible to implement such thing as Intl has out-of-the-box: https://formatjs.io/docs/react-intl/components#wraprichtextchunksinfragment to avoid "keys" issue in another way.

### DIFF
--- a/packages/use-intl/src/TranslationValues.tsx
+++ b/packages/use-intl/src/TranslationValues.tsx
@@ -1,7 +1,7 @@
 import {ReactNode} from 'react';
 
 // From IntlMessageFormat#format
-type TranslationValue = ReactNode | string | number | boolean | Date | null | undefined;
+type TranslationValue = ReactNode | Date;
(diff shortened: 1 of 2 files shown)


amannn:
Hi @omnomnomka and thank you for your PR! This looks pretty good. Can you add some tests and fix the lint error that is reported on CI? Thanks! 🙌

amannn:
Closing as stale …

PR #139, opened 2022-10-27
feat: Next.js 13 RSC integration

amannn:
**Things to figure out**
 - [ ] How to provide the context on the server side?
   - `createServerContext` [seems to be sent to the client](https://twitter.com/sebmarkbage/status/1587615484870098945) so we could use it for auto-serialization
   - But how can we make this ergonomic on the server side?
   - If React context is no longer an option we have to figure out another API to configure the lib (e.g. global setter, process.env, Next.js plugin, tsconfig path override or something else)—IMO none of these options are as good as context.
   - Next Auth was able to do this because they [only rely on `headers()` and `cookies()`](https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/next/index.ts).
   - See https://github.com/vercel/next.js/discussions/42301#discussioncomment-4066668
 - [ ] How to get the locale? Should we parse this ourselves since i18n routing is gone ([ref 1](https://beta.nextjs.org/docs/app-directory-roadmap#not-planned-features), [ref 2](https://beta.nextjs.org/docs/upgrade-guide)) ?
 - [ ] Should messages for the server- and client side be separately provided?
 - [ ] Maybe we need to merge messages now when providers are nested?
 - [ ] Support the same i18n config that Next.js currently has?
 - [ ] Potential new features
   - [ ] Locale negotiation (middleware? [API inspiration](https://next-auth.js.org/configuration/nextjs#middleware))
   - [ ] Saving the locale to a cookie
- [ ] Should we utilize [`react-server` exports](https://github.com/reactjs/rfcs/blob/main/text/0227-server-module-conventions.md)?

**Afterwards**
 - [ ] Update docs
 - [ ] Update examples


packages/example/src/app/layout.tsx
```diff
@@ -0,0 +1,29 @@
...
+import ServerOnlyContext from './ServerOnlyContext';
+
+type Props = {
+  children: ReactNode;
+};
+
+export default function RootLayout({children, ...rest}: Props) {
+  console.log(rest);
+
+  // How to get this from the URL?
```
farreldarian, line 12: You can get through layout props

https://beta.nextjs.org/docs/routing/defining-routes#example
_*works in layout also_

farreldarian, line 12: Maybe wrap all pages under optional `[[locale]]` then get using `params`

https://beta.nextjs.org/docs/routing/defining-routes#example

amannn: Hey and thank you for your interest! Yep, that seems to work, I've experimented with it here: https://github.com/amannn/next-intl/blob/ccb9c8afabc962f043c731677d2f68f871e01836/packages/example/src/app/%5Blocale%5D/page.tsx#L11-L15

I think that part is doable, the part where I currently don't see an ergonomic solution is how global configuration like messages can be passed to the library (see the topic about context in the PR description).

farreldarian: I've tried some workarounds, 
1, Using singleton — Failed because for RSC, initializing on root layout requires 2 render to assign the `messages` object.
2. Using third-party state manager like zustand — Failed because of the same reason as number 1.

Probably the solution is gonna be around client component. To use in a server component we could provide utility component like `<T key='dashboard.hi' />`. Performance-wise, next will also server-render client components so the downside is on developer experience. But won't be much I think.

packages/example/src/middleware.ts
```diff
@@ -0,0 +1,26 @@
+import acceptLanguageParser from 'accept-language-parser';
+import {headers} from 'next/headers';
+import {NextResponse} from 'next/server';
+import type {NextRequest} from 'next/server';
+import i18n from './i18n';
+
+// Somehow not invoked?
```
daniel-koudouna, line 7: From a quick glance at the example WIP, the middleware doesn't seem to fire unless there is a "pages" folder as well. After that, it fires on both "pages" and "app" folders.

daniel-koudouna:
Hi, thanks for your work. We had decided to use next-intl for a project a while ago, and we are in the process of slowly migrating to next 13. 

I'm not sure if this will be of any use, but I've added a working example of a header-based solution: https://github.com/daniel-koudouna/next-intl/tree/feat/next-13 

Some things to note:
- I eliminated the need for a root `[locale]` path, as we can just find it from the middleware and rewrite it accordingly.
- I added back the `domains` mappings, which can also be taken from the middleware

I'm not well versed with react internals by any means, this is just another experiment. Let me know what you think, I can open a PR to this branch if it helps.

amannn:
Hey @daniel-koudouna and thanks a lot for chiming in!

This branch is a really raw draft at this point, mostly for experimenting. I'm currently using Next.js internal APIs in `Storage.tsx`, which could break in a patch release. I'd like to get rid of this in favour of a proper solution.

Using `headers` for detecting the locale is a good idea, but we'd also need some storage for the messages and other config. Otherwise we'd have to work around it, but if possible, I'd really like to continue using the current API. I really hope that the Next.js/React will consider an API to help with this (see also https://github.com/vercel/next.js/discussions/42301#discussioncomment-4066668).

That is a bit the dealbreaker currently in my prototype and I'd like to find a better solution before moving forward. I think the other parts should be manageable.

But since Next.js is moving i18n functionality out of the core, probably this library will also participate in locale negotiation in the future. Would be really cool to get input from you when there's a first working prototype! Also great that you mention `domains`, probably we should support the same configuration as Next.js currently does, for an easy migration.

packages/example/src/middleware.ts
amannn: Oh cool, thanks for figuring this out! Seems like a bug on the Next.js side.

daniel-koudouna:
> Hey @daniel-koudouna and thanks a lot for chiming in!
> 
> This branch is a really raw draft at this point, mostly for experimenting. I'm currently using Next.js internal APIs in `Storage.tsx`, which could break in a patch release. I'd like to get rid of this in favour of a proper solution.
> 
> Using `headers` for detecting the locale is a good idea, but we'd also need some storage for the messages and other config. Otherwise we'd have to work around it, but if possible, I'd really like to continue using the current API. I really hope that the Next.js/React will consider an API to help with this (see also [vercel/next.js#42301 (comment)](https://github.com/vercel/next.js/discussions/42301#discussioncomment-4066668)).
> 
> That is a bit the dealbreaker currently in my prototype and I'd like to find a better solution before moving forward. I think the other parts should be manageable.
> 
> But since Next.js is moving i18n functionality out of the core, probably this library will also participate in locale negotiation in the future. Would be really cool to get input from you when there's a first working prototype! Also great that you mention `domains`, probably we should support the same configuration as Next.js currently does, for an easy migration.

No worries, happy to help if I can, although I'm not at all familiar with React internals. I found the headers-based solution pretty easy to add domains to.

amannn:
Closed in favour of #149

Issue #141, opened 2022-11-18
Automatically build examples after installing dependencies in the repo


Kasea:
### Description

Cloned the repo and tried to run example locally, but i keep getting the below error.

```
yarn run v1.22.19
$ next dev
ready - started server on 0.0.0.0:3000, url: http://localhost:3000
error - ./src/pages/_app.tsx:1:0
Module not found: Can't resolve 'next-intl'
> 1 | import {NextIntlProvider} from 'next-intl';
  2 | import {AppProps} from 'next/app';
  3 | 
  4 | export default function App({Component, pageProps}: AppProps) {

https://nextjs.org/docs/messages/module-not-found
```

### Reproduction

Steps to reproduce:

1. git clone https://github.com/amannn/next-intl
2. cd next-intl
3. yarn install && yarn workspace example dev

NOTE: tried various methods of 3, but none seemed to have worked and yielded the same error

### Expected behaviour

Expect it to boot up and be accessible on localhost:3000

### Additional information

The root node_modules seems to be setup correctly, it symlinks the workspaces properly, but it seems like it doesn't resolve the parent's node_modules inside of the packages.

Node version: 16.10.0
Yarn version: 1.22.19

amannn:
Hey and thanks for the report! The reason is likely that `use-intl` and `next-intl` need to be built first in the project. That's what I'm doing on CI:

https://github.com/amannn/next-intl/blob/6c26d0ae98f1c0ccd93ce55efe6ef5d26c07f994/.github/workflows/main.yml#L16-L17

Maybe we could include a postinstall step that does this. I'd be happy to review a PR if you'd like to contribute this change!

Otherwise the examples are also deployed on CodeSandboxe if you want to play around: https://next-intl-docs.vercel.app/examples/minimal

Kasea:
Yeah that did it, I'll be adding the postinstall to my PR when I submit it. Although might still be confusing for people looking to contribute since the build will be required to be ran for every change, but I suppose that might be the norm for workspace projects, not sure as this is the first time I'm working with one.

amannn:
Yep, that's my local workflow as well where I rebuild packages after changes when trying them out in examples. I see that it's a bit a hurdle, but I'm unfortunately not aware of a better solution. On CI the workflow is managed, so we can rely on up-to-date versions across the workspaces for the tests.

Thanks!

amannn:
[...]

PR #142, opened 2022-11-22
feat: Add standalone "getTranslation" for React

Kasea:
Fixes https://github.com/amannn/next-intl/issues/141

This adds a `getTranslation` helper that will work the same as `useTranslations(<namespace>)(<message>)`. It uses the same translations as the closest translation Provider*.

The use case for this helper is in the event you want to access translations outside of a hook/react component, but within your react application, for instance an error handler without needing to pass a "translation" function to the error handler.

Caveats:
* It won't use the closest Provider in the event of memo'ing of component(s) that stops full tree revalidation. (would say it isn't really a relevant case)

* needs to be used within the Provider, but only if you'll be using this in the initial render. Outside of initial render it won't _technically_ be a requirement to be needed to be used within the Provider, it will instead use the "closest" provider. Which provider that ends up being depends on where you'll be using the helper. (would say not really relevant due to single provider)

* it will not be replacing the react hook, it's more of a "edge case" helper

Notes / Todos:
* Not super happy with the API in the current PR, any alternative suggestions?

* PR doesn't update docs, as I'm not sure if we'll keep this API and/or that such a helper will get accepted. -- In the event both are resolved, it will update docs.

* PR doesn't add tests, same reasoning as the above; will also be added in the event both of the above gets resolved.

* PR doesn't follow project coding conventions in `getTranslation.ts`. Will be updated in the event the above gets resolved.

* PR updates example code, but this leads to the text in the example code to look confusing. This should be removed and/or changed in some way.

* Didn't create this as a draft PR, even though it technically is. Reasoning being that I want feedback and not sure if the maintainer(s) would do that for a draft PR. Feel free to change it.

### DIFF
--- a/packages/use-intl/src/react/IntlContext.tsx
+++ b/packages/use-intl/src/react/IntlContext.tsx
@@ -21,4 +21,7 @@ export type IntlContextShape = {
 
 const IntlContext = createContext<IntlContextShape | undefined>(undefined);
 
+// used for getTranslation helper
+export const IntlContextValue: IntlContextShape = {} as any;
+
 export default IntlContext;
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .DS_Store
 node_modules
 dist
+.vscode
--- a/package.json
+++ b/package.json
@@ -3,7 +3,8 @@
   "private": true,
   "scripts": {
     "publish": "lerna publish --yes --no-verify-access",
-    "prerelease": "lerna publish prerelease --yes --no-verify-access --dist-tag next --no-changelog --no-git-tag-version"
+    "prerelease": "lerna publish prerelease --yes --no-verify-access --dist-tag next --no-changelog --no-git-tag-version",
+    "postinstall": "yarn workspace use-intl run build && yarn workspace next-intl run build"
   },
   "workspaces": [
     "packages/use-intl",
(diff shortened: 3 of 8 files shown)


packages/use-intl/src/react/IntlContext.tsx
```diff
@@ -21,4 +21,7 @@ export type IntlContextShape = {
 
 const IntlContext = createContext<IntlContextShape | undefined>(undefined);
 
+// used for getTranslation helper
+export const IntlContextValue: IntlContextShape = {} as any;
```
amannn, line 25: This is problematic as it breaks the isolated nature of context and instead creates a global singleton.

amannn (commented):
Hello @Kasea and thank you for your interest in contributing!

Unfortunately I don't think this a good direction for the library, as a) the feature is already supported in user land (you can use and wrap [`createTranslator`](https://next-intl-docs.vercel.app/docs/usage/core-library) as necessary) and b) the implementation itself is problematic (see my inline comment).

For usage within React, the `useTranslations` hook is the go-to way and IMO the places where it's not possible / not convenient to use the hook, typically are affected by some architectural issues. So refactoring in the app would be my suggested solution. If you need help with refactoring, I'd recommend adding a question to Stack Overflow with a reduced CodeSandbox, I check in occassionally there to answer questions and other times, other people from the community help out!

I hope you understand!

Kasea:
Understandable, can't say I disagree with most of what you said, the exception being the architectural standpoint.

I suppose I'll create a specific helper that accomplishes this in the codebase instead. Thanks for your time and reply.

PR #144, opened 2022-11-24
feat: Add support for using `next-intl` in the `app` folder with Next.js 13

### LINKED ISSUE #147: Support for nextjs v13

Actually I see that all the examples are based on nextjx v12 that used only the `pages` folder, but now in nextjs v13 they implemented the `app` page (https://nextjs.org/blog/next-13), so I wonder if there is a way to use it with nextjs v13 and if not, it would be great to implement it.

Examples or compatibility with nextjs v13.

I do not know.

amannn:
This PR includes the minimal change that we have to do to be able for `next-intl` to be used in the app folder.

**Features:**
 - `NextIntlProvider` can now be used in the `app` folder. Note that all components that make use of `useTranslations` or `useIntl` still need to be marked with `'use client';`. A way to remove this restriction is explored in #149.
 - Expose a middleware that can be used to set up i18n routing. This is available via `import {createIntlMiddleware} from 'next-intl/server';`

**Pre-release:**

```
npm install next-intl@2.10.0-alpha.6
```

[CodeSandbox](https://csb-k2ien9-7ytkomg4x-amann.vercel.app/)

Fixes #147

---

**What's missing**
1. A way to read the current `pathname` in the root layout to set the `<html lang={…} />` prop correctly ([reference](https://github.com/amannn/next-intl/pull/144/files#diff-ee0c62bc389b67f49f92829f0a5c056f46198f7f9549102b505e7e43343fa8be)).

**What's nice to have**
1. Setting arbitrary values like `locale` directly in the request storage (like `cookies()`) to be able to use this conveniently everywhere
2. Being able to use `headers()` and `cookies()` in middleware for a unified interface
3. Setting values with the `cookies()` API

**Potential TODOs**
 - [x] Publish `next-intl@2.10.0-alpha.6`
 - [ ] ~~Generate locales statically~~ This is not called for me during the build, no idea why. Bug?
 - [x] Explore using `next-intl` in head
 - [x] Tests
 - [ ] Domain routing
 - [ ] Set `html[lang]`


packages/example-next-13/src/next-intl/README.md
```diff
@@ -0,0 +1 @@
+This folder contains a draft of APIs that will be moved to `next-intl`.
```
amannn, line 1: TODO: Evaluate if we want to move some of them inside immediately.

PR #156, opened 2022-12-20
feat: RSC with static config

amannn:
See proposed docs: https://next-intl-dtas8vhjp-amann.vercel.app/docs/next-13/server-components

**TODO:**
 - [ ] Wait for this to be resolved: https://github.com/vercel/next.js/issues/44222
 - [ ] Create a bug report that tsconfig paths don't work for node_modules
 - [ ] Get rid of `Warning: Invalid hook call.`

Changes:
 - `NextIntlServerProvider` now only takes the props which can only be known at runtime
 - Other config goes into a static config
 - `LocalizedLink` was added (TODO: Add docs)

It seems like config gets a bit more convoluted, but [that might be necessary](https://twitter.com/sebmarkbage/status/1587615485910188032). The best part currently IMO is that we can continue to use the existing APIs in components.

Notes to myself:
- Difference between `createServerContext` and `'use client'` + `createContext`: `'use client'` can only be used to mark components, not arbitrary modules. So this relies on the fact that the APIs we provide are components and not e.g. hooks. That's why we can't use `createContext` for the runtime config, since we need to consume it from hooks. Related: https://github.com/vercel/examples/pull/496


daniel-koudouna:
Is there something else that is needed in the `tsconfig.json` ? I get the error that the file could not be found despite having the correct path in `tsconfig.json` . Looking at the compiled code, the default file is still loaded for me:

```js
/***/ "(sc_server)/./node_modules/next-intl/config.js":
/*!******************************************!*\
  !*** ./node_modules/next-intl/config.js ***!
  \******************************************/
/***/ (() => {

eval("throw new Error(\"Couldn't find next-intl config file. Please follow the instructions at https://
```

I'm on `2.11.0-alpha.4`

amannn:
> Is there something else that is needed in the `tsconfig.json` ?

Hmm, not really! Here's a working example: https://github.com/amannn/next-intl/tree/feat/next-13-rsc-static-config/packages/example-next-13.

If you can provide a CodeSandbox reproduction or so, I'm happy to have a look!

But also, this PR is still WIP. Please don't feel obliged to adapt to every change, there will likely be more! I'm sharing this early to be transparent and for potential feedback.

amannn:
I've also created [a bug report for the issue I'm seeing in this branch](https://github.com/vercel/next.js/issues/44222).

amannn:
> I get the error that the file could not be found despite having the correct path in `tsconfig.json`

I've just reproduced the error and will look into this, thanks for reporting! Seems to happen only with the version that's published to NPM …

amannn:
@daniel-koudouna Got it working now in case you're curious: https://github.com/amannn/next-intl-example-next-13 ([demo](https://next-intl-example-next-13.vercel.app/en))

Seems like [Next.js module path aliases](https://nextjs.org/docs/advanced-features/module-path-aliases) have an issue when being used from within `node_modules`. I might have to open a bug report for this, for now I've switched to a webpack alias.

I generally still have hope to simplify the configuration, but this is at least the current state. The biggest blocker currently is [not being able to switch pages](https://github.com/vercel/next.js/issues/44222).

daniel-koudouna:
I can confirm that using the webpack config works for me. At this point, we are using both providers, e.g

```ts
      <NextIntlServerProvider locale={locale}>
        <NextIntlClientProvider locale={locale} messages={localeMessages}>
```

which, for now, enables the whole site to work. Not running through the middleware also allows us to preserve app state.

The standalone example doesn't work for me, with a hydration error:

```
TypeError: Cannot read properties of undefined (reading 'call')
```

seemingly originating from:

```
/***/ "./node_modules/next/dist/client/link.js":
```

PR #167, opened 2023-01-12
Better dark mode support for docs

apriltaoyvr:
## Purpose
The docs site does not look great in dark mode. This PR is to make some small adjustments to improve the dark mode experience for it.

## Changes
- Edited `logo.svg` to be transparent
- Added modular css ala the [Nextra Docs repo](https://github.com/shuding/nextra/tree/main/docs)
  - `/styles/index.module.css` contains the current inline styles for the home page
  - Announcement banner now has dark styling
- Moved `styles.css` to `/styles`
- Updated `tailwind.config.js` to match Tailwind V3 settings 
  - Removed JIT since it is now the default mode for Tailwind
  - Changed `purge` to `content`

### Current site
image
### After style and logo adjustments
image


amannn:
Oh wow, thank you so much for looking into this! ❤️❤️❤️

I have to say I didn't even know that the docs support dark mode (even if broken). How did you switch to dark mode? I'm generally a light mode user, but when I switch to dark mode on macOS the next-intl docs sight remains bright for me (unlike other pages which switch automatically, like the Tailwind docs).

However with the generated preview of your changes, I do see the dark mode. Is that due to the Tailwind update?

I'm definitely interested in supporting dark mode. Would you mind inverting the logo text color for dark mode?

I hope that I'll find some time to update Nextra at some point, I think that should also improve the styling of the page a bit.

apriltaoyvr:
> I have to say I didn't even know that the docs support dark mode (even if broken). How did you switch to dark mode? I'm generally a light mode user, but when I switch to dark mode on macOS the next-intl docs sight remains bright for me (unlike other pages which switch automatically, like the Tailwind docs).
>
> However with the generated preview of your changes, I do see the dark mode. Is that due to the Tailwind update?
I always browse in dark mode so it defaulted to dark for me, but I'm not sure why Nextra isn't detecting dark mode for you when you switch.

There is a theme switcher available in the navigation bar on the [Nextra template](https://nextra-docs-template.vercel.app/) and it seems to be toggled by [`darkMode`](https://nextra.site/docs/docs-theme/theme-configuration#dark-mode-and-themes) which is currently set to false. I've reenabled it in this push and it should be available now.

> I'm definitely interested in supporting dark mode. Would you mind inverting the logo text color for dark mode?
I wasn't sure how to do it so I copied the idea from the [Nextra docs repo](https://github.com/shuding/nextra/blob/main/docs/theme.config.tsx). 

image

The downside is that the svg is now in the theme config file. I didn't delete the main .svg just in case, but if you want to save a bit of extra space it should be safe to do.

I made a few more pushes to add the above things in along with updating eslint and prettier configs to prevent a lot of unneccessary errors (they were not quite synchronized so it would spit out eslint errors for single quotes + a lot of Delete `␍` which could not be autofixed due to no end of line rule being stated). I'll update the OP message in this PR to reflect this.

packages/website/theme.config.js
```diff
@@ -1,4 +1,86 @@
...
+        transform="translate(-197.5 -171.5)"
+        fill="currentColor"
```
amannn, line 16: Cool idea! 👏

packages/website/.eslintrc.js
```diff
@@ -4,12 +4,13 @@ module.exports = {
   extends: [
     'molindo/typescript',
     'molindo/react',
-    'plugin:@next/next/recommended'
+    'plugin:@next/next/recommended',
+    "prettier"
```
amannn, line 8: Prettier is already added [via `molindo/typescript`](https://github.com/molindo/eslint-config-molindo/blob/539c8f16049a120a4b0b7ee900b2a9738d386f34/javascript.js#L74-L77).

Can you revert this?

packages/website/.eslintrc.js
```diff
@@ -4,12 +4,13 @@ module.exports = {
...
   ],
   rules: {
     'react/react-in-jsx-scope': 'off',
     'react/prop-types': 'off',
     'jsx-a11y/anchor-is-valid': 'off',
-    'react/display-name': 'off'
-  }
+    'react/display-name': 'off',
+  },
```
amannn, line 15: Can you revert the formatting changes please?

packages/website/.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+  "trailingComma": "es5",
+  "tabWidth": 2,
+  "semi": true,
+  "singleQuote": true,
+  "endOfLine": "auto"
+}
```
amannn, line 7: As mentioned above, this is already handled via ESLint. There's no need to call or use prettier separately.

packages/website/theme.config.js
```diff
@@ -9,11 +91,11 @@ export default {
   prevLinks: true,
   search: true,
   customSearch: null,
-  darkMode: false,
-  footer: true,
+  darkMode: true,
+  footer: false,
```
amannn, line 95: Ah right, that's better—thanks!

packages/website/package.json
```diff
@@ -5,7 +5,6 @@
   "scripts": {
     "dev": "next dev",
     "lint": "eslint pages components",
-    "test": "echo 'No tests yet'",
```
amannn, line 8: This needs to be reverted, otherwise CI fails.

amannn (commented):
Thank you so much, this looks great! 👏  I left some minor nits, but generally this is good to go.

> There is a theme switcher available

That's really cool! 🙌 

> I don't mind making another PR to update Nextra if you would like to update.

If you feel like it, you're more than welcome! ❤️ 

> You can use the banner config for announcements which will display a thin banner at the very top of the screen

We could look into that, not sure what it looks like. The current banner is not too bad IMO, but I'm open to suggestions!

> More options for callouts

That sounds interesting. I wasn't very happy with the current style of callouts, they look more like warnings but are in fact more intended as tips—so something more greenish would be more suitable IMO.

> I don't think the tailwind package is being used (but that makes the preview dark mode weirder), which is why I used modular css. This is because there's no tailwind preflight in styles.css. I think you noticed this too because some elements were inline but others use tailwind classes.

Yep, exactly. I noticed some problems where the stylesheet didn't seem to be compiled based on used classes, but I didn't have the time to look into it. Maybe something to do after / as part of a Nextra upgrade? Or not sure if the situation here is already improved based on your changes to `tailwind.config.js`?

packages/website/theme.config.js
```diff
@@ -1,4 +1,86 @@
...
+        fill="currentColor"
+      />
+      <g id="Oval-2">
+        <path
+          id="Oval"
+          d="M528.8,509q-6.3,6-13.2,11.4a190,190,0,1,1,38-40.2"
+          transform="translate(-197.5 -171.5)"
+          fill="none"
+          stroke="#5fc3e7"
+          stroke-width="18"
```
amannn, line 25: Oh and I just noticed there are a few warnings on the console, the dashed attributes should be converted to camel case.

packages/website/.prettierrc
apriltaoyvr: The prettier and prettier-eslint plugins were both unfortunately unable to detect styling without it and would error out without it. The alternative is to do the same but in the eslint file or for future editors to also be unable to autolint

apriltaoyvr:
I took a read through your comments and I'll have the suggested changes done pretty quickly. Thanks for letting me making some changes to the docs 🙏 and I'll update it to Nextra when I'm free again

 >Or not sure if the situation here is already improved based on your changes to `tailwind.config.js`?
>
It didn't make much of a difference to be honest 😢. My first approach was to do tailwind for the inline styles and that's when it hit me that you already tried that too too. So I already had the config set up and figured it wouldn't hurt to have the updated one in the repo if it's already done.

I tried Tailwind with the preflight with the repo's version of Nextra and the current one. For this repo's version, it'll override some of Nextra's default styling (mostly links and other basic elements) so I stayed away from it. But it'll allow you to use Tailwind as intended. 

For the current version of Nextra, it seems to work without overriding anything. They prefix their styles with `nx` now so it's fairly safe to do. It's up to you which CSS approach you prefer

amannn (approved):
Really cool, thank you so much for looking into this! 🙌 🌚 

> For the current version of Nextra, it seems to work without overriding anything. They prefix their styles with nx now so it's fairly safe to do. It's up to you which CSS approach you prefer

I don't have a strong opinion here, generally I'd like to stay close to what Nextra does out-of-the-box to make updates as easy as possible in the future.

amannn:
Just FYI, I had some time to look into the Nextra upgrade: https://github.com/amannn/next-intl/pull/197. The PR is WIP currently.

PR #186, opened 2023-02-07
feat: Unprefixed routing for default locale & domain-based locale matching

amannn:
This enables to request pages without a locale prefix, if the user is matched to the default locale (e.g. `/about`). If another locale matches, the user is redirected (e.g. `/about` -> `/de/about`).

On top of this, this enables domain-based locale matching (e.g. `de.example.com/about`, `example.de/about`).

**Deprecation info**

`NextIntlClientProvider` should now be imported from `next-intl` (not `next-intl/client`).

The reasoning is that we need to reserve the namespace `next-intl/client` for anything that can only be imported from a Client Component. The provider can apply `use client` itself, therefore it's suitable for RSC as well.


Don-Cryptus:
when is this getting merged?

amannn:
> when is this getting merged?

Now 🙂

PR #189, opened 2023-02-14
fix: add missing 'use client' directive for usePathname

wxh06:
```
error - ./node_modules/.pnpm/next-intl@2.11.0-beta.8_next@13.1.6+react@18.2.0/node_modules/next-intl/dist/src/client/usePathname.js
ReactServerComponentsError:

You're importing a component that needs useState. It only works in a Client Component but none of its parents are marked with "use client", so they're Server Components by default.

   ,-[./node_modules/.pnpm/next-intl@2.11.0-beta.8_next@13.1.6+react@18.2.0/node_modules/next-intl/dist/src/client/usePathname.js:1:1]
 1 | import { usePathname as useNextPathname } from 'next/navigation';
 2 | import { useEffect, useState } from 'react';
   :                     ^^^^^^^^
 3 | import getCookieLocale from './getCookieLocale';
 4 | import hasPathnamePrefixed from './hasPathnamePrefixed';
 5 | export function unlocalizePathname(pathname) {
   `----

Maybe one of these should be marked as a client entry with "use client":
  node_modules/.pnpm/next-intl@2.11.0-beta.8_next@13.1.6+react@18.2.0/node_modules/next-intl/dist/src/client/usePathname.js
  node_modules/.pnpm/next-intl@2.11.0-beta.8_next@13.1.6+react@18.2.0/node_modules/next-intl/dist/src/client/index.js
  app/[locale]/layout.tsx
```

Reference: <https://beta.nextjs.org/docs/rendering/server-and-client-components#convention>

Introduced in https://github.com/amannn/next-intl/commit/68ce7db79383900de2d46bb69dd5c175a6d0c7d1#diff-b67520fde3f77290ca84cfb11466b0d730d23909f7c67ee7150ee2888e9307b1R2

https://github.com/amannn/next-intl/blob/b301d472c9d1fd7ec27eff344b677c04540333dc/packages/next-intl/src/client/usePathname.tsx#L2

### DIFF
--- a/packages/next-intl/src/client/usePathname.tsx
+++ b/packages/next-intl/src/client/usePathname.tsx
@@ -1,3 +1,5 @@
+'use client';
+
 import {usePathname as useNextPathname} from 'next/navigation';
 import {useEffect, useState} from 'react';
 import getCookieLocale from './getCookieLocale';


amannn:
Hey @wxh06,

thanks for the PR. Does this improve the error in some way?

The way I see it is that the component (or a parent) that uses `usePathname` needs to be marked with `'use client';` nonetheless, so this doesn't really change the result.

wxh06:
In fact, I had never used `usePathname` in my code.

The error appears when trying to import anything from `next-intl/client` in a server component. It works now after changing the import statement as I noticed that importing `NextIntlClientProvider` from `next-intl/client` has been deprecated:

```diff
-import { NextIntlClientProvider } from "next-intl/client";
+import { NextIntlClientProvider } from "next-intl";
```

I guess the `import` statements in `index.tsx` cause the problem:
<https://github.com/amannn/next-intl/blob/feat/next-13-rsc/packages/next-intl/src/client/index.tsx#L5-L6>

amannn:
Yep, the import for `NextIntlClientProvider` was changed for this very reason that the `/client` path is only available to the client side.

> I guess the import statements in index.tsx cause the problem

It's not a problem, but by design that you can't import client-only functionality on the server side.

amannn:
That's a fair point! Unfortunately, I wasn't aware of a way to deprecate the old import for `NextIntlClientProvider` in a backward-compatible way.

I'll mark the latest beta version as a new major version and will include a note about upgrading in the documentation.

Thanks!

wxh06:
In fact, the current approach of reserving deprecated export of `NextIntlClientProvider` is ineffective.
However, the error can be resolved by adding `'use client';` to the beginning of `usePathname.tsx` as a temporary solution.

amannn:
Oh, that's interesting, you're right! Sorry, now I finally see your point, I thought this would still print an error message! 🙂 

Sounds good!

PR #190, opened 2023-02-14
docs: update deprecated import of `NextIntlClientProvider`

wxh06:
Update deprecated import of `NextIntlClientProvider` in Next 13 client component docs

```diff
-import {NextIntlClientProvider} from 'next-intl/client';
+import {NextIntlClientProvider} from 'next-intl';
```

Importing `NextIntlClientProvider` from `next-intl/client` cannot work properly at all (#189)


amannn:
Hey and thanks again for a PR,

in the current stable release it's correct to import `NextIntlClientProvider` from `next-intl/client`—so this is still up to date.

You're referring to the latest beta version that's being covered on the Server Components docs page, where this import has changed. The beta will likely turn into a new major version where the import needs to be updated. Until then, the docs for usage in Client Components is still up to date.

wxh06:
I'm sorry. I forgot only the server component requires installing beta release

PR #191, opened 2023-02-16
docs: fix path in server-components.mdx

Tobi-mmt:
This fixes the wrong path in `server-components.mdx`.  the Folder `messages` and `i18n.ts` are at the same level.


amannn (approved):
Right, thank you very much! 🙌

PR #206, opened 2023-03-01
feat: Next.js 13 RSC caching

amannn:
Unfortunately [Vercel doesn't support the `vary` header](https://github.com/vercel/vercel/discussions/7612#discussioncomment-2434736) currently which is used by RSC to differentiate prefetch requests, this [breaks page navigation](https://github.com/amannn/street-photography-viewer/pull/2).

Potential solutions:
 - Vercel supports `vary`
 - We can detect prefetch requests in the middleware to not cache those
 - Add caching via `next.config.js` or `vercel.json` and opt-out of prefetch header?

### DIFF
--- a/packages/example-next-13-advanced/src/middleware.ts
+++ b/packages/example-next-13-advanced/src/middleware.ts
@@ -1,6 +1,8 @@
+import ms from 'ms';
 import createIntlMiddleware from 'next-intl/middleware';
+import {NextRequest} from 'next/server';
 
-export default createIntlMiddleware({
+const middleware = createIntlMiddleware({
   locales: ['en', 'de', 'es'],
   defaultLocale: 'en',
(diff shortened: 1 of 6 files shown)


packages/example-next-13-advanced/src/middleware.ts
```diff
@@ -15,6 +17,20 @@ export default createIntlMiddleware({
...
 });
 
+export default function handler(req: NextRequest) {
+  const response = middleware(req);
+
+  response.headers.set(
+    'Cache-Control',
+    [
+      `s-maxage=` + ms('1d') / 1000,
+      `stale-while-revalidate=` + ms('1y') / 1000
```
amannn, line 27: Maybe we should use non-CDN caching here too so we get a failing test.

PR #225, opened 2023-03-24
feat: BigInt support for useFormatter.format() (#222)

### LINKED ISSUE #222: BigInt support for useFormatter.format()

In TypeScript, cannot easily use `BigInt` with `useFormatter().format(value)` as only accepted type of value is `number`.

Change type of `value` parameter of `useFormatter().number()` function from `number` to `number | BigInt`.

https://github.com/amannn/next-intl/blob/be79c1b816314f16f6825ee9cde50d076ba342e6/packages/use-intl/src/core/createFormatter.tsx#L133

Add a new function as opposed to changing existing, if compatibility is a concern. Note, however, that `Intl.NumberFormat(...).format(value)` does support `BigInt`.

tacomanator:
Added a single `BigInt` test to each  `createFormatter` and `useFormatter`. I also noticed was no test for plain-old 'number' without currency. I thought it made sense to have one so I broke the existing test into two.

Closes #222

### DIFF
--- a/packages/use-intl/test/core/createFormatter.test.tsx
+++ b/packages/use-intl/test/core/createFormatter.test.tsx
@@ -12,9 +12,23 @@ it('formats a date and time', () => {
 });
 
 it('formats a number', () => {
+  expect(intl.number(123456)).toBe('123,456');
+});
+
+it('formats a bigint', () => {
+  expect(intl.number(123456789123456789n)).toBe('123,456,789,123,456,789');
+});
+
+it('formats a number as currency', () => {
   expect(intl.number(123456.789, {style: 'currency', currency: 'USD'})).toBe(
     '$123,456.79'
   );
+});
+
+it('formats a bigint as currency', () => {
+  expect(
+    intl.number(123456789123456789n, {style: 'currency', currency: 'USD'})
+  ).toBe('$123,456,789,123,456,789.00');
 });
 
 it('formats a relative time', () => {
--- a/packages/use-intl/test/react/useFormatter.test.tsx
+++ b/packages/use-intl/test/react/useFormatter.test.tsx
@@ -191,7 +191,7 @@ describe('dateTime', () => {
 });
 
 describe('number', () => {
-  function renderNumber(value: number, options?: NumberFormatOptions) {
+  function renderNumber(value: number | bigint, options?: NumberFormatOptions) {
     function Component() {
       const format = useFormatter();
       return <>{format.number(value, options)}</>;
@@ -207,6 +207,11 @@ describe('number', () => {
   it('formats a number', () => {
     renderNumber(2948192329.12312);
     screen.getByText('2,948,192,329.123');
+  });
+
+  it('formats a bigint', () => {
+    renderNumber(123456789123456789n);
+    screen.getByText('123,456,789,123,456,789');
   });
 
   it('accepts options', () => {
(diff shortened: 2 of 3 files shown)


packages/use-intl/test/core/createFormatter.test.tsx
```diff
@@ -12,6 +12,18 @@ it('formats a date and time', () => {
 });
 
 it('formats a number', () => {
+  expect(intl.number(123456)).toBe(
+    '123,456'
+  );
+});
+
+it('formats a bigint', () => {
+  expect(intl.number(123456789123456n)).toBe(
```
amannn, line 21: Is this a big int? I guess based on the suffix "n" it is, but the `Number.MAX_SAFE_INTEGER` is 9007199254740991, so this could be represented by a regular number as well.

Maybe we could make the number a bit larger so it exceeds `MAX_SAFE_INTEGER` just to be safe?

packages/use-intl/test/react/useFormatter.test.tsx
```diff
@@ -209,6 +209,11 @@ describe('number', () => {
     screen.getByText('2,948,192,329.123');
   });
 
+  it('formats a bigint', () => {
+    renderNumber(294819232912312);
```
amannn, line 213: This is a regular number, not a big int—right? See the comment above.

amannn (commented):
Thank you so much for your pull request—looks really good already! 🙌

packages/use-intl/test/core/createFormatter.test.tsx
tacomanator: Yep. `123n` and `BigInt(123)` (and `BigInt(123n)` for that matter) appear to be equivalent. Truly not sure how I managed to use a number smaller than `MAX_SAFE_INTEGER` but anyway I fixed it up with a larger number.

packages/use-intl/test/react/useFormatter.test.tsx
tacomanator: Completely batched job. Apologies for that. Updated PR.

tacomanator:
Sorry for the terrible PR. Should be fixed up now.

I also decided to add a test for formatting `BigInt` as currency. I had left it out thinking it was redundant, but it does have value. `BigInt`s cannot be fractions, but `Intl.NumberFormat().format()` still treats them like one. This might not be obvious to so I added the test as a form of 'documentation.'

```ts
Intl.NumberFormat("US-en", {style: 'currency', currency: 'USD'}).format(123) // $123.00
Intl.NumberFormat("US-en", {style: 'currency', currency: 'USD'}).format(123.00) // $123.00
Intl.NumberFormat("US-en", {style: 'currency', currency: 'USD'}).format(BigInt(123)) // $123.00
Intl.NumberFormat("US-en", {style: 'currency', currency: 'USD'}).format(BigInt(123.00)) // error
```

amannn (approved):
> I also decided to add a test for formatting `BigInt` as currency. I had left it out thinking it was redundant, but it does have value. `BigInt`s cannot be fractions, but `Intl.NumberFormat().format()` still treats them like one. This might not be obvious to so I added the test as a form of 'documentation.'

That's a great point, thank you so much for adding the test!

I really appreciate how unobtrusive your proposed change is and how you've picked up the existing code style!

Thank you so much for this great PR! 👏

tacomanator:
Thank you for the feedback _and_ thank you for providing this library for us all to use 🥇 👍

Issue #6, opened 2020-11-29
Exception on not existing key or value - is this the wanted behavior? 


ilanbm:
If a key does not exist in the namespace, or if the value is empty, `t` function raises an exception. 

The key can be debatable, but why would an empty value raise an exception? 

Is there a default value to set if there is no key/value?

amannn:
Thanks for the report!

> If a key does not exist in the namespace

Do you mean a scenario like this?

```js
function Component() {
  const t = useTranslations('Component');

  return <Text>{t('nonExistingKey')}</Text>
}
```

```js
// en.json
{
  "Component": {
    "existingKey": "Hello"
  }
}
```

Currently `next-intl` would throw here and I think that's helpful to fix the error. What would be your expected behaviour?

> if the value is empty

That's true, I guess we could avoid the error here. However I'm wondering when you'd practically want to use an empty label? Why would you want a translation for that? Isn't this a situation where the developer might has forgot to set up a message?

> Is there a default value to set if there is no key/value?

What would you want to put there? Something like "Sorry, no translation available currently."? I'm wondering if that's really helpful to the user. If you have a locale where you haven't completed all translations yet, you could deep merge messages from the default locale to have a fallback (if this is desired – you wouldn't get errors for missing translations then).

ilanbm:
> Currently next-intl would throw here and I think that's helpful to fix the error. What would be your expected behaviour?

Well I think it is wrong to consider that an error, not having a key is legit. It is not code, it can come from various sources and in different languages.
Besides that, its ok not to have a key and expect a predefined default value (that's the common case in most of the other libraries)

 

> That's true, I guess we could avoid the error here. However I'm wondering when you'd practically want to use an empty label? Why would you want a translation for that? Isn't this a situation where the developer might has forgot to set up a message?

Nope, you can't narrow it to a case where a developer/translator forgot to set a value to a key, there are various scenarios of having a wanted empty value. It caused me to look for workarounds here, ended using a transparent char... it my case some keys in some languages don't have a known translation so if there is no value I use the English translation. So having an empty value is a wanted scenario. 

> What would you want to put there? Something like "Sorry, no translation available currently."? I'm wondering if that's really helpful to the user. If you have a locale where you haven't completed all translations yet, you could deep merge messages from the default locale to have a fallback (if this is desired – you wouldn't get errors for missing translations then).

Keep it simple and predictable, return exactly what it is: an empty string.

ilanbm:
[...]

PR #231, opened 2023-03-30
docs: add multiple locales command tip

eug-vs:
I found an easy way to merge locale files, figured this might be useful to add to the docs.

## The problem
Suppose I have default `en.json` file, which I was updating when I was developing the application and getting Typescript auto-completions from it:
```json
{
  "ProductPage": {
    "description": "Description",
    "location": "Location: {location}"
  },
  "Header": {
    "navigation": {
      "themeSwitch": "Toggle dark mode",
      "newRental": "Add new rental",
      "bookings": "My bookings",
      "favorites": "Favorite rentals",
      "avatar": "My profile"
    }
  }
}
```

Now let's say my application also supports BY locale, but my `by.json` is incomplete yet:
```json
{
  "ProductPage": {
    "description": "Апiсанне",
    "location": "Месцазнаходжанне: {location}"
  }
}
```

Adding new messages to BY locale manually sucks. What if I had more locales? What if I had more messages? Here's a quick command that I came up with to merge JSON files and mark missing translations with TODO:
```bash
sed 's/: "/: "TODO: /' en.json | cat - by.json | jq add -s
```
image

This is not ideal of course (doesn't work on windows, depends on `jq`), we could probably write a JS script for that (some `_.mapValues` + `_.defaults` combination), there are many ways to do it, but I just wanted to raise this to see if there's already a better solution for it? And if not we can think about other possible options

### DIFF
--- a/packages/website/pages/docs/faq.mdx
+++ b/packages/website/pages/docs/faq.mdx
@@ -9,7 +9,7 @@
 
 1. This library is built around the concept of namespaces and that components consume a single namespace (however this is not required).
 2. This library offers a hooks-only API for message consumption. The reason for this is that the same API can be used for attributes as well as `children`.
-3. This library currently doesn't support AST-based extraction like `react-intl`.
+3. This library currently doesn't support AST-based extraction like `react-intl`. You can easily [work with multiple locales](#can-next-intl-update-locale-files-based-on-default-locale-file) though.
 4. This library is a bit smaller in size ([next-intl](https://bundlephobia.com/result?p=next-intl) vs [react-intl](https://bundlephobia.com/result?p=react-intl) on BundlePhobia).
 5. This library offers [type safety for message keys](/docs/usage/typescript).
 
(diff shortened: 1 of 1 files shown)


amannn:
Hello @eug-vs thank you so much for sharing your solution and creating this pull request!

This question has come up before and back then my recommendation was to merge messages on the server side at runtime: https://github.com/amannn/next-intl/issues/6#issuecomment-737485734

I feel like the solution here really depends on your expectations. E.g. you might want to do this before the build or at runtime, maybe you want to merge messages or maybe you want to add a message for all missing keys—or a prefix like you did here.

I think I wouldn't add this to the docs at this point, since it's hard to recommend a single solution. But I've linked to this PR from the original issue (#6) so that other developers might find this. 

Does this sound reasonable to you?

eug-vs:
Hey @amannn thanks for looking into this! I understand that there's a possibility to merge messages on-the-fly at runtime, but I was trying to solve a bit different problem. My main concern is a bit different, let's follow the original example again:
1. As a developer, I'm working on a new feature
2. I've added messages to default locale file (`en.json`) and use them to power auto-completions throughout my code
3. I've finished the feature, I already have default EN messages, but I'm still lacking other locales, like BY
4. I could leave it like this and fallback to EN messages at runtime as you mentioned, but let's say I wanna get it fully localized for the purpose of this example
5. Now I need to edit `by.json` manually, but this JSON still has the old shape (doesn't even have new messages in place), so I need to visually "diff" both files, copy the new messages from `en.json` to `by.json` and then finally start translating. This is very error-prone and it's really easy to miss new messages somewhere in the middle of the file, especially if the file is large. Even after "copying the shape" step I can also miss a message if the languages are visually similar, leaving it untranslated.

Step 5 is very painful - and that's the problem I'm proposing a solution to. If there's already a better/standard way to do it, I'd be happy to know about it. Also this gets incredibly harder if the person who's doing translations is not a developer. With my solution, they can just go to the `by.json` file, search for `TODO` in the file and translate the messages.

eug-vs:
> I think I wouldn't add this to the docs at this point, since it's hard to recommend a single solution. But I've linked to this PR from the original issue (https://github.com/amannn/next-intl/issues/6) so that other developers might find this.
Does this sound reasonable to you?

Sure, as long as other people can find it I'm good! I probably should've started a github discussion first, not a PR haha

PR #241, opened 2023-04-13
fix: invalid i18n warning

narakhan:
The default configuration object provided when [dynamically constructing](https://nextjs.org/docs/api-reference/next.config.js/introduction) the Next.js configuration sets the `i18n` property to `null`, this causes `withNextIntl` to incorrectly emit a warning about the use of Next.js' builtin `i18n` library.

Additionally `withNextIntl` incorrectly types the `i18nPath` as `string` when it should be `string | undefined`.

### DIFF
--- a/packages/next-intl/plugin.d.ts
+++ b/packages/next-intl/plugin.d.ts
@@ -1,3 +1,3 @@
 import {NextConfig} from 'next';
 
-export default function withNextIntl(i18nPath: string): ((config: NextConfig) => unknown);
+export default function withNextIntl(i18nPath?: string): ((config: NextConfig) => unknown);
--- a/packages/next-intl/plugin.js
+++ b/packages/next-intl/plugin.js
@@ -39,7 +39,7 @@ module.exports = withNextIntl({
     }
   }
 
-  if ('i18n' in nextConfig) {
+  if (nextConfig.i18n != null) {
     console.warn(
       "\nnext-intl has found an `i18n` config in your next.config.js. This likely causes conflicts and should therefore be removed if you use the React Server Components integration.\n\nIf you're in progress of migrating from the `pages` folder, you can refer to this example: https://github.com/amannn/next-intl/tree/feat/next-13-rsc/packages/example-next-13-with-pages\n"
     );


narakhan:
Fair warning, I was unable to install the development version of the package on my machine (`@napi-rs` doesn't export binaries for my host). So I was unable to run the tests locally, but I did verify the changes manually.

narakhan:
Also it might be a good idea to add a `contributing.md` to the project, didn't know about the PR formatting rules and the `lint-pr` errors didn't provide an example of a valid title.

packages/next-intl/plugin.d.ts
```diff
@@ -1,3 +1,3 @@
 import {NextConfig} from 'next';
 
-export default function withNextIntl(i18nPath: string): ((config: NextConfig) => unknown);
+export default function withNextIntl(i18nPath?: string): ((config: NextConfig) => unknown);
```
amannn, line 3: Good point, thanks!

packages/next-intl/plugin.js
```diff
@@ -39,7 +39,7 @@ module.exports = withNextIntl({
     }
   }
 
-  if ('i18n' in nextConfig) {
+  if (nextConfig.i18n != null) {
```
amannn, line 42: The change looks good, but I don't really understand what you mean by "dynamically constructing" the Next.js configuration. Can you elaborate?

amannn (approved):
Hey, thanks a lot for this PR—this looks great!

I've also just added a [`CONTRIBUTORS.md`](https://github.com/amannn/next-intl/blob/main/CONTRIBUTORS.md) file, been meaning to set that up for ages. Thanks for the reminder!

packages/next-intl/plugin.js
amannn: I verified the change locally, looks good!

narakhan: As explained in the [docs](https://nextjs.org/docs/api-reference/next.config.js/introduction), in your `next.config.js` module there are two supported configuration methods. Either the module exports as `default` `NextConfig` object, or a function accepting a `phase: string` (possible literal strings are found in `next/dist/shared/lib/constants`) and a `config: { defaultConfig: NextConfig }`.

The function export is useful for dynamically populating the Next.js configuration depending on the current `phase`. At my workplace we're using this functionality to control the headers Next.js serves, which is useful for selectively disabling the [content security policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) headers in the development environment (as it's a pain to get them working with `http://localhost` and we'd rather not forget to enable them for our production builds). We also perform validation of our environmental variables (using [zod](https://github.com/colinhacks/zod)) and add them to the `ProcessEnv` NodeJS typing namespace within this function, that way Next.js is prevented from starting up until the issue is resolved (as opposed to throwing these errors during runtime, which is easier to overlook and harder to debug).

The issue that part of the PR resolves is that Next.js provides a `defaultConfig` object with the `i18n` property initialised as `null`, since the `next-intl` plugin module was issuing warnings based on the presence of the field (rather than its value) it was incorrectly throwing out warnings.

The other issue was found as we enable type-checking in the `next.config.js` module. Might be worth adding a test to `next-intl` which also does this, otherwise it's unlikely you'll catch any issues with parts of the interface consumed by non-Typescript code. The process how to do this well documented within the Typescript documentation section on [type checking javascript files](https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html).

narakhan: Anyways sorry for the infodump, couldn't help myself 😊

amannn: Hey, thanks a lot for the extensive answer! That's what I thought is happening with the `phase`, but I've never used it personally! 🙂 

You can upgrade to `next-intl@2.14.0-beta.2` to get rid of the issue!

narakhan: Thank you very much, I can yeet my `delete config.i18n` line now!

PR #269, opened 2023-05-03
fix: Ability to modify request headers

### LINKED ISSUE #266: Modifying the request headers doesn't work
### Description

Using `next-intl@2.14.0-beta.3` version and following the example [withExtraMiddleware](https://github.com/amannn/next-intl/issues/247#issuecomment-1523292342) the request headers aren't modified.

https://github.com/ARochniak/next-intl-middleware.git

### Reproduction description

Steps to reproduce:
1. Clone reproduction repo
2. Run `yarn && yarn dev`
3. Open `https://localhost:3000` in the browser
4. Restart the page
5. You can see in the log message `X-test header: null`, but it should be `X-test header: test`

### Expected behaviour

According to the example [withExtraMiddleware](https://github.com/amannn/next-intl/issues/247#issuecomment-1523292342), request headers should be modified correctly.

ARochniak:
fix #266


packages/next-intl/src/middleware/middleware.tsx
```diff
@@ -58,8 +58,16 @@ export default function createMiddleware(config: MiddlewareConfig) {
     const hasUnknownHost = configWithDefaults.domains != null && !domain;
 
     function getResponseInit() {
-      // Nothing yet
-      return undefined;
+      if (hasOutdatedCookie) {
+        request.headers.set(HEADER_LOCALE_NAME, locale);
```
amannn, line 62: This is not necessary for the current implementation in the main branch. The header is the workaround that is currently used in the RSC branch to "send" the matched locale to Server Components. It's also the reason why SSG doesn't work at this point in the RSC branch, I hope to get rid of this once ServerContext works well.

In the main branch we currently fully support SSG and don't need this workaround, therefore we shouldn't introduce it here.

Can you remove this?

packages/next-intl/src/shared/constants.tsx
```diff
@@ -1,3 +1,6 @@
 // Reuse the legacy cookie name
 // https://nextjs.org/docs/advanced-features/i18n-routing#leveraging-the-next_locale-cookie
 export const COOKIE_LOCALE_NAME = 'NEXT_LOCALE';
+
+// Should take precedence over the cookie
+export const HEADER_LOCALE_NAME = 'X-NEXT-INTL-LOCALE';
```
amannn, line 6: See the comment above.

packages/next-intl/src/middleware/middleware.tsx
```diff
@@ -58,8 +58,16 @@ export default function createMiddleware(config: MiddlewareConfig) {
...
+      }
+      const responseInit = {
+        request: {
+          headers: request.headers
+        }
```
amannn, line 67: This is the fix that allows forwarding headers, right? Can you include a regression test in [middleware.test.tsx](https://github.com/amannn/next-intl/blob/main/packages/next-intl/test/middleware/middleware.test.tsx) that fails for the previous implementation and now passes?

amannn (commented):
Thank you very much for contributing this fix @ARochniak! 👏

packages/next-intl/src/middleware/middleware.tsx
ARochniak: @amannn Sorry, my bad! Sure I will remove it.

packages/next-intl/src/middleware/middleware.tsx
ARochniak: Yes, sure. But need some time, if it's fine for you.

packages/next-intl/src/middleware/middleware.tsx
amannn: No worries! This is unfortunately a bit convoluted currently.

Btw. I'm currently in process of modernizing the code base a bit, the main branch now uses pnpm/turbo. Feel free to rebase to get a performance boost during development! The [contributors guide](https://github.com/amannn/next-intl/blob/main/CONTRIBUTORS.md) is updated as well.

Next up is modernizing the test runner and build system …

ARochniak:
Hi @amannn, while I was adding test case I realised that the previous solution was not good enough as it implicitly set request headers, even in cases when it's not necessary. But the current implementation also looks not clean enough for me. I would like to hear you opinion.
P.s. Also I realised how long ago I wrote tests 😁

packages/next-intl/src/middleware/middleware.tsx
```diff
@@ -24,14 +24,15 @@ function receiveConfig(config: MiddlewareConfig) {
...
+type MiddlewareResponseInit = Parameters<typeof NextResponse.next>[0]
 export default function createMiddleware(config: MiddlewareConfig) {
   const configWithDefaults = receiveConfig(config);
 
   // Currently only in use to enable a seamless upgrade path from the
   // `{createIntlMiddleware} from 'next-intl/server'` API.
   const matcher: Array<string> | undefined = (config as any)._matcher;
 
-  return function middleware(request: NextRequest) {
+  return function middleware(request: NextRequest, _?: NextFetchEvent, responseInit?: MiddlewareResponseInit) {
```
amannn, line 35: Hmm, this doesn't look quite right. You're accepting a third parameter now for the middleware, but the goal is to keep existing incoming headers, right?

I think these tests could help to verify the implementation:

```tsx
it('retains request headers for the default locale', () => {
  middleware(
    createMockRequest('/', 'en', 'http://localhost:3000', undefined, {
      'x-test': 'test'
    })
  );
  expect(
    MockedNextResponse.rewrite.mock.calls[0][1].request.headers.get(
      'x-test'
    )
  ).toBe('test');
});

it('retains request headers for secondary locales', () => {
  middleware(
    createMockRequest('/de', 'de', 'http://localhost:3000', undefined, {
      'x-test': 'test'
    })
  );
  expect(
    MockedNextResponse.next.mock.calls[0][0].request.headers.get('x-test')
  ).toBe('test');
});
```

These tests are set up to be placed at the spot in the test suite where you've added your test!

I've added another parameter for `createMockRequest` here to accept `extraHeaders?: Record<string, string>`.

If the goal of the consumer is to modify request headers, this can be done [before the middleware is invoked](https://next-intl-docs.vercel.app/docs/next-13/middleware#composing-other-middlewares).

Does this help? Let me know if I can help!

amannn (commented):
Thanks for adding the test, hope you found your way around! 🙌 

Good point that working with test runners might not be that common, I've just updated the contributors guide to contain [some information in regard to running and writing tests](https://github.com/amannn/next-intl/blob/main/CONTRIBUTORS.md#tests). Hopefully this helps a bit!

packages/next-intl/src/middleware/middleware.tsx
ARochniak: @amannn My initial intention was to provide ability to set custom request headers that can be consumed within server components. Previous implementation did it by implicitly forwarding headers from incoming request. But my concern is that it could go wrong at some point and such explicit setting of the `responseInit` looks more robust and extendable. That is how I see adjusted `withExtraMiddleware` example:
``` typescript
import createMiddleware, { NextMiddleware } from 'next-intl/middleware';
import { NextMiddlewareResult } from 'next/dist/server/web/types';
import {NextFetchEvent, NextRequest, NextResponse} from 'next/server';

export default withExtraMiddleware(
  createMiddleware({
    locales: ['en', 'de'],
    defaultLocale: 'en'
  })
);

type MiddlewareResponseInit = Parameters<typeof NextResponse.next>[0]

function withExtraMiddleware(next: NextMiddleware) {
  return async (request: NextRequest, event: NextFetchEvent) => {
    // Step 1: Potentially change the incoming request
    request.headers.set('x-test', 'test');

    // Step 2: Call the nested next-intl middleware
    const response = next(request, event, { request: { headers: request.headers } });

    // Step 3: Potentially change the response
    if (response) {
      response.headers.set('x-test', 'test');
    }

    return response;
  };
}

export const config = {
  // Skip all paths that should not be internationalized
  matcher: ['/((?!_next|.*\\..*).*)']
};
```

Thank you for the tests!

amannn: Thanks for explaining @ARochniak!

> Previous implementation did it by implicitly forwarding headers from incoming request. But my concern is that it could go wrong at some point.

Do you have anything particular in mind that could go wrong? I think that by adding tests we can guarantee that this continues to work in the future.

Generally, if something can be solved with the existing APIs, I wouldn't want to add features. By doing this, we have to maintain less features within the library and also for the consumer there is a single way to achieve a particular goal (if we'd support a third argument, it's questionable what happens to existing request headers for example).




Therefore I'd really be in favor of not adding more API than necessary. Hope this makes sense!

ARochniak: @amannn Thanks for clarification. My concern is mostly based on some "God feeling" 😊. I haven't any issue about using previous approach, and I just wanted to discuss with you different option.

amannn (approved):
Looks great, thanks!

I've rebased your PR and fixed a lint issue. If tests pass, I'll merge this and I'll also publish a new RSC release rebased on top of this.

ARochniak:
> Looks great, thanks!
> 
> I've rebased your PR and fixed a lint issue. If tests pass, I'll merge this and I'll also publish a new RSC release rebased on top of this.

Thank you, it was a pleasure to contribute and communicate with you!

amannn:
Thanks, the pleasure's all mine! 🙂 

Here's the new RSC version (I think that's what you needed originally): `2.15.0-beta.3`

PR #328, opened 2023-06-12
refactor: Remove experimental config for appDir and @next/font

grzegorzpokorski:
Changes:
- Removed experimental config for appDir because starting from nextjs v13.4, this feature is stable (for now in the console at development mode we have some warns about that this config is necessary but it was fixed in the latest version of nextjs),
- Removed the `@next/font` package since it is now built into nextjs itself.

### DIFF
--- a/examples/example-next-13-next-auth/next.config.js
+++ b/examples/example-next-13-next-auth/next.config.js
@@ -1,1 +1,0 @@
-module.exports = {};
--- a/examples/example-next-13-next-auth/next.config.js
+++ b/examples/example-next-13-next-auth/next.config.js
@@ -1,3 +1,1 @@
-module.exports = {
-  experimental: {appDir: true}
-};
+module.exports = {};
--- a/examples/example-next-13/next.config.js
+++ b/examples/example-next-13/next.config.js
@@ -1,1 +1,0 @@
-module.exports = {};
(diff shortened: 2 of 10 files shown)


examples/example-next-13-next-auth/next.config.js
```diff
@@ -1,3 +1 @@
-module.exports = {
-  experimental: {appDir: true}
-};
+module.exports = {};
```
amannn, line 1: Maybe we can remove the config files entirely now if they're empty?

examples/example-next-13/next.config.js
```diff
@@ -1,3 +1 @@
-module.exports = {
-  experimental: {appDir: true}
-};
+module.exports = {};
```
amannn, line 1: Same here.

amannn (commented):
Many thanks for this PR, always great to see old cruft removed! 🙌

pnpm-lock.yaml
```diff
@@ -1,4 +1,8 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
```
amannn, line 1: I think the pnpm version needs to be updated in the GitHub workflows for this to work.

examples/example-next-13-next-auth/next.config.js
grzegorzpokorski: Good point, i didn't thought about it 😂

amannn (approved):
Many thanks! 🙌

PR #149, opened 2022-12-08
feat!: Next.js 13 RSC integration

amannn:
→ Please refer to [the announcement](https://next-intl-docs.vercel.app/blog/next-intl-3-0)

---

Fixes #488
Fixes https://github.com/amannn/next-intl/issues/402
Fixes https://github.com/amannn/next-intl/issues/578

PR #359, opened 2023-06-27
feat: Add type safety for timeZone parameter

A7med3bdulBaset:
This PR introduces type safety for the `timeZone` parameter instead of using `string` as the type. It enhances the developer experience by ensuring that only valid time zone values are accepted and providing better code completion and type-checking support.

## Changes Made:
- Create a `TimeZone` type that is a union of all valid time zone values in `/use-intl/src/core/utils/TimeZones.tsx`
- Update the `timeZone` parameter in these files:
   - `IntlConfig`
   - `DateTimeFormatOptions`
   - `createFormatter`
   - `convertFormatsToIntlMessageFormats`

## Data Source:
Time zone values sourced from [tz database](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones).


packages/use-intl/src/core/DateTimeFormatOptions.tsx
```diff
@@ -1,5 +1,7 @@
 // https://github.com/microsoft/TypeScript/issues/35865
 
+import { TimeZone } from "./utils/TimeZones";
```
amannn, line 3: Some nits:
1. You can move this out of `utils` to the same level where e.g. `DateTimeFormatOptions` is defined.
2. Please use a default export/import like in `DateTimeFormatOptions`.
3. Please make sure the file name matches the import name (the singular `TimeZone` is good)

packages/use-intl/src/core/utils/TimeZones.tsx
```diff
@@ -0,0 +1,422 @@
...
+	| "Pacific/Rarotonga"
+	| "Pacific/Tahiti"
+	| "Pacific/Tarawa"
+	| "Pacific/Tongatapu"
+	| "Pacific/Wake"
+	| "Pacific/Wallis"
+	| "UTC"
+	| "W-SU"
+	| "WET"
+	| "Zulu";
```
amannn, line 422: I'm wondering if we should allow unknown strings here too (see [this trick](https://stackoverflow.com/a/74467583/343045)).

The benefit of allowing unknown strings is that we don't have to publish a new version in case our list gets outdated. This might even be critical for users on outdated versions, if we've released breaking changes in the meantime, otherwise we'd have to backport changed time zones. AFAIK React does this too, e.g. for ARIA roles.

The downside is of course that we don't fail for invalid time zones.

However, JavaScript runtimes throw an error for unknown time zones:

Screenshot 2023-06-28 at 17 09 31

(I've supplied "Unknown" in this example)

Therefore, I'd say we should support unknown strings.

packages/use-intl/src/react/useTimeZone.tsx
```diff
@@ -1,5 +1,6 @@
+import { TimeZone } from '../core/utils/TimeZones';
 import useIntlContext from './useIntlContext';
 
-export default function useTimeZone() {
+export default function useTimeZone(): TimeZone {
```
amannn, line 4: The return type is wrong, also `undefined` is allowed. Better just let TypeScript infer the return type and don't annotate it additionally!

amannn (commented):
Hey and many thanks for this PR! This is a great DX improvement! 🙌 

I've added a few nits, also I think CI isn't passing yet. I've just approved the `main` workflow, please have a look in regard to failed linting and type checking!

A7med3bdulBaset:
All Done and Ready for Review
All CI checks passed! other than the Vercel deployment, which requires an authorized user to approve.

I ran locally and all tests passed.
```bash
pnpm build && pnpm lint && pnpm test
```

packages/next-intl/prettier.config.js
```diff
@@ -0,0 +1,9 @@
+/* eslint-env node */
+/* eslint-disable import/no-extraneous-dependencies */
+
+module.exports = {
+  singleQuote: true,
+  trailingComma: 'none',
+  bracketSpacing: false,
+  plugins: [require('prettier-plugin-tailwindcss')]
+};
```
amannn, line 9: Sorry, there was some information missing here in the contributors guide.

I've just added [a new section about code formatting](https://github.com/amannn/next-intl/blob/main/CONTRIBUTORS.md#code-formatting-eslint--prettier).

Can you remove the two prettier configs you've added?

amannn (approved):
Looks fantastic, thank you so much! 🙌 

I've left one minor note about the prettier configs. If you could remove them, we can merge this PR!

amannn:
Looks great, will be released in a minute!

Are you interested in adopting this in [the RSC support branch](https://github.com/amannn/next-intl/pull/149/files) as well (e.g. `i18n.ts`)?

A7med3bdulBaset:
It seems like Eslint doesn't have configurations for `.tsx` files

Alert

Also, In the `CONTRIBUTORS.md` file: 
>Alternatively, you can run ESLint via the command line:
>
>```sh
>pnpm eslint src --fix
>```

It should be:
```diff
- pnpm eslint src --fix
+ pnpm dlx eslint src --fix
```
I still propose to include config files for formatting, linting, and settings. In the final It's OPEN-source. Also, in the `.gitignore` file we should remove `.vscode` from ignored files.

PR #361, opened 2023-06-28
fix: public pages matcher in middleware

jonathanwilke:
I found that when "/" is not included in the public pages, the regex would not work.

### DIFF
--- a/docs/pages/docs/next-13/middleware.mdx
+++ b/docs/pages/docs/next-13/middleware.mdx
@@ -276,7 +276,9 @@ const authMiddleware = withAuth(
 
 export default function middleware(req: NextRequest) {
   const publicPathnameRegex = RegExp(
-    `^(/(${locales.join('|')}))?(${publicPages.join('|')})?/?$`,
+    `^(/(${locales.join('|')}))?(${publicPages.join('|')})${
+      publicPages.includes("/") ? "?" : "+"
+    }/?$`,
     'i'
   );
   const isPublicPage = publicPathnameRegex.test(req.nextUrl.pathname);


docs/pages/docs/next-13/middleware.mdx
```diff
@@ -276,7 +276,9 @@ const authMiddleware = withAuth(
 
 export default function middleware(req: NextRequest) {
   const publicPathnameRegex = RegExp(
-    `^(/(${locales.join('|')}))?(${publicPages.join('|')})?/?$`,
+    `^(/(${locales.join('|')}))?(${publicPages.join('|')})${
+      publicPages.includes("/") ? "?" : "+"
+    }/?$`,
```
amannn, line 281: > I found that when "/" is not included in the public pages, the regex would not work.

By it "would not work", you mean the regex matches `/`, even though it shouldn't, right?

I'm wondering if we could simplify the regex instead of introducing a condition. I think the `/?` at the end is currently only there to support optional trailing slashes. [Next.js removes these by default anyway](https://nextjs.org/docs/app/api-reference/next-config-js/trailingSlash), so I think we could remove that part.

I think that would fix your issue, right?

Could you update `examples/example-next-13-next-auth/src/middleware.tsx` in sync with this change, so we can verify the test via playwright?

Many thanks!

amannn (commented):
Hey @jonathanwilke, nice to see you here and thank you so much for the PR! 🙌

docs/pages/docs/next-13/middleware.mdx
jonathanwilke: I will try that!

amannn: Did you get a chance to try this out? Is this PR still relevant?

PR #369, opened 2023-06-29
feat: Add auto completion to `timeZone` in the Next 13 RSC branch

A7med3bdulBaset:
The same as #359 but for a different branch


amannn:
Sorry, I think there's been some misunderstanding. What I meant in https://github.com/amannn/next-intl/pull/359#issuecomment-1613355279 is, that the RSC branch has some further usage of `timeZone` (e.g. [in the `i18n.ts` config file](https://next-intl-docs.vercel.app/docs/configuration#time-zone)). The branch adds new code in the `packages/next-intl` folder that uses the time zone as well—that was what I'm referring to.

We can get your changes from https://github.com/amannn/next-intl/pull/359 into the RSC branch by rebasing, I just quickly did that.

However, I now saw that all the usage of `timeZone` in the branch automatically infer the new `TimeZone` type you've introduced due to shared types, so practically there's in fact nothing to do.

I should have been more precise and should have probably checked before, sorry for the confusion!

Thanks again for your help @A7med3bdulBaset!

A7med3bdulBaset:
> However, I now saw that all the usage of timeZone in the branch automatically infer the new TimeZone type you've introduced due to shared types, so practically there's in fact nothing to do.

That what I noticed.
You can close the PR

A7med3bdulBaset:
Can you take a look at #365

amannn:
Yep, I have it on my list, quite busy with work these days!

PR #380, opened 2023-07-05
docs: Add docs for RTL languages

amannn:
Based on https://github.com/amannn/next-intl/pull/365/

### DIFF
--- a/docs/pages/docs/usage/messages.mdx
+++ b/docs/pages/docs/usage/messages.mdx
@@ -278,3 +278,62 @@ Messages are always parsed and therefore e.g. for rich text you need to supply t
 </Callout>
 
 The value of a raw message can be any valid JSON value: strings, booleans, objects and arrays.
+
+## Right-to-left languages
+
+Languages such as Arabic, Hebrew and Persian use [right-to-left script](https://en.wikipedia.org/wiki/Right-to-left_script) (often abbreviated as RTL). For these languages, writing begins on the right side of the page and continues to the left.
+
+In addition to providing translated messages, proper RTL localization requires:
+
+1. Providing [the `dir` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir) on the document
+2. Support layout mirroring, e.g. by using [CSS logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values)
+3. Support element mirroring, e.g. by customizing icons
+
+To handle these cases in your components, you can set up a reusable hook:
+
+```tsx filename="useTextDirection.tsx"
+import detectRtl from 'rtl-detect';
+import {useLocale} from 'next-intl';
+
+export default function useTextDirection(locale) {
+  const defaultLocale = useLocale();
+  if (!locale) locale = defaultLocale;
+  return detectRtl(locale) ? 'rtl' : 'ltr';
+}
+```
+
+… and use it accordingly:
+
+```tsx filename="app/[locale]/layout.tsx"
+import useTextDirection from './useTextDirection';
+
+export default function Layout({children, params: {locale}}) {
+  const direction = useTextDirection(locale);
+
+  return (
+    <html lang={locale} dir={direction}>
+      {/* ... */}
+    </html>
+  );
+}
+```
+
+```tsx filename="components/Breadcrumbs.tsx"
+import {useTranslations} from 'next-intl';
+import useTextDirection from './useTextDirection';
+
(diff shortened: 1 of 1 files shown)


docs/pages/docs/usage/messages.mdx
```diff
@@ -278,3 +278,62 @@ Messages are always parsed and therefore e.g. for rich text you need to supply t
...
+To handle these cases in your components, you can set up a reusable hook:
+
+```tsx filename="useTextDirection.tsx"
+import detectRtl from 'rtl-detect';
+import {useLocale} from 'next-intl';
+
+export default function useTextDirection(locale) {
+  const defaultLocale = useLocale();
+  if (!locale) locale = defaultLocale;
+  return detectRtl(locale) ? 'rtl' : 'ltr';
```
A7med3bdulBaset, line 301: the default export of `rtl-detect` is not a function but an object of two functions `isRtlLang(local: string): boolean` and `getLangDir(local: string): "rtl" | "ltr"`.

The correct code will be:
```js
return detectRtl.isRtlLang(locale) // If we want to return a boolean
return detectRtl.getLangDir(locale) // If we want to return either `"rtl"` or `"ltr"`
```

docs/pages/docs/usage/messages.mdx
```diff
@@ -278,3 +278,62 @@ Messages are always parsed and therefore e.g. for rich text you need to supply t
...
+}
+```
+
+… and use it accordingly:
+
+```tsx filename="app/[locale]/layout.tsx"
+import useTextDirection from './useTextDirection';
+
```
A7med3bdulBaset, line 309: Since the file extension of `app/[locale]/layout.tsx` is `.tsx` we need to type the params:

```tsx
interface LayoutProps {
  children: React.ReactNode;
  params: {
    locale: string;
  }
}

export default function Layout({children, params: {locale}}: LayoutProps) {
// ...
```

docs/pages/docs/usage/messages.mdx
```diff
@@ -278,3 +278,62 @@ Messages are always parsed and therefore e.g. for rich text you need to supply t
 </Callout>
 
 The value of a raw message can be any valid JSON value: strings, booleans, objects and arrays.
+
+## Right-to-left languages
+
+Languages such as Arabic, Hebrew and Persian use [right-to-left script](https://en.wikipedia.org/wiki/Right-to-left_script) (often abbreviated as RTL). For these languages, writing begins on the right side of the page and continues to the left.
```
A7med3bdulBaset, line 284: Would it be beneficial to include a textual example in order to enhance clarity?
Like:
```txt
النص في اللغة العربية _مثلا_ يُقرأ من اليمين لليسار
```

Just to show the problem that we are solving

docs/pages/docs/usage/messages.mdx
```diff
@@ -278,3 +278,62 @@ Messages are always parsed and therefore e.g. for rich text you need to supply t
...
+3. Support element mirroring, e.g. by customizing icons
+
+To handle these cases in your components, you can set up a reusable hook:
+
+```tsx filename="useTextDirection.tsx"
+import detectRtl from 'rtl-detect';
+import {useLocale} from 'next-intl';
+
+export default function useTextDirection(locale) {
+  const defaultLocale = useLocale();
```
A7med3bdulBaset, line 299: I'm not sure if this hook will work as we expect if we use it in the layout file because it depends on the context that is provided inside the layout.

A7med3bdulBaset (changes requested):

docs/pages/docs/usage/messages.mdx
amannn: Oh right, good catch—thanks! Adapted it to the boolean check now and returned strings, so the function is self-documenting.

docs/pages/docs/usage/messages.mdx
amannn: `next-intl` currently doesn't take a stand in the docs on how props should be typed (I've seen many different preferences of developers). I think the Next.js docs do a good job of explaining this.

docs/pages/docs/usage/messages.mdx
amannn: Good idea, thanks! I was quite amazed that in the editor even the cursor keys immediately switch once you enter a RTL script!

docs/pages/docs/usage/messages.mdx
amannn: Yep, that's true. I added the optional `locale` param for this case!

A7med3bdulBaset (approved):

PR #381, opened 2023-07-05
fix: Bring back ESM build (and fix bundle size)

amannn:


packages/next-intl/package.json
```diff
@@ -18,26 +18,39 @@
     "postpublish": "git checkout . && rm ./README.md"
   },
   "main": "dist/index.js",
+  "module": "dist/index.mjs",
```
amannn, line 21: This is not used by Node.js, only by ESM-enabled bundlers.

packages/next-intl/package.json
```diff
@@ -18,26 +18,39 @@
...
   "typings": "dist/index.d.ts",
   "exports": {
     ".": {
       "react-server": {
+        "node": "./dist/react-server/index.js",
```
amannn, line 26: It's important that this goes first, so it has priority over `import` when compiled under Node.js.

We're not 100% ESM compatible: https://nodejs.org/api/esm.html#esm_mandatory_file_extensions

But we don't want to be ESM compatible anyway, because Next.js itself isn't, therefore the bundler will crash when it encounters imports from Next.js.

Rather, we want to stay in CJS land for Node.js and only use `import` when bundling for the browser (webpack).

Issue #393, opened 2023-07-11
using createMiddleware disables compression on files passing through middleware


mfrlic:
### Description

*Tested on a local build and Azure app service deployment, works as expected on Vercel*
next-intl version: **3.0.0-beta.7**
next version: **13.4.9**

Accept-encoding header can be seen in the end request, but in getResponseInit() it does not exist, probably because next strips forbidden headers in its own middleware. 

The end result: 

Any request that passes through middleware will have a request header Accept-Encoding: "gzip, deflate, br", but will not have Content-Encoding: "gzip" (Next.js default compression). Even when brotli was enabled on AFD (Azure Front Door), css or js files would be compressed, but RSC or html wouldn't.

### Mandatory reproduction URL (CodeSandbox or GitHub repository)

N/A, any app using next-intl createMiddleware can achieve this

### Reproduction description

Steps to reproduce:
1. Build any app using createMiddleware locally
2. Run npm start
3. Check response headers of requests passing through middleware, notice the lack of Content-Encoding

### Expected behaviour

Content is compressed correctly

amannn:
> Any request that passes through middleware will have a **request** header Accept-Encoding: "gzip, deflate, br", but will not have Content-Encoding: "gzip" (Next.js default compression).
> (…)
> Check **response** headers of requests passing through middleware, notice the lack of Content-Encoding

(emphasis mine)

Sorry, I have a hard time following your bug report. What is the issue you're seeing in your app? Is there a _request_ header missing that you're reading in one of your components or is there a _response_  header missing that is returned?

[The official Next.js App Router example](https://nextjs-template.vercel.app/) has the exact same response headers as the [next-intl App Router example](https://next-intl-example-next-13.vercel.app/) (with the one difference, that we return `Link` additionally). Among them is `Content-Encoding: br`.

Providing a reproduction is not optional, please use [the bug template](https://codesandbox.io/p/sandbox/next-intl-bug-template-forked-yow8ep) that is mentioned in the bug report template to illustrate the problem.

mfrlic:
Basically, when i run npm run build or deploy to azure, any request that passes through middleware is missing the response header Content-Encoding: gzip (it is not compressed), and has the request header Accept-Encoding: gzip, deflate, br. Setting the accept-encoding header in intl middleware like I did in https://github.com/mfrlic/next-intl solves the issue. 

This might be a bug in NextResponse aswell.

amannn:
Yep, I think this is a Next.js bug:
[...]

PR #394, opened 2023-07-11
FIX: added accept encoding header in getResponseInit()

mfrlic:
This fixes the following issue: https://github.com/amannn/next-intl/issues/393


amannn:
Thank you for taking the time to contribute! I'll close this based on https://github.com/amannn/next-intl/issues/393#issuecomment-1632720563.

PR #409, opened 2023-07-17
feat: Accept `locale` with `useRouter` APIs

### LINKED ISSUE #408: Navigation: simple function to get current url with different locale
Currently, it is possible to change the current locale by providing a [Link](https://next-intl-docs.vercel.app/docs/routing/navigation#link) with a different locale.
This can be combined with [usePathname](https://next-intl-docs.vercel.app/docs/routing/navigation#usepathname) to show a link in a different locale. See https://github.com/amannn/next-intl/discussions/278

For my project, I want to use a custom select component to change the locale. I cannot use the `Link` component in this situation. Instead, I want to change the locale through a function. So I tried to use [useRouter](https://next-intl-docs.vercel.app/docs/routing/navigation#userouter) but it does not have any option to pass the new locale to it.

It would be great to have a simple function to get the current url with a specific locale to simplify the development of a component to change the locale.

The documentation should contain some kind of solution because this question has been asked multiple times, mostly by people who do not understand that the single source of truth does rely on the current URL. See: https://github.com/amannn/next-intl/discussions/320, https://github.com/amannn/next-intl/discussions/407

amannn:
Closes #408
Ref #407, #320

### DIFF
--- a/docs/pages/docs/environments/error-files.mdx
+++ b/docs/pages/docs/environments/error-files.mdx
@@ -93,7 +93,7 @@ When an `error` file is defined, Next.js creates [an error boundary within your 
 
 Since the `error` file must be defined as a Client Component, you have to use [`NextIntlClientProvider`](/docs/configuration#client-server-components) to provide messages in case the `error` file renders.
 
-If you've [set up `next-intl` to be used in Client Components](http://localhost:3001/docs/getting-started/app-router-client-components), this is already the case and there's no additional setup needed. If you're using [the Server Components beta](/docs/getting-started/app-router-server-components) though, you have to provide the relevant messages in the wrapping layout.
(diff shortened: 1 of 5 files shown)


WilliamZimmermann:
This works like a charm for App Router. How about for Pages Router?

I mean, I'm trying to follow the example on the page for migration Pages/App Router (https://github.com/amannn/next-intl/tree/main/examples/example-app-router-migration), but when it refers to links, it redirects for something like [locale]/myPage instead give the right locale. Any suggestion?

amannn:
The Pages Router has this [built in](https://nextjs.org/docs/pages/api-reference/functions/use-router#routerpush).

WilliamZimmermann:
Thanks Amann. So I should do something like this: 

```
router.push("/", "", { locale: "en" });
```

I tested with your "example-app-router-migration" and it works fine. However, in our project (we are still migrating from Pages to App) it is not working.

In our project, the same "pattern" of URL is simple **removing** the "locale" from the URL instead to replace the [locale]. Probably some bad settings.

Thank you again for the clarification.

PR #435, opened 2023-08-01
fix: Handle case where the `locale` param is an array and set cookie expiration to one year

### LINKED ISSUE #431: Keep locale after session was closed

When I close and open my site, it resumes my language preferences, instead of keeping the language I had chosen recently.

Add the ability to keep the previous language.

The language is stored in a cookie whose age is the current session. You should fix that or store it in localStorage. Or have the possibility of saying which language to use from a user that is retrieved via database or from a token (Auth.js). That would be the best thing to do.

amannn:
Changes:
 - Handle case where the locale param is an array
 - Set cookie expiration to one year
 - Pin Next.js version in examples
 - Use `latest` tag in examples (for CSB)

Fixes #431

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -16,7 +16,7 @@
     "@vercel/analytics": "^1.0.1",
     "clsx": "^1.2.1",
     "http-status-codes": "^2.2.0",
-    "next": "^13.4.7",
+    "next": "^13.4.12",
     "nextra": "^2.8.0",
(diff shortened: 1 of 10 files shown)


aXenDeveloper:
Hi 👋
I see you're pinned version for `next`, but do you have plan to upgrade `next` to the latest version?
For me `not-found.tsx` file not working from `13.4.10+` version of `next`.

PR #436, opened 2023-08-01
docs: fix messages path

luzhenqian:


docs/pages/docs/getting-started/app-router-client-components.mdx
```diff
@@ -77,7 +77,7 @@ export function generateStaticParams() {
 export default async function LocaleLayout({children, params: {locale}}) {
   let messages;
   try {
-    messages = (await import(`../../messages/${locale}.json`)).default;
```
amannn, line 80: This matches the directory structure outlined at the top here, doesn't it?

https://next-intl-docs.vercel.app/docs/getting-started/app-router-client-components#getting-started

luzhenqian: Sorry, it was my mistake.

amannn: No worries!

FWIW the default for the `src` directory config in the Next.js CLI was switched to "Yes" now:

Screenshot 2023-08-02 at 09 48 36

So I think it would make sense to generally move to that.

PR #451, opened 2023-08-12
feat: add next-intl/link to be used in locale switcher in next 13 example

agvst:
improves locale switcher by making use of next-intl/link to be able to switch without losing current page


amannn (approved):
Many thanks! 🙌

PR #462, opened 2023-08-19
fix: Detect host and protocol for alternate links correctly when running behind a proxy

HHongSeungWoo:
We are currently utilizing a reverse proxy in front of Next.js.
`request.url` is returning an internal address instead of an external one.
Consequently, the rewrite functionality is not behaving as expected, and the `alternatelink` header is inadvertently exposing the internal address.

### DIFF
--- a/packages/next-intl/src/middleware/middleware.tsx
+++ b/packages/next-intl/src/middleware/middleware.tsx
@@ -7,7 +7,6 @@ import getAlternateLinksHeaderValue from './getAlternateLinksHeaderValue';
 import resolveLocale from './resolveLocale';
 import {
   getBestMatchingDomain,
-  getHost,
   getLocaleFromPathname,
   isLocaleSupportedOnDomain
 } from './utils';
@@ -33,9 +32,6 @@ export default function createMiddleware(config: MiddlewareConfig) {
   const matcher: Array<string> | undefined = (config as any)._matcher;
 
   return function middleware(request: NextRequest) {
-    // if Next.js behind a reverse proxy, it cannot get the correct host. ex) localhost, 172.x.x.x
-    // We need to set the host here, because Next.js doesn't do it for us
-    request.nextUrl.host = getHost(request.headers) ?? request.nextUrl.host;
     const matches =
       !matcher ||
       matcher.some((pattern) => request.nextUrl.pathname.match(pattern));
--- a/packages/next-intl/src/middleware/middleware.tsx
+++ b/packages/next-intl/src/middleware/middleware.tsx
@@ -7,6 +7,7 @@ import getAlternateLinksHeaderValue from './getAlternateLinksHeaderValue';
 import resolveLocale from './resolveLocale';
 import {
   getBestMatchingDomain,
+  getHost,
   getLocaleFromPathname,
   isLocaleSupportedOnDomain
 } from './utils';
@@ -32,6 +33,9 @@ export default function createMiddleware(config: MiddlewareConfig) {
   const matcher: Array<string> | undefined = (config as any)._matcher;
 
   return function middleware(request: NextRequest) {
+    // if Next.js behind a reverse proxy, it cannot get the correct host. ex) localhost, 172.x.x.x
+    // We need to set the host here, because Next.js doesn't do it for us
+    request.nextUrl.host = getHost(request.headers) ?? request.nextUrl.host;
     const matches =
       !matcher ||
       matcher.some((pattern) => request.nextUrl.pathname.match(pattern));
@@ -68,7 +72,10 @@ export default function createMiddleware(config: MiddlewareConfig) {
     }
 
     function rewrite(url: string) {
-      return NextResponse.rewrite(new URL(url, request.url), getResponseInit());
+      return NextResponse.rewrite(
+        new URL(url, request.nextUrl.href),
+        getResponseInit()
+      );
     }
 
     function next() {
@@ -76,7 +83,7 @@ export default function createMiddleware(config: MiddlewareConfig) {
     }
 
     function redirect(url: string, host?: string) {
-      const urlObj = new URL(url, request.url);
+      const urlObj = new URL(url, request.nextUrl.href);
 
       if (domainConfigs.length > 0) {
         if (!host) {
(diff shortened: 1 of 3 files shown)


HHongSeungWoo:
I'm concerned about trusting the host of the header, but if there are multiple external addresses on the same server, I can't think of any other solution :(

packages/next-intl/src/middleware/middleware.tsx
```diff
@@ -32,6 +33,9 @@ export default function createMiddleware(config: MiddlewareConfig) {
   const matcher: Array<string> | undefined = (config as any)._matcher;
 
   return function middleware(request: NextRequest) {
+    // if Next.js behind a reverse proxy, it cannot get the correct host. ex) localhost, 172.x.x.x
+    // We need to set the host here, because Next.js doesn't do it for us
+    request.nextUrl.host = getHost(request.headers) ?? request.nextUrl.host;
```
amannn, line 38: If possible, I'd like to avoid overwriting the incoming request. If the user [composes the middleware](https://next-intl-docs.vercel.app/docs/routing/middleware#composing-other-middlewares), this could lead to surprises.

packages/next-intl/src/middleware/middleware.tsx
```diff
@@ -68,15 +72,18 @@ export default function createMiddleware(config: MiddlewareConfig) {
     }
 
     function rewrite(url: string) {
-      return NextResponse.rewrite(new URL(url, request.url), getResponseInit());
+      return NextResponse.rewrite(
+        new URL(url, request.nextUrl.href),
+        getResponseInit()
+      );
```
amannn, line 78: Is it necessary to rewrite to the external host here (same for the redirect)?

In your PR comment you mentioned:
> Consequently, the rewrite functionality is not behaving as expected

Can you explain what is not behaving as expected? My impression would be that your Next.js app already receives requests on an internal host, why can't it handle rewrites to the internal address too?

That being said, your point is true that the alternate links should point to the external host. I'm just wondering if we could limit the fix to choosing the correct host in `getAlternateLinksHeaderValue`, not having to modify anything outside of that file.

What do you think?

amannn (commented):
Thank you so much for proposing this fix @HHongSeungWoo!

It's definitely a good idea to use the external host for the alternate links, please see my inline comment for details.

Can you please also include tests that would have failed previously and will now pass after the fix?

Thank you so much! 🙌

packages/next-intl/src/middleware/middleware.tsx
HHongSeungWoo: This seems to be a new issue arising in Next.js version ^13.4.13. Since it might not be a problem that needs to be addressed within this library, I will remove it for now.

HHongSeungWoo: In previous versions, the host of request.url was pointing to localhost, allowing for correct rewriting or redirection. However, it seems that the behavior has changed in version ^13.4.13.

packages/next-intl/src/middleware/middleware.tsx
HHongSeungWoo: I haven't read about what alternatelink does exactly, but I'm curious about why the domain needs to be included. Can't the desired functionality be achieved with just the pathname without including the domain?

amannn: As far as I know, the domain needs to be included: https://developers.google.com/search/docs/specialty/international/localized-versions#http

HHongSeungWoo:
스크린샷 2023-08-21 18-23-30
스크린샷 2023-08-21 18-22-31

before / after

packages/next-intl/src/middleware/middleware.tsx
amannn: Interesting. Do you have a link to the corresponding change or a bug report on the Next.js side by chance?

HHongSeungWoo: It seems like this might be related to https://github.com/vercel/next.js/issues/54133.

dzenand:
Is this somehow related to this issue? 
https://github.com/vercel/next.js/pull/54277

amannn:
This looks great! The only thing missing would be a regression test that helps to ensure that we keep this functionality in the future.

Thanks also for linking to the Next.js issue, I'll keep an eye on that, but seems like it should be addressed on the Next.js side.

HHongSeungWoo:
> This looks great! The only thing missing would be a regression test that helps to ensure that we keep this functionality in the future.
> 
> Thanks also for linking to the Next.js issue, I'll keep an eye on that, but seems like it should be addressed on the Next.js side.


I'm unsure about the necessary test cases. The request.url within the middleware is a string generated by Next.js. 
If a hostname is configured, it will be used otherwise, localhost will be used. Upon reviewing the code, it appears that the option trustHostHeader, which is not currently utilized, generates the URL with the request's host when set to true. However, since this isn't documented, it seems to be an unused option. These changes are independent of external configurations or code modifications, so I can't think of any regression cases unless there are changes to the HTTP protocol itself.

amannn:
@HHongSeungWoo I've added a commit where the now fixed behavior is handled within a single test case. I also noticed that we should consider the protocol as well and fixed that accordingly.

Thanks for noticing that we can use the `NextRequest` constructor instead of constructing a mock object, I've adapted the tests to use that consistently to avoid unnecessary redirection.

Would you like to have a last look? From my opinion this is good to go! 👍

amannn (approved):

HHongSeungWoo:
> @HHongSeungWoo I've added a commit where the now fixed behavior is handled within a single test case. I also noticed that we should consider the protocol as well and fixed that accordingly.
> 
> Thanks for noticing that we can use the `NextRequest` constructor instead of constructing a mock object, I've adapted the tests to use that consistently to avoid unnecessary redirection.
> 
> Would you like to have a last look? From my opinion this is good to go! 👍


Thank you for making the necessary corrections. It looks better to me as well. :)

CarbonC:
Hello @amannn ! This looks great, but please would it be possible to push this fix to the beta branch as well? Currently im stuck with the bug on the beta-10 :-/
Thanks for the great work

amannn:
@CarbonC Yep, definitely! I've got another release on the main branch coming, then I'll rebase the RSC branch and publish a new beta!

amannn:
@CarbonC `next-intl@3.0.0-beta.11` is out now with all recent improvements from the main branch!

PR #492, opened 2023-08-31
Feat: Export localized pathnames link props Typescript helper

jlalmes:
## Feat: Export localized pathnames link props Typescript helper

```ts
import { createLocalizedPathnamesNavigation, LocalizedPathnamesLinkProps } from 'next-intl/navigation'

const pathnames = {
  ...
}

const { Link } = createLocalizedPathnamesNavigation({ ..., pathnames })

type LinkProps = LocalizedPathnamesLinkProps<keyof typeof pathnames>
```

Screenshot 2023-08-31 at 23 08 42

Screenshot 2023-08-31 at 23 07 59

### DIFF
--- a/packages/next-intl/src/navigation/createLocalizedPathnamesNavigation.tsx
+++ b/packages/next-intl/src/navigation/createLocalizedPathnamesNavigation.tsx
@@ -15,13 +15,16 @@ import {
   HrefOrUrlObjectWithParams
 } from './utils';
 
-export type LocalizedPathnamesLinkProps<Pathname extends keyof Pathnames<Locales>> = Omit<
+export type LocalizedPathnamesLinkProps<
+  Locales extends AllLocales,
+  Pathname extends keyof Pathnames<Locales>
+> = Omit<
   ComponentProps<typeof BaseLink>,
   'href' | 'name'
 > & {
   href: HrefOrUrlObjectWithParams<Pathname>;
   locale?: Locales[number];
-}
+};
 
 export default function createLocalizedPathnamesNavigation<
   Locales extends AllLocales,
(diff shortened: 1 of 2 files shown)


packages/next-intl/src/navigation/createLocalizedPathnamesNavigation.tsx
```diff
@@ -15,6 +15,17 @@ import {
...
+export type LocalizedPathnamesLinkProps<
+  Locales extends AllLocales,
+  Pathname extends keyof Pathnames<Locales>
+> = Omit<
+  ComponentProps<typeof BaseLink>,
+  'href' | 'name'
+> & {
+  href: HrefOrUrlObjectWithParams<Pathname>;
+  locale?: Locales[number];
+};
```
amannn, line 27: Hey, thank you so much for the proposal!

From your code snippet in the PR comment:

```tsx
type LinkProps = LocalizedPathnamesLinkProps<keyof typeof pathnames>
```

I noticed as well that the type of the pathnames need to be passed to be able to compose the `href` prop in a wrapping component.

This is briefly mentioned in [the docs](https://next-intl-docs.vercel.app/docs/routing/navigation#link):

Screenshot 2023-09-01 at 08 47 38

From my perspective the `LocalizedPathnamesLinkProps` helper doesn't do much more than `ComponentProps<typeof Link<Pathname>>`.

I tend to be on the side of not exposing more API than necessary if something can already be achieved somewhat reasonably.

Would that work for you? Can something be improved in regard to the docs?

On a side note, it would of course be fantastic if the `href` can be composed without passing the type argument again, but I couldn't find a way so far.

PR #493, opened 2023-08-31
fix: Allow unknown routes in localized pathnames navigation APIs

jlalmes:
Very annoying! This should be (at most) logged as a warning.

Screenshot 2023-08-31 at 22 55 55

### DIFF
--- a/packages/next-intl/src/navigation/createLocalizedPathnamesNavigation.tsx
+++ b/packages/next-intl/src/navigation/createLocalizedPathnamesNavigation.tsx
@@ -118,7 +118,7 @@ export default function createLocalizedPathnamesNavigation<
     };
   }
 
-  function usePathname(): keyof PathnamesConfig | (string & {}) {
+  function usePathname(): keyof PathnamesConfig {
     const pathname = useBasePathname();
     const locale = useTypedLocale();
     return getRoute({pathname, locale, pathnames});
--- a/packages/next-intl/src/navigation/createLocalizedPathnamesNavigation.tsx
+++ b/packages/next-intl/src/navigation/createLocalizedPathnamesNavigation.tsx
@@ -118,7 +118,7 @@ export default function createLocalizedPathnamesNavigation<
     };
   }
 
-  function usePathname(): keyof PathnamesConfig {
+  function usePathname(): keyof PathnamesConfig | (string & {}) {
     const pathname = useBasePathname();
     const locale = useTypedLocale();
     return getRoute({pathname, locale, pathnames});
--- a/packages/next-intl/src/navigation/utils.tsx
+++ b/packages/next-intl/src/navigation/utils.tsx
@@ -158,12 +158,12 @@ export function getRoute<Locales extends AllLocales>({
   pathname: string;
   pathnames: Pathnames<Locales>;
 }) {
-  pathname = unlocalizePathname(pathname, locale);
+  const unlocalizedPathname = unlocalizePathname(pathname, locale);
 
   let template = Object.entries(pathnames).find(([, routePath]) => {
     const routePathname =
       typeof routePath !== 'string' ? routePath[locale] : routePath;
-    return matchesPathname(routePathname, pathname);
+    return matchesPathname(routePathname, unlocalizedPathname);
   })?.[0];
 
   if (!template) {
--- a/packages/next-intl/src/navigation/utils.tsx
+++ b/packages/next-intl/src/navigation/utils.tsx
@@ -177,5 +177,5 @@ export function getRoute<Locales extends AllLocales>({
     }
   }
 
-  return template as keyof Pathnames<Locales>;
+  return template as keyof Pathnames<Locales> | (string & {});
 }
--- a/packages/next-intl/src/navigation/utils.tsx
+++ b/packages/next-intl/src/navigation/utils.tsx
@@ -177,5 +177,5 @@ export function getRoute<Locales extends AllLocales>({
     }
   }
 
-  return template as keyof Pathnames<Locales> | (string & {});
+  return template as keyof Pathnames<Locales>;
 }
--- a/packages/next-intl/src/navigation/utils.tsx
+++ b/packages/next-intl/src/navigation/utils.tsx
@@ -96,15 +96,16 @@ export function compileLocalizedPathname<Locales extends AllLocales, Pathname>({
   query?: Record<string, SearchParamValue>;
 }) {
   function getNamedPath(value: keyof typeof pathnames) {
-    const namedPath = pathnames[value];
+    let namedPath = pathnames[value];
     if (!namedPath) {
-      throw new Error(
-        process.env.NODE_ENV !== 'production'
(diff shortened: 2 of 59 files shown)


jlalmes:
The more I think about this current behaviour the more I dislike it. To be clear I am absolutely LOVING this new `createLocalizedPathnamesNavigation` feature, I just think the API needs some DX tidy-up.

The current implementation assumes that you have input **every** valid URL (pages, routes or other) your NextJS site will ever interact with in the pathnames config. Not to mention external links. I think is an unreasonable requirement.

See the following example:
Screenshot 2023-09-01 at 02 11 20

So yes I could do the following, but there will still always be edge cases we cannot catch:

```ts
if (href.startsWith('/api') || href.startsWith('http') || href.startsWith('mailto')) {
 return <NextLink href={href} />
} else {
 return <IntlLink href={href} />
}
```

Wouldn't it be great if we could just use the localized navigation APIs for everything?!

Will update this PR now.

jlalmes:
So you can now see that although `/unknown` URLs do not cause a Typescript error, the autocomplete is still available.

Screenshot 2023-09-01 at 02 40 59

Dynamic URLs will still throw a Typescript error if not specified via urlObject with params.

Screenshot 2023-09-01 at 02 43 22

jlalmes:
Let me know what you think, but this feels like we're moving in the right direction 🚀

packages/next-intl/src/navigation/createLocalizedPathnamesNavigation.tsx
```diff
@@ -118,7 +118,7 @@ export default function createLocalizedPathnamesNavigation<
     };
   }
 
-  function usePathname(): keyof PathnamesConfig {
+  function usePathname(): keyof PathnamesConfig | (string & {}) {
```
amannn, line 121: Hmm, this is an interesting one! I think there are some pros and cons here.

**Pro**

Using a unified API for all links, even external ones etc.

**Con**

Not enforcing a valid route. One of the use cases I had in mind while working on the feature is that it should be seamless to upgrade from shared pathnames to localized pathnames. A benefit of the current API is that you'll get compile-time errors for all places in your code base where you have to adjust a link.

I think there will undoubtedly be users asking for strict validation, so if we decide to allow arbitrary strings, we need a strong argument here.

---

My current perspective is that we could document that a user can use something like `<Link pathname={'/unknown' as any} />` for cases where it's intended to use an unknown pathname. It'd retain the strict typing behavior, but the user can opt-in to link to an unknown route for the cases where this is necessary, allowing for a unified API.


What do you think?

packages/next-intl/src/navigation/utils.tsx
```diff
@@ -96,15 +96,16 @@ export function compileLocalizedPathname<Locales extends AllLocales, Pathname>({
...
-      throw new Error(
-        process.env.NODE_ENV !== 'production'
-          ? `No route found for "${value}". Available routes: ${Object.keys(
-              pathnames
-            ).join(', ')}`
-          : undefined
-      );
+      namedPath = value;
+      if (process.env.NODE_ENV !== 'production') {
+        console.warn(
```
amannn, line 103: Throwing an error is definitely too much, thanks for raising this point. A warning is better, but I'm wondering if we should just process the `pathname` as-is in this case, without any message to the user? Wouldn't that be necessary when you pass unknown strings to the navigation APIs?

amannn (commented):
Thank you so much for this proposal! 🙌 

I think there are two aspects to this, I've added two comments so we can discuss them separately.

packages/next-intl/src/navigation/createLocalizedPathnamesNavigation.tsx
jlalmes: > I think there will undoubtedly be users asking for strict validation, so if we decide to allow arbitrary strings, we need a strong argument here.

Hmm. I don't think it will be common, but I do understand why someone might want this.

> `<Link pathname={'/unknown' as any} />`

I dislike this.

---

Ok, so I think we should keep the change from `throw Error` to `console.warn`. I also think that we should add a note to the docs to encourage users to do the following:

```ts
export const { ... } = createLocalizedPathnamesNavigation({
  locales,
  // 👇 allows any URL & maintains autocomplete
  pathnames: pathnames as typeof pathnames & Record<string & {}, string>,
});
```

It's not perfect, but I think this satisfies both use-cases.

packages/next-intl/src/navigation/utils.tsx
jlalmes: Agreed.

1. If strict, we want to warn user of mistake
2. If allow any URL, we don't want to break the app.

As I mentioned in the above comment, `console.warn` isn't perfect, but I think it satisfies both use-cases.

packages/next-intl/src/navigation/createLocalizedPathnamesNavigation.tsx
amannn: Oh that's a good idea! Might be a good fit for [the introductory section about localized pathnames](https://next-intl-docs.vercel.app/docs/routing/navigation#localized-pathnames).

packages/next-intl/src/navigation/utils.tsx
amannn: I'm just worried that even `console.warn` creates unnecessary noise if the user intentionally passes an unknown pathname. E.g. in an app like yours, wouldn't it be annoying to see a warning every time you're redirected to `/api/v1/billing/checkout` during development?

I think at this point it's quite safe to assume that most developers use TypeScript, so I think we can rely a bit more on the types for catching errors. If you want to globally opt-in to allowing unknown routes, you can modify the types for `pathnames` as shown above and if you want more granular control, you can still `// @ts-expect-error` for individual cases.

packages/next-intl/src/navigation/createLocalizedPathnamesNavigation.tsx
amannn: As a side note, we can probably also take some inspiration from how Next.js implements [statically typed routes](https://nextjs.org/docs/app/building-your-application/configuring/typescript#statically-typed-links). Their implementation is currently marked as experimental, so I think we should consider changes from their side, but based on the docs it seems like they currently show a TypeScript error for unknown routes. I haven't experimented with their implementation yet though.

jlalmes:
@amannn I have updated the PR to reflect our discussion - lmk what you think!

amannn (commented):
@jlalmes Looks really good!

I took the liberty to rephrase the docs a bit and move this to a `details` block. I hope that's ok, let me know if you have feedback!

Would you mind adding tests for `Link`, `useRouter`, `usePathname` and `redirect` in case an unknown route is encountered?

jlalmes:
@amannn - docs look great, I have added tests for each + typing tests.

amannn:
@jlalmes Looks great, thanks! I added a commit that fixes some linter issues. Once CI passes, I'll merge and release this. 

Thank you for the great PR! 🙌

amannn:
Out in `3.0.0-beta.16`!

Issue #396, opened 2023-07-12
Statically types links for `next-intl/link`


amannn:
### Is your feature request related to a problem? Please describe.

[Next.js now supports statically typed links](https://nextjs.org/blog/next-13-2#statically-typed-links), we should support this too in [`next-intl/link`](https://next-intl-docs.vercel.app/docs/routing/navigation#link). [Also rewrites/redirects should be supported now](https://twitter.com/shuding_/status/1629116730621538304).

### Describe the solution you'd like

Compose existing types from Next.js for static typing to work

### Describe alternatives you've considered

Leaving the routes untyped as `string`. Also [named routes](https://github.com/amannn/next-intl/tree/feat/next-13-rsc/examples/example-next-13-named-routes) provide type safety.

Aperrix:
up pls ?

amannn:
I'm currently working on pathname localization in https://github.com/amannn/next-intl/pull/426 (see details in https://github.com/amannn/next-intl/issues/402). This will introduce statically typed links when using this feature.

However it would still be cool if we can statically validate links when not using localized pathnames, so I'll leave this issue open for now.

tacomanator:
Since the two issues referenced are closed, can I assume this is possible using the localized pathnames feature now? I could not find the right docs, and so far it's not working for me so want to make sure I'm not trying to do something that's not possible.

`typedRoutes` is enabled in next config, and using the `<Link />` component from `createSharedPathnamesNavigation` as described in the [docs](https://next-intl-docs.vercel.app/docs/routing/navigation#shared-pathnames):

```
import { locales } from "@/i18n";
import { createSharedPathnamesNavigation } from "next-intl/navigation";

export const localePrefix = "always"; // Default

export const { Link, redirect, usePathname, useRouter } =
createSharedPathnamesNavigation({ locales, localePrefix });
```

but the Link exported from here does not give me type errors bad urls:

```
import { Link } from "@/navigation";

export default function Index() {
  const t = useTranslations("Index");
  const f = useFormatter();

  return (
    <>
      <h1 className="text-xl">{t("title")}</h1>
      <h2>CURRENCY: {f.number(1234, "currency")}</h2>
      <h2>DATE: {f.dateTime(new Date(), "wide")}</h2>
      <h2>LIST: {f.list(["a", "b", "c"], "enumeration")}</h2>
      <IntlLink href="/asldkfj">INTL LINK</IntlLink>
      <ul>
        <li>
[...]

PR #495, opened 2023-09-01
Proposal for a Webpack Plugin to Support Next.js typedRoute

HHongSeungWoo:
Related issue: #396

How about using a webpack plugin to support Next.js typedRoute? this PR includes test code for the proposal.

### DIFF
--- a/examples/example-next-13/next.config.mjs
+++ b/examples/example-next-13/next.config.mjs
@@ -1,0 +1,64 @@
+import webpack from "webpack"
+
+const ROUTE_DECLARATION_FILE = "../types/link.d.ts"
+class TypedRoutePlugin {
+
+  apply(compiler) {
+      compiler.hooks.compilation.tap("TypedRoutePlugin", (compilation) => {
+          compilation.hooks.processAssets.tapAsync(
+              {name: "TypedRoutePlugin", stage: webpack.Compilation.PROCESS_ASSETS_STAGE_OPTIMIZE_TRANSFER},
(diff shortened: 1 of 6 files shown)


examples/example-next-13/next.config.mjs
```diff
@@ -0,0 +1,64 @@
...
+      OriginalLinkProps,
+      "href"
+  >;
+
+  export type LinkProps<RouteInferType> = LinkRestProps & {
+    /**
+     * The path or URL to navigate to. This is the only required prop. It can also be an object.
+     * @see https://nextjs.org/docs/api-reference/next/link
+     */
+    href: __next_route_internal_types__.RouteImpl<RouteInferType> | UrlObject;
```
amannn, line 36: Oh wow, thanks for proposing this!

I haven't experimented with typed routes from Next.js yet, but based on the docs, shouldn't this be achievable by composing `Route` from `next`? See [the docs](https://nextjs.org/docs/app/building-your-application/configuring/typescript#statically-typed-links).

Btw. we do already have a webpack plugin in https://github.com/amannn/next-intl/pull/149 that creates an alias for the RSC config. If a webpack plugin is the only way to achieve this, we could integrate with that.

HHongSeungWoo: Thank you for the review.
I appreciate your feedback. It seems that I missed some important details in the documentation.
At this point, it appears that using Route directly is a much better approach.
I look forward to contributing again once typedRoute stabilizes. Thank you.

PR #501, opened 2023-09-08
docs: Update next-auth docs to narrow matcher regex

kieranm:
Currently the regex marks the index route `/` as public even if it's not included in the public pages array. Making the page a required component ensures this won't happen.

### DIFF
--- a/docs/pages/docs/routing/middleware.mdx
+++ b/docs/pages/docs/routing/middleware.mdx
@@ -362,7 +362,7 @@ const authMiddleware = withAuth(
 
 export default function middleware(req: NextRequest) {
   const publicPathnameRegex = RegExp(
-    `^(/(${locales.join('|')}))?(${publicPages.join('|')})?/?$`,
+    `^(/(${locales.join('|')}))?(${publicPages.join('|')})/?$`,
     'i'
   );
   const isPublicPage = publicPathnameRegex.test(req.nextUrl.pathname);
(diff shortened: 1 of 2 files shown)


amannn (commented):
Many thanks for the PR @kieranm!

I think this is what @jonathanwilke also observed in https://github.com/amannn/next-intl/pull/361, but the PR was discontinued.

Would you mind updating the [middleware matcher in `example-next-13-next-auth`](https://github.com/amannn/next-intl/blob/main/examples/example-next-13-next-auth/src/middleware.tsx) with the proposed regex? There are some tests that can validate the change.

Thank you!

**EDIT:** I took the liberty to do this in https://github.com/amannn/next-intl/pull/501/commits/54be263fade55524c242ea116984254292d54554

amannn:
Hmm, no; seems to fail some tests: https://github.com/amannn/next-intl/actions/runs/6198929005/job/16830397571?pr=501. 

I can also reproduce this locally. When navigating to http://localhost:3000/de, I get redirect to http://localhost:3000/login?callbackUrl=%2Fde (i.e. `/de` is not detected as a public path).

kieranm:
Hmm this is weird. The root path in our app isn't public so I haven't ran into this. I have a very hectic week but will try to find time to look into this.

amannn:
This was fixed in https://github.com/amannn/next-intl/pull/637!

PR #506, opened 2023-09-14
fix(RSC): Always read header in RSC render

amannn:
Should fix https://github.com/amannn/next-intl/discussions/446

As far as I understand the issue currently, something like this happens:
1. The middleware receives a request where the `NEXT_LOCALE` cookie is set
2. During the RSC render, the `NEXT_LOCALE` cookie is suddenly not available

If this is the case, this would fix the issue.

If however the middlware doesn't run at all for some reason, this will not make a difference. However, since this doesn't seem to have any downsides (possible since newer Next.js versions), this might simplify the code anyway and is a welcome change.

### DIFF
--- a/packages/next-intl/src/middleware/middleware.tsx
+++ b/packages/next-intl/src/middleware/middleware.tsx
@@ -119,20 +119,9 @@ export default function createMiddleware<Locales extends AllLocales>(
     const hasUnknownHost = configWithDefaults.domains != null && !domain;
 
     function getResponseInit() {
-      const responseInit = {
-        request: {
-          headers: request.headers
-        }
-      };
-
-      if (hasOutdatedCookie) {
-        responseInit.request.headers = new Headers(
-          responseInit.request.headers
-        );
-        responseInit.request.headers.set(HEADER_LOCALE_NAME, locale);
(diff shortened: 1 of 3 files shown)


carlosmca44:
At this moment I get a new error that I don't know if it has to do with nextjs 13.5.4 or with next-intl. The thing is that in all the tutorials that I have seen and even documentation, everywhere in the rootlayout they use the useLocale() hook and no one has a problem, it does give me a problem since it is a hook and the hooks cannot be server side
The error shown is the following:

```bash
 ✓ Compiled in 761ms (1242 modules)
 ⨯ node_modules/next-intl/dist/esm/server/RequestLocale.js (1:580) @ eval
 ⨯ Error: Unable to find `next-intl` locale, have you configured the middleware?`
    at RootLayout (./src/app/[locale]/layout.tsx:29:73)
    at async Promise.all (index 0)
null

```

My configuration is to the letter with the documentation and on par with other programmers, but no one else gets the same error as me, or at least for the same reason.
I will be really thankful if you or anybody can help me with this problem.

matee-safi:
I get the same error as well

PR #514, opened 2023-09-20
fix: Use correct port for alternate links when running behind a proxy…

iza-w:
When the application runs behind a proxy, the `x-forwarded-host` header contains the correct host and port information. However, the port in the `request.url` may not match the actual port used by the proxy. By overriding the port with an empty string, we ensure that alternate links are generated correctly, reflecting the proxy's host and eliminating potential inconsistencies.


amannn (commented):
Great point, thank you so much for this PR @iza-w!

I took the liberty to add another test and tweak the handling slightly.

Will merge this as soon as CI is green.

Issue #290, opened 2023-05-12
Unable to deploy on Netlify with latest packages


KrustyC:
### Description

I created a test Repo to report the problem. I am not sure whether this is dependant on `next-intl` or Netlify but I believe the new middleware created here certainly plays a role in this issue. 

When I run [this repo](https://github.com/KrustyC/next-netlify-test) locally with `netlify dev` all looks fine until I upgrade `@netlify/plugin-nextjs` to "4.36.1", at which point as soon as I run `netlify dev` I start getting an error "middleware is not a function" with the following stack trace 
```
TypeError: middleware is not a function
    at handler (file:///PATH/TO/REPO/repo/sites/next-netlify-test/.netlify/edge-functions/next-dev/index.js:78:28)
    at async FunctionChain.runFunction (https://64523ab4e7865600087fc3df--edge.netlify.app/bootstrap/function_chain.ts:390:22)
    at async FunctionChain.run (https://64523ab4e7865600087fc3df--edge.netlify.app/bootstrap/function_chain.ts:310:20)
    at async handleRequest (https://64523ab4e7865600087fc3df--edge.netlify.app/bootstrap/handler.ts:80:22)
    at async Server.#respond (https://deno.land/std@0.170.0/http/server.ts:299:18)
```

If I downgrade it seems to work just fine. As I said above I am not sure where the problem lies here, but I thought it may be useful to report it here for your awareness as I believe there may be quite a few people using this library and deploying their app on Netlify. If you want to test this out you can fork my repo and test it locally, following the README.

If you need anything else from me, please let me know.

### Mandatory reproduction URL (CodeSandbox or GitHub repository)

https://github.com/KrustyC/next-netlify-test

### Reproduction description

Steps to reproduce:
.1 Install the modules
2. Run `npm i @netlify/plugin-nextjs@latest next@latest`
3. Run `netlify dev`
4. Access http://localhost:8888 on your browser

### Expected behaviour

The page should load and the website should not break.

amannn:
Hey @KrustyC, thanks for adding this report!

I haven't used Netlify for deployment yet and haven't seen this error so far. As far as I understand, the latest `@netlify/plugin-nextjs` version leads to the issue, with no changes to `next-intl`—right? That sounds like it might be a bug on the Netlify side then.

I saw you also added a bug report in the Netlify repo: https://github.com/netlify/next-runtime/issues/2106. I'll follow along there in case I can provide help from the `next-intl` side!

[...]

PR #564, opened 2023-10-13
Expose rewrite option as a config

cprussin:
@amannn See comment [here](https://github.com/amannn/next-intl/pull/149#issuecomment-1761646354)

In short, this PR adds the ability to specify how to handle a rewrite, and allows the rewrite to be an async function.  The use case here is for netlify, which [does not properly support NextResponse.rewrite](https://answers.netlify.com/t/next-js-internationalized-site-error-too-many-redirects/96654) and requires use of the async [rewrite function from their custom Middleware wrapper library](https://github.com/netlify/next-runtime/blob/e9bbc283a40fadef42bcb3e98cc78ef14f0e7e0c/packages/next/src/middleware/request.ts#L69-L76).

Without something like this, trying to use next-intl on netlify results in all rewritten paths turning into an infinite redirect loop.

### DIFF
--- a/packages/next-intl/src/middleware/NextIntlMiddlewareConfig.tsx
+++ b/packages/next-intl/src/middleware/NextIntlMiddlewareConfig.tsx
@@ -1,3 +1,4 @@
+import type {NextRequest, NextResponse} from 'next/server';
 import {AllLocales, Pathnames} from '../shared/types';
 
 type LocalePrefix = 'as-needed' | 'always' | 'never';
@@ -25,6 +26,11 @@ export type DomainConfig<Locales extends AllLocales> = Omit<
   locales?: RoutingBaseConfig<Array<Locales[number]>>['locales'];
 };
 
+type AsyncRewriteFn = (
+  request: NextRequest,
+  ...args: Parameters<typeof NextResponse.rewrite>
+) => Promise<ReturnType<typeof NextResponse.rewrite>>;
+
 type MiddlewareConfig<Locales extends AllLocales> =
   RoutingBaseConfig<Locales> & {
     /** Can be used to change the locale handling per domain. */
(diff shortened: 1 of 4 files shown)


amannn:
Hi @cprussin and thanks for proposing this fix!

So as far as I understand, this is an area where Netlify doesn't correctly handle something that works in plain Next.js, is that correct?

We had a similar issue previously: https://github.com/amannn/next-intl/issues/290. A Netlify employee was quite helpful there in fixing the incorrect behavior on their side. I guess this also makes the most sense to avoid other users running into the issue. Something that works during development but then breaks in production is really not desirable.

Would you like to raise a bug report on the Netlify side?

amannn:
Btw. in regard to your fork: I can really recommend [patch-package](https://www.npmjs.com/package/patch-package) for such temporary fixes.

cprussin:
hey @amannn thanks a bunch for the reply, and for suggesting patch-package!

After some more investigation I'm less convinced that the issue is that netlify doesn't support `NextResponse.rewrite` and more convinced that there's something broken in the way they handle _any_ rewrite that involves static files.  My app at work is migrating from a static site generator to next, so we have a lot of static files and I thought it was a more general restriction with Netlify due to how they've extended how middleware function.  As it turns out, when interacting with static files, both `NextResponse.rewrite` and netlify's `MiddlewareRequest.rewrite` result in broken behavior (although broken in different ways).

I created minimal repros and opened tickets with Netlify for both issues, if you're interested you can follow along [here](https://github.com/netlify/next-runtime/issues/2331) for the issue with `MiddlewareRequest.rewrite`, and [here](https://github.com/netlify/next-runtime/issues/2332) for the issue with `NextResponse.rewrite`.

I can close this ticket, but I could also potentially see some value in exposing hooks into some of this kind of functionality (maybe someone wants to add some logging or something).  Do you think it would be valuable to keep this open in case you wanted functionality like this in the future?

amannn:
Thank you for investigating! I'll keep an eye on the Netlify issues then, hopefully there's a response soon!

In regard to the middleware adaption here: I largely prefer to not introduce more configuration options, especially if they might not be needed. A lot of use cases like logging can already be handled by [composing the middleware](https://next-intl-docs.vercel.app/docs/routing/middleware#composing-other-middlewares). In case there's a need for this option, we could always revive the PR!

I hope that makes sense! 🙂

cprussin:
Makes sense, thanks a bunch for the assistance here.  Will close this PR.

Issue #565, opened 2023-10-14
Support custom unit option of `relativeTime`


ktmouk:
### Is your feature request related to a problem? Please describe.

`format.relativeTime` currently only supports returning a relative time using an automatically calculated unit in [this function](https://github.com/amannn/next-intl/blob/70863eefc7c653624ed3f0dadb37eb4f50be1975/packages/use-intl/src/core/createFormatter.tsx#L15C11-L15C11). 
However, there may be cases that we sometimes want to use a specific unit to develop an application.

For example, if we were to create a timeline UI that groups photos by date in a photo album app, we might want to specify the `day` unit when creating a title for the timeline, since, with other units, it might get too detailed.

### Describe the solution you'd like

One solution I came up with is to add a `unit` argument to [relativeTime](https://github.com/amannn/next-intl/blob/70863eefc7c653624ed3f0dadb37eb4f50be1975/packages/use-intl/src/core/createFormatter.tsx#L157-L160) function, like this:

```js
const dateTime = new Date('2020-11-20T00:00:00Z');
const now = new Date('2020-11-20T01:00:00Z');

const format = useFormatter();
format.relativeTime(dateTime, now, 'days');  // returns 'today', not '1 hours'
```

### Describe alternatives you've considered

We can also use `RelativeTimeFormat` directly instead:

```js
console.log(new Intl.RelativeTimeFormat('en', { numeric: 'auto' }).format(0, 'day')); // 'today'
```

amannn:
Seems like a good enhancement, yes!

API-wise I have two thoughts:
1. Maybe we should accept an object, in case we allow to configure further properties in the future. E.g. `{unit: 'day'}`.
2. `now` is an optional parameter, maybe we should overload the second argument to avoid something like `format.relativeTime(dateTime, undefined, {unit: 'day'})`.

Based on this, `format.relativeTime(dateTime: Date, nowOrOptions: Date | {now?: Date, unit?: string})` could be an idea.

What do you think? Would you be interested in contributing this feature and the corresponding docs?

ktmouk:
Thank you for your reply! Your idea is superior to mine. I would like to contribute to this.

PR #566, opened 2023-10-15
feat: Add the custom unit option of `relativeTime`

ktmouk:
This PR adds the `relativeTime` options to support custom units and updates the documentation accordingly.
Close: https://github.com/amannn/next-intl/issues/565

### DIFF
--- a/packages/use-intl/src/core/createFormatter.tsx
+++ b/packages/use-intl/src/core/createFormatter.tsx
@@ -2,47 +2,64 @@ import DateTimeFormatOptions from './DateTimeFormatOptions';
 import Formats from './Formats';
 import IntlError, {IntlErrorCode} from './IntlError';
 import NumberFormatOptions from './NumberFormatOptions';
+import RelativeTimeFormatOptions from './RelativeTimeFormatOptions';
 import TimeZone from './TimeZone';
 import {defaultOnError} from './defaults';
 
-const MINUTE = 60;
+const SECOND = 1;
+const MINUTE = SECOND * 60;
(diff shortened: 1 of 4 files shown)


packages/use-intl/src/core/createFormatter.tsx
```diff
@@ -2,47 +2,64 @@ import DateTimeFormatOptions from './DateTimeFormatOptions';
...
+  days: DAY,
+  week: WEEK,
+  weeks: WEEK,
+  month: MONTH,
+  months: MONTH,
+  quarter: QUARTER,
+  quarters: QUARTER,
+  year: YEAR,
+  years: YEAR
+} as const;
```
ktmouk, line 35: I added `quarter` and plural forms to make it match [RelativeTimeFormatUnit](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat/format#parameters) type.

packages/use-intl/src/core/createFormatter.tsx
```diff
@@ -2,47 +2,64 @@ import DateTimeFormatOptions from './DateTimeFormatOptions';
...
+
+function resolveRelativeTimeUnit(seconds: number) {
```
ktmouk, line 37: I didn't add quarter here because I believe we don't want to use the quarter unit in most cases.

amannn (approved):
What a fantastic PR, thank you so much for this great contribution! 👏👏

I only added a minor commit from a local experiment where I wanted to check if the map you've introduced covers all relative time units.

Issue #568, opened 2023-10-15
Redirecting to a different locale pathname does not pass search params


BrendonSled:
### Description

When the middleware redirects a pathname to a localized version the search params are removed.

This has been tested with the latest RC 4

### Mandatory reproduction URL (CodeSandbox or GitHub repository)

https://codesandbox.io/p/sandbox/next-intl-bug-template-app-forked-f22qnx

### Reproduction description

Steps to reproduce:
1. Open reproduction
2. Click on about. Notice the search params are working, now click the link with a different local
3. See error: When the middleware redirects the pathname to the localized version the search param is removed

### Expected behaviour

Search params should stay intact when redirecting

amannn:
Good point, thanks for the report!

Would you be interested in adding a fix for this?

The relevant code is:
- [middleware.test.tsx](https://github.com/amannn/next-intl/blob/feat/next-13-rsc/packages/next-intl/test/middleware/middleware.test.tsx)
- [middleware.tsx](https://github.com/amannn/next-intl/blob/feat/next-13-rsc/packages/next-intl/src/middleware/middleware.tsx)

The RC is based on the `feat/next-13-rsc` branch.

Let me know if you're interested!

BrendonSled:
Sure thing! I created the PR

PR #569, opened 2023-10-16
fix(RSC): Fixed the search not being added on localized pathname redirects

BrendonSled:
Fixes https://github.com/amannn/next-intl/issues/568


amannn (approved):
Looks fantastic, thank you so much @BrendonSled!

I've added another commit that reuses an existing utility function and adds another test, but the result is the same.

Thank you so much! 🙌

Issue #243, opened 2023-04-14
Middleware: Support for `basePath` in the `app` directory


amannn:
Currently the middleware assumes that `/` is the root of the app and works on top of that.

We could adapt it as follows:
1. Don't hardcode the root URL
https://github.com/amannn/next-intl/blob/ed100335aebf921e90299bfd711070d64e796a99/packages/next-intl/src/middleware/middleware.tsx#L14
2. All the places where an absolute URL is constructed, the `basePath` needs to be considered
https://github.com/amannn/next-intl/blob/ed100335aebf921e90299bfd711070d64e796a99/packages/next-intl/src/middleware/middleware.tsx#L96
3. Somehow the `basePath` needs to be passed to the middleware. Maybe we could use the plugin that is worked on in https://github.com/amannn/next-intl/pull/149 to read it from the Next.js config and then make it automatically available to the middleware (e.g. with `webpack.DefinePlugin`).
 
Potentially this could also help to implement region-specific routing, see https://github.com/amannn/next-intl/issues/243#issuecomment-1655085673 and https://github.com/amannn/next-intl/issues/549

(discussed in https://github.com/amannn/next-intl/discussions/242)

amannn:
Hey, thanks for starting this discussion! This is definitely a valid feature request, I'll convert this to an issue. 

I haven't explored this topic in detail yet, but possibly, we'd have to wait for the lead from Next.js here:

> `basePath` has been removed. The alternative will not be part of useRouter. It has not yet been implemented.

(from [the upgrade guide](https://beta.nextjs.org/docs/upgrade-guide#step-5-migrating-routing-hooks))

I just saw that in the `pages` folder with i18n routing, Next.js appends the locale to the base path (e.g. `domain.tld/base-path/en/about` and even `domain.tld/base/path/en/about`). Is that also what you need?

setowilliam:
I'm having this problem too. It only happens when I try to go to the locale set to `defaultLocale`. For example:
```
export default createMiddleware({
  // A list of all locales that are supported
  locales: ['en', 'ja'],

  // If this locale is matched, pathnames work without a prefix (e.g. `/about`)
  defaultLocale: 'ja'
});
```

and `next.config.js`
```
const nextConfig = {
  reactStrictMode: true,
  basePath: '/path',
  experimental: {
    appDir: true
  }
};
```

If I try to navigate to `/path/en`, it will be fine. But navigating to `/path/ja` or `/path` will redirect to `/404`

[...]

PR #589, opened 2023-10-31
Feature/243 base path

Robjam:
This PR add support for `basePath` in the next-intl configuration and solves #243.
Documentation regarding middleware is also included (59f25d18).

### DIFF
--- a/packages/next-intl/src/middleware/NextIntlMiddlewareConfig.tsx
+++ b/packages/next-intl/src/middleware/NextIntlMiddlewareConfig.tsx
@@ -1,6 +1,9 @@
 type LocalePrefix = 'as-needed' | 'always' | 'never';
 
 type RoutingBaseConfig = {
+  /** The base path of your application (optional). */
+  basePath?: string;
+
   /** A list of all locales that are supported. */
   locales: Array<string>;
 
@@ -37,6 +40,7 @@ export type MiddlewareConfigWithDefaults = MiddlewareConfig & {
   alternateLinks: boolean;
   localePrefix: LocalePrefix;
   localeDetection: boolean;
+  basePath: string;
 };
 
 export default MiddlewareConfig;
--- a/packages/next-intl/src/middleware/getAlternateLinksHeaderValue.tsx
+++ b/packages/next-intl/src/middleware/getAlternateLinksHeaderValue.tsx
@@ -6,6 +6,11 @@ import {getHost, isLocaleSupportedOnDomain} from './utils';
 
 function getUnprefixedUrl(config: MiddlewareConfig, request: NextRequest) {
   const url = new URL(request.url);
+
+  if (config.basePath && !url.pathname.startsWith(config.basePath)) {
+    url.pathname = `${config.basePath}${url.pathname}`;
+  }
+
   const host = getHost(request.headers);
   if (host) {
     url.port = '';
@@ -45,10 +50,12 @@ export default function getAlternateLinksHeaderValue(
 
   const links = config.locales.flatMap((locale) => {
     function localizePathname(url: URL) {
-      if (url.pathname === '/') {
-        url.pathname = `/${locale}`;
+      if(url.pathname === config.basePath) {
+        url.pathname = `${url.pathname}/${locale}`;
+      } else if(config.basePath && url.pathname.startsWith(config.basePath)) {
+        url.pathname = url.pathname.replace(config.basePath, `${config.basePath}/${locale}`);
       } else {
-        url.pathname = `/${locale}${url.pathname}`;
+        url.pathname = `/${locale}`;
       }
       return url;
     }
@@ -71,6 +78,10 @@ export default function getAlternateLinksHeaderValue(
           config.localePrefix === 'always'
         ) {
           localizePathname(url);
+        }
+
+        if(config.basePath && !url.pathname.startsWith(config.basePath)) {
+          url.pathname = `${config.basePath}${url.pathname}`;
         }
 
         return getAlternateEntry(url.toString(), locale);
--- a/packages/next-intl/src/middleware/middleware.tsx
+++ b/packages/next-intl/src/middleware/middleware.tsx
@@ -18,7 +18,8 @@ function receiveConfig(config: MiddlewareConfig) {
     ...config,
     alternateLinks: config.alternateLinks ?? true,
     localePrefix: config.localePrefix ?? 'as-needed',
-    localeDetection: config.localeDetection ?? true
+    localeDetection: config.localeDetection ?? true,
+    basePath: config.basePath ?? '',
   };
 
   return result;
@@ -68,7 +69,10 @@ export default function createMiddleware(config: MiddlewareConfig) {
     }
 
     function rewrite(url: string) {
-      return NextResponse.rewrite(new URL(url, request.url), getResponseInit());
+      if (configWithDefaults.basePath && !url.startsWith(configWithDefaults.basePath)) {
+        url = `${configWithDefaults.basePath}${url}`;
+      }
+      return NextResponse.rewrite(new URL(url,request.url), getResponseInit());
     }
 
     function next() {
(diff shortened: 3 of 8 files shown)


amannn:
Thank you so much for looking into this @Robjam! I was on vacation the last few days and need to catch up on a few things. I'll look into this PR next week and get back to you!

mwhiter:
Will be watching this PR, our team is running into issues with `basePath` as well and would like to see this merged in.

Let me know if there's anything we can do to help support this fix, seems like a very important feature to include in `next-intl`.

nextjs-stonks:
Hi @amannn, would you be able to provide an ETA for this PR to be reviewed and subsequently fix the issue with utilizing `basePath` ? Our team is also running into issues with utilizing the library and would appreciate for a fix to be put out as soon as possible :)

packages/next-intl/src/middleware/getAlternateLinksHeaderValue.tsx
```diff
@@ -73,6 +80,10 @@ export default function getAlternateLinksHeaderValue(
           localizePathname(url);
         }
 
+        if(config.basePath && !url.pathname.startsWith(config.basePath)) {
```
amannn, line 83: There are btw. some small formatting issues, you can fix them via ESLint:

```tsx
pnpm eslint src test --fix
```

… or autofix them on save in case you use VSCode:

```
  "[typescript]": {
    "editor.defaultFormatter": "dbaeumer.vscode-eslint",
    "editor.formatOnSave": true
  },
```

I should really consider setting up https://github.com/wearerequired/lint-action to make it easier for contributors.

packages/next-intl/src/middleware/NextIntlMiddlewareConfig.tsx
```diff
@@ -37,6 +40,7 @@ export type MiddlewareConfigWithDefaults = MiddlewareConfig & {
   alternateLinks: boolean;
   localePrefix: LocalePrefix;
   localeDetection: boolean;
+  basePath: string;
```
amannn, line 43: I did a first test with `example-next-13` in the repo. With `localePrefix: 'always'` this already seems to work quite well! Pretty cool that `Link` automatically considers the `basePath` if it's set in `next.config.js` ([source](https://nextjs.org/docs/pages/api-reference/next-config-js/basePath#links)).

With `localePrefix: 'as-necessary'` I run into infinite redirects though when trying to access the app root. Can you reproduce that?

packages/next-intl/src/middleware/middleware.tsx
```diff
@@ -68,7 +69,10 @@ export default function createMiddleware(config: MiddlewareConfig) {
     }
 
     function rewrite(url: string) {
-      return NextResponse.rewrite(new URL(url, request.url), getResponseInit());
```
amannn, line 71: I think the problem here is the construction with `request.url`. If we apply `request.pathname` then maybe we can avoid introducing more conditions.

packages/next-intl/src/middleware/middleware.tsx
```diff
@@ -93,7 +97,7 @@ export default function createMiddleware(config: MiddlewareConfig) {
               bestMatchingDomain.defaultLocale === locale &&
               configWithDefaults.localePrefix === 'as-needed'
             ) {
-              urlObj.pathname = urlObj.pathname.replace(`/${locale}`, '');
+              urlObj.pathname = urlObj.pathname.replace(`${configWithDefaults.basePath}/${locale}`, '');
```
amannn, line 100: Also here above we're using `request.url`, which includes the `basePath`.

packages/next-intl/src/middleware/middleware.tsx
```diff
@@ -102,13 +106,15 @@ export default function createMiddleware(config: MiddlewareConfig) {
...
+      if (configWithDefaults.basePath && !urlObj.pathname.startsWith(configWithDefaults.basePath)) {
+        urlObj.pathname = `${configWithDefaults.basePath}${urlObj.pathname}`;
+      }
       return NextResponse.redirect(urlObj.toString());
     }
 
     let response;
     if (isRoot) {
-      let pathWithSearch = `/${locale}`;
+      let pathWithSearch = `${configWithDefaults.basePath}/${locale}`;
```
amannn, line 117: Maybe we can avoid this change too, `pathWithSearch` is passed to `rewrite` and `redirect` below, which—as mentioned above—could likely be adjusted to automatically consider the `basePath`.

packages/next-intl/test/middleware/getAlternateLinksHeaderValue.test.tsx
```diff
@@ -1,17 +1,22 @@
...
 import {NextRequest} from 'next/server';
-import {it, expect} from 'vitest';
+import {it, expect, describe} from 'vitest';
 import {MiddlewareConfigWithDefaults} from '../../src/middleware/NextIntlMiddlewareConfig';
 import getAlternateLinksHeaderValue from '../../src/middleware/getAlternateLinksHeaderValue';
 
+describe.each([
+  { basePath: undefined },
+  { basePath: '/sample' }
+])('when basePath is $basePath', ({ basePath }: { basePath: string }) => {
```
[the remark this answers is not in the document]
mwhiter: Does `basePath` need to be `undefined` here? It appears in `NextIntlMiddlewareConfig.tsx`, this property is defined as non-optional, so would we just want a default `/` here to test instead, or should the `MiddlewareConfigWithDefaults` type be updated to accept `undefined`?

The tests rely on a reference to `basePath` to do string concatenation; if the intention is for `basePath` to be `undefined` here, you may want to have another param here called `expectedPath` or something that will be defined (empty string in the case `basePath` is `undefined`) to use instead.

packages/next-intl/src/middleware/NextIntlMiddlewareConfig.tsx
Robjam: I didn't test locally with `localePrefix: 'as-necessary'` as we require the locale to be in the path. I'll scaffold a new next app and reproduce it.

packages/next-intl/test/middleware/getAlternateLinksHeaderValue.test.tsx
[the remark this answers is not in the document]
Robjam: good point!
This already has a default set and so `/` is a better case.

packages/next-intl/src/middleware/getAlternateLinksHeaderValue.tsx
Robjam: Thanks for the how-to on setting up VsCode.
Opening VSCode from the repository root path gives an error with `Parsing error: Cannot read file '/users/robjam/code/community/next-intl/tsconfig.json'. eslint` so I use the cli to format things. Maybe we can fix this with a root tsconfig.json and extend it from each workspace?

amannn: My understanding of monorepos is that individual folders should be opened when working inside of them. E.g. when working on the `next-intl` package, you'd open this as your workspace root. If that isn't the case, ESLint will not work correctly.

[I've added an info about this](https://github.com/amannn/next-intl/commit/d6b1e8b24cbae0768d6482ac73293c2a44f662a1) to [the contributors guide](https://github.com/amannn/next-intl/blob/main/CONTRIBUTORS.md) for now.

amannn:
Hey @Robjam, did you have another minute to look into this?

next-intl 3.0 has been released in the meantime with support for Server Components, as well as some improvements for the middleware. It seems like that has caused some conflicts with this PR. Can you figure out how to incorporate the changes or can I help with something?

kjxbyz:
When will this pr be merged? This feature is very important to me.

packages/next-intl/src/middleware/NextIntlMiddlewareConfig.tsx
```diff
@@ -1,6 +1,9 @@
 type LocalePrefix = 'as-needed' | 'always' | 'never';
 
 type RoutingBaseConfig = {
+  /** The base path of your application (optional). */
+  basePath?: string;
```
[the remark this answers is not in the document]
amannn: I noticed the `basePath` is available on the `request` object:

```tsx
import type { NextRequest } from 'next/server'
 
export function middleware(request: NextRequest) {
  // Either '' (default) or a value like '/test'
  console.log({basePath: request.nextUrl.basePath});
}
```

That way we don't have to make this configurable in the first place.

As mentioned in the other comment, I think we only need it for the alternate links though.

Issue #596, opened 2023-11-03
Support for NextAuth v5 multiple middleware


JulianJorgensen:
### Is your feature request related to a problem? Please describe.

I use both next-intl and next-auth middleware. There are some nice examples of this combination for next-auth v4 (`withAuth()`), but not for v5 when using `auth`.

Would love to see.

### Describe the solution you'd like

I imagine the usage something like this:

```js
import { auth } from 'auth';
import createMiddleware from 'next-intl/middleware';

const intlMiddleware = createMiddleware({
  locales: ['en', 'dk'],
  defaultLocale: 'en'
});

export default auth((req) => intlMiddleware(req))
```

### Describe alternatives you've considered

Alternatively I can stick with using next-auth v4

amannn:
I currently don't have experience with v5. We have [an example within this repo](https://github.com/amannn/next-intl/tree/main/examples/example-next-13-next-auth) as well as [documentation on usage with v4](https://next-intl-docs.vercel.app/docs/routing/middleware#example-auth-js).

Would you be interested in setting up a pull request with an update to v5? We have [a few automated tests in example-next-13-next-auth](https://github.com/amannn/next-intl/blob/main/examples/example-next-13-next-auth/tests/main.spec.ts) that can help to verify if the updated code works as expected.

As far as I know, v5 is in beta currently though. We should wait until v5 becomes stable before updating our example in the `main` branch.

amarincolas:
I am trying to make it work but apparently, this does not work anymore

```
const authMiddleware = auth(
  (req) => intlMiddleware(req),
  {
    callbacks: {
      authorized: ({token}) => token != null
    },
    pages: {
      signIn: '/login'
    }
  }
);

```

So I decided to follow the next-auth and next.js guidelines about this and I came up with this:

```
export const authConfig = {
  pages: {
    signIn: '/signin',
  },
  providers: [],
  callbacks: {
    authorized({ auth, request: { nextUrl } }) {
      ...
    },
  },
} satisfies NextAuthConfig
```

```
export const { auth, signIn, signOut } = NextAuth({
  ...authConfig,
  providers: [
    CredentialsProvider({
      name: 'Credentials',
      credentials: {
        username: { type: 'text' },
        password: { type: 'password' },
      },
      authorize(credentials) {
        ...
    }),
  ],
})
```

```
const authMiddleware = auth(
  (request) => intlMiddleware(request)
)
```

But the thing is this does not work either...

[...]

PR #598, opened 2023-11-03
Change from NextAuth v4 syntax to v5 syntaxes

JulianJorgensen:
Here's initial draft of pull request. Apologies for format issues.


amarincolas:
Apparently, this does not work anymore

`
const authMiddleware = auth(
  // Note that this callback is only invoked if
  // the `authorized` callback has returned `true`
  // and not for pages listed in `pages`.
  (req) => intlMiddleware(req),
  {
    callbacks: {
      authorized: ({token}) => token != null
    },
    pages: {
      signIn: '/login'
    }
  }
);`

DiegoGonzalezCruz:
I'm having issues with this... any ideas?

alessandrojcm:
I got this working using [this](https://github.com/amannn/next-intl/issues/596#issuecomment-1857635793) workaround. This is necessary due to a [bug](https://github.com/nextauthjs/next-auth/issues/9045#issuecomment-1810875045) in `next-auth` itself. So this may need to be parked until that bug is resolved.

Robin-Ln:
https://github.com/JulianJorgensen/next-intl/pull/1

amannn:
@ScreamZ shared his implementation here: https://github.com/amannn/next-intl/discussions/1170. Might be worth checking that out when we continue with this PR.

Also https://github.com/amannn/next-intl/pull/899 provides a potential implementation.

For context, I didn't look further into this yet as I'm waiting for Auth.js v5 stable to be released. I'd like to avoid keeping up to date with their beta versions, as long as the API might still change.

Many thanks to everyone who has already looked into this and provided help! 🙌

amannn:
Since [Auth.js has joined Better Auth](https://www.better-auth.com/blog/authjs-joins-better-auth), Auth.js is no longer recommended in favor of Better Auth. Due to this, I'd like to not maintain a NextAuth example in this repo anymore.

You can however check out the Better Auth example if you're interested: https://github.com/amannn/next-intl/issues/596#issuecomment-3733057974

PR #619, opened 2023-11-11
docs: match example with the messages

fcasibu:
To match the example with the above dictionary/messages json as it's currently incorrect.

image


amannn (approved):
Thank you so much!

PR #637, opened 2023-11-16
fix(examples/with-next-auth): fix regex in middleware

A7med3bdulBaset:
The regex `^(/(${locales.join('|')}))?(${publicPages.join('|')})/?$` considers `/` or `/${locale}/` public.

This PR fixes that

### DIFF
--- a/docs/pages/docs/routing/middleware.mdx
+++ b/docs/pages/docs/routing/middleware.mdx
@@ -470,7 +470,7 @@ const authMiddleware = withAuth(
 
 export default function middleware(req: NextRequest) {
   const publicPathnameRegex = RegExp(
-    `^(/(${locales.join('|')}))?(${publicPages.join('|')})?/?$`,
+    `^(/(${locales.join('|')}))?(${publicPages.join('|')})/?$`,
     'i'
   );
   const isPublicPage = publicPathnameRegex.test(req.nextUrl.pathname);
--- a/docs/pages/docs/routing/middleware.mdx
+++ b/docs/pages/docs/routing/middleware.mdx
@@ -470,11 +470,11 @@ const authMiddleware = withAuth(
 
 export default function middleware(req: NextRequest) {
   const publicPathnameRegex = RegExp(
-		`^(/(${locales.join("|")}))?(${publicPages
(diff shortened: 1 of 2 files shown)


amannn (commented):
Oh cool, thanks for this fix!

I think this is the same that was proposed in https://github.com/amannn/next-intl/pull/501. The PR was abandoned though, since [a test failed](https://github.com/amannn/next-intl/pull/501#issuecomment-1721401555) when updating the corresponding example.

Would you like to have a look?

A7med3bdulBaset:
@amannn resolved

docs/pages/docs/routing/middleware.mdx
```diff
@@ -470,7 +470,7 @@ const authMiddleware = withAuth(
 
 export default function middleware(req: NextRequest) {
   const publicPathnameRegex = RegExp(
-    `^(/(${locales.join('|')}))?(${publicPages.join('|')})?/?$`,
+    `^(/(${locales.join('|')}))?(${publicPages.join('|')}|)/?$`,
```
amannn, line 473: Hmm, doesn't this still consider `/` or `/${locale}` public, even if `/` is not included in `publicPages`?

Screenshot 2023-11-17 at 10 17 50

amannn: Looks great!

Note that Next.js by default [removes trailing slashes from pathnames](https://nextjs.org/docs/app/api-reference/next-config-js/trailingSlash). I think we can simplify the code slightly based on that, but other than this, this should be good to go! 👏

A7med3bdulBaset: The only change for trailing slash will be removing the last `/?` which does not simplify a lot. I think we should keep the regex handles both with/without trailing slashes since both are possible.

> By default Next.js will redirect urls with trailing slashes to their counterpart without a trailing slash... **You can configure this behavior to act the opposite way**...

 or do you mean other simplification?

A7med3bdulBaset (commented):
Should I add a unit test for the regex?

amannn:
> Should I add a unit test for the regex?

Fine by me to keep it like this, as long as we have e2e test coverage.

Thank you so much @A7med3bdulBaset! 🙌

PR #644, opened 2023-11-17
Update middleware.tsx

seitbekir:
This makes it possible to prevent adding redundunt prefix, when it is not needed by configuration

### DIFF
--- a/packages/next-intl/src/middleware/middleware.tsx
+++ b/packages/next-intl/src/middleware/middleware.tsx
@@ -216,7 +216,7 @@ export default function createMiddleware<Locales extends AllLocales>(
               (configWithDefaults.localePrefix === 'as-needed' ||
                 configWithDefaults.domains))
           ) {
-            response = rewrite(`/${locale}${pathWithSearch}`);
+            response = rewrite(`/${pathWithSearch}`);
           } else {
             response = redirect(`/${locale}${pathWithSearch}`);
           }


packages/next-intl/src/middleware/middleware.tsx
```diff
@@ -216,7 +216,7 @@ export default function createMiddleware<Locales extends AllLocales>(
               (configWithDefaults.localePrefix === 'as-needed' ||
                 configWithDefaults.domains))
           ) {
-            response = rewrite(`/${locale}${pathWithSearch}`);
+            response = rewrite(`/${pathWithSearch}`);
```
amannn, line 219: This breaks existing functionality (see [the tests](https://github.com/amannn/next-intl/actions/runs/6906900504/job/18839973654?pr=644)).

I'm not sure what you're trying to achieve here, it might help to discuss this in an issue first (in case this hasn't been discussed yet).

zipme: Not sure if this is related: I found a bug when calling `redirect` (via `createLocalizedPathnamesNavigation`) from a server action, even though the `localePrefix` is set to `never`, the prefix will still be prepended. Calling `redirect` from a rsc seems to work fine.

amannn: I think that's this bug: https://github.com/vercel/next.js/issues/58281

amannn:
Closing due to inactivity …

PR #661, opened 2023-11-23
feat: add ability to access translation values on getMessageFallback

artpopov:
In some cases i really need translation values to construct fallback message.
This PR intends to solve this

### DIFF
--- a/packages/use-intl/src/core/IntlConfig.tsx
+++ b/packages/use-intl/src/core/IntlConfig.tsx
@@ -27,6 +27,7 @@ type IntlConfig<Messages = AbstractIntlMessages> = {
     error: IntlError;
     key: string;
     namespace?: string;
+    values?: RichTranslationValues;
   }): string;
   /**
    * Providing this value will have two effects:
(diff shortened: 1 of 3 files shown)


packages/use-intl/src/core/IntlConfig.tsx
```diff
@@ -27,6 +27,7 @@ type IntlConfig<Messages = AbstractIntlMessages> = {
     error: IntlError;
     key: string;
     namespace?: string;
+    values?: RichTranslationValues;
```
amannn, line 30:
```suggestion
    values?:
      | TranslationValues
      | RichTranslationValues
      | MarkupTranslationValues;
```

packages/use-intl/src/core/createBaseTranslator.tsx
```diff
@@ -370,7 +376,9 @@ function createBaseTranslatorImpl<
       );
 
       onError(error);
-      return getMessageFallback({error, key, namespace});
+      // @ts-expect-error -- `MarkupTranslationValues` is practically a sub type
+      // of `RichTranslationValues` but TypeScript isn't smart enough here.
```
amannn, line 380: We can remove this by following my suggestion from above.

amannn (commented):
Hey, thank for the proposal!

I think generally this should be ok, but can you tell me a bit more about your use case? I'd like to make sure we're solving the right problem.

A test would also be necessary before this can be merged.

PR #689, opened 2023-12-04
refactor: Improve route matching in middleware

adrigar94:
the config.matcher now sets the languages from config.locales

### DIFF
--- a/examples/example-app-router/src/middleware.ts
+++ b/examples/example-app-router/src/middleware.ts
@@ -10,5 +10,5 @@ export default createMiddleware({
 
 export const config = {
   // Match only internationalized pathnames
-  matcher: ['/', '/(de|en)/:path*']
+  matcher: ['/', '/('+locales.join('|')+')/:path*']
 };


amannn:
I think this doesn’t work:

> Good to know: The matcher values need to be constants so they can be statically analyzed at build-time. Dynamic values such as variables will be ignored.

from https://nextjs.org/docs/app/building-your-application/routing/middleware#matcher

adrigar94:
Yes, you are right. It doesn't work because the matcher doesn't accept dynamic values, and it seems that in typescript there is no such thing as a static variable at build time.

nextor2k:
```javascript
import { MiddlewareFactory } from './factory';
import { NextFetchEvent, NextRequest, NextResponse } from 'next/server';
import { match } from 'path-to-regexp';

type withLocaleOptions = {
  nextIntl: any;
  locales: string[];
};

export const withLocale = (options: withLocaleOptions): MiddlewareFactory => {
  const { nextIntl, locales } = options;
  const matcher = match(['/', '/-/:path*', `/(${locales.join('|')})/:path*`]);
  return (next) => {
    return async (req: NextRequest, event: NextFetchEvent) => {
      const { pathname } = req.nextUrl;

      if (matcher(pathname)) {
        const response = nextIntl(req);
        return next(req, event, response);
      }

      return next(req, event, NextResponse.next());
    };
  };
};
```

PR #691, opened 2023-12-05
fix: Prefer `locale` from `NextIntlClientProvider` when using `useLocale` in Client Components. `useParams().locale` is still used as a fallback.

### LINKED ISSUE #688: useLocale returns locale from path params instead of provider value in client components
### Description

`useLocale` will always return the locale path parameter in client components even though `NextIntlClientProvider` is configured with a hard-coded locale. 

I cannot find anything related to this behaviour in the documentation so this should probably be documented and/or add an option to useLocale to retrieve the locale from the provider and not the path parameters for client components.

https://github.com/amannn/next-intl/blob/4369c7612b860ab31a173e7f045a839aa742f711/packages/next-intl/src/react-client/useLocale.tsx#L9-L18

https://codesandbox.io/p/devbox/next-intl-bug-template-app-forked-w4q7qq

### Reproduction description

1. Open up the `/_next/info` page and view the resulting locales.
2. Open up the `/en/info` page and view the resulting locales.
3. Remove the `notFound();` call in `src/app/[locale]/layout.tsx`.
4. Open up the `/_next/info` page and view the resulting locales.

### Expected behaviour

The universal "useLocale" hook should return the expected locale configured in the provider in client components.
All displayed locales should be "en" in client components.

This ensures that components can be re-used in 404:s from _next/* paths.

amannn:
Fixes #688

### DIFF
--- a/docs/pages/blog/next-intl-3-0.mdx
+++ b/docs/pages/blog/next-intl-3-0.mdx
@@ -134,7 +134,7 @@ If you're not, be aware that this will by default opt the corresponding page int
 6. `createTranslator(…).rich` now returns a `ReactNode`. Previously, this was somewhat confusing, since `t.rich` accepted and returned either React elements or strings depending on if you retrieve the fuction via `useTranslations` or `createTranslator`. Now, an explicit [`t.markup`](/docs/usage/messages#html-markup) function has been added to generate markup strings like `'<b>Hello</b>'` outside of React components.
 7. `useIntl` has been replaced with [`useFormatter`](/docs/usage/dates-times) (deprecated since v2.11).
 8. `createIntl` has been replaced with [`createFormatter`](/docs/environments/core-library) (deprecated since v2.11).
-9. `useLocale` currently requires `next@>=13.3.0`. If you're on an older version of Next.js and you're using the Pages Router, please use `useRouter().locale` instead.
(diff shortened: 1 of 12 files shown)


docs/pages/blog/next-intl-3-0.mdx
```diff
@@ -134,6 +134,7 @@ If you're not, be aware that this will by default opt the corresponding page int
 6. `createTranslator(…).rich` now returns a `ReactNode`. Previously, this was somewhat confusing, since `t.rich` accepted and returned either React elements or strings depending on if you retrieve the fuction via `useTranslations` or `createTranslator`. Now, an explicit [`t.markup`](/docs/usage/messages#html-markup) function has been added to generate markup strings like `'<b>Hello</b>'` outside of React components.
 7. `useIntl` has been replaced with [`useFormatter`](/docs/usage/dates-times) (deprecated since v2.11).
 8. `createIntl` has been replaced with [`createFormatter`](/docs/environments/core-library) (deprecated since v2.11).
+9. `useLocale` currently requires `next@>=13.3.0`. If you're on an older version of Next.js and you're using the Pages Router, please use `useRouter().locale` instead.
```
amannn, line 137:
```suggestion
9. `useLocale` now requires `next@>=13.3.0`. If you're on an older version of Next.js and you're using the Pages Router, please use `useRouter().locale` instead.
```

PR #692, opened 2023-12-05
fix: Adjustment to the pathname generation when localePrefix is 'as-needed'

anna-colenso:
Right now in my current project if a url has `/${locale}` string in the url it's erased in several places. 
E.g. if the url is `https://domain.com/en/projects/energy` and the default locale is 'en' -> we end up with `https://domain.com/projectsergy`.
The above behaviour happens only when localePrefix is set to `as-needed`

This patch makes sure we are only replacing it if the pathname starts with `/${locale}` and using a word regex for safety.

### DIFF
--- a/packages/next-intl/src/middleware/middleware.tsx
+++ b/packages/next-intl/src/middleware/middleware.tsx
@@ -87,8 +87,9 @@ export default function createMiddleware<Locales extends AllLocales>(
             if (
               bestMatchingDomain.defaultLocale === locale &&
               configWithDefaults.localePrefix === 'as-needed'
-            ) {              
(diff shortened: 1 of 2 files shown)


amannn (commented):
Oh sure, thank you so much for contributing this fix! While reviewing your change, I found another issue in the alternate links as well as an opportunity to unify some code paths.

Can we merge https://github.com/colensobbdo/next-intl/pull/1 into this?

anna-colenso:
Sure thing!

amannn:
@anna-colenso I think the PR was only approved, not merged. I've now set up https://github.com/amannn/next-intl/pull/698 that includes all relevant commits and therefore supersedes this PR. Hope this is ok for you!

PR #698, opened 2023-12-06
fix: Correctly normalize pathname for redirects and alternate links when `localePrefix: 'as-needed'` is used with the default locale

amannn:
When removing the default locale from a path like `/en/energy`, the middleware would erroneously remove the characters `/en` not only from the beginning, but also from the middle when redirecting to the unprefixed version (i.e. `/ergy`).

This PR is based on the fix from https://github.com/amannn/next-intl/pull/692 by @anna-colenso.

### DIFF
--- a/packages/next-intl/src/middleware/middleware.tsx
+++ b/packages/next-intl/src/middleware/middleware.tsx
@@ -10,7 +10,6 @@ import resolveLocale from './resolveLocale';
 import {
   getInternalTemplate,
   formatTemplatePathname,
-  getBasePath,
   getBestMatchingDomain,
   getKnownLocaleFromPathname,
   getNormalizedPathname,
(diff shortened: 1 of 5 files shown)


quyphan97:
@amannn Can you give me an example of localPrefix: 'as-needed', I'm confused because I want to set the default locale :((

anna-colenso:
> @amannn Can you give me an example of localPrefix: 'as-needed', I'm confused because I want to set the default locale :((

You can find it in the docs here: https://next-intl-docs.vercel.app/docs/routing/middleware#locale-prefix-as-needed

quyphan97:
image
image
The structure in the app overlaps with [locale], is there any way to remove pages outside of [locale].

anna-colenso:
> image image @anna-colenso The structure in the app overlaps with [locale], is there any way to remove pages outside of [locale].

Not sure I understood your question, sorry.
If you place your pages outside of the [locale] folder, they won't be a part of the localisation with the /locale/ path. Because they won't be able to use the locale param value from the path.
There are 2 strategies this package supports ([locale] and domain based) and they are described in [here](https://next-intl-docs.vercel.app/docs/routing/middleware#strategies).
Anything other than that would require you to work on your own logic in the nextjs middleware.

quyphan97:
Well thank you very much!

quyphan97:
@anna-colenso How to use localePrefix: "as-needed" is a 301 redirect? Help me bro.

PR #705, opened 2023-12-06
feat: AOT compilation with `icu-to-json` (experiment)

### LINKED ISSUE #962: Ahead-of-time compilation of messages

First results from [`icu-to-json`](https://github.com/jantimon/icu-to-json) show, that messages can be compiled to an ultra-slim AST (or even just a plain string—in the case of plain strings) that has very little overhead in comparison to passing raw messages to the client side.

**The benefits are:**
 - Faster message formatting at runtime
 - No parsing on the client side
 - Smaller bundle size

Switch out `intl-messageformat` for `icu-to-json` and make sure the parser only runs on the server side (potentially cached) / at compile time.

Sticking with `intl-messageformat`.

Also related:
 - https://github.com/lingui/js-lingui/blob/main/packages/message-utils/src/compileMessage.ts#L94
 - https://github.com/adobe/react-spectrum/pull/3294. However, I think this pattern won't work well with RSC since functions can't be serialized.

---

https://github.com/amannn/next-intl/pull/2220 brings initial support for AOT compilation (experimental).

**Future work**
- [ ] Can we parallelize in loader? (Rust?)
- [ ] Deprecate `toJSONString`?
- [ ] Rip out `intl-messageformat` entirely

amannn:
Fixes #962 

(WIP)

This is an experiment to use `icu-to-json` for AOT compilation of messages on the server side / at build time and to reduce the bundle size in Client Components.

**Current progress:** For now, both compilation as well as evaluation happens in the same module (and therefore on the client side). This is mostly intended to evaluate initial compatibility. Later, we could move the compilation step to the server.

**Questions:**
- [ ] How to support `t.raw` or `t.markup`? We need to decide if a message needs to be parsed before it's used. Do we need to analyze the usages? Or could we revert the parsing when `t.raw` is called?

**TODO**
- [ ] Can this help with https://github.com/amannn/next-intl/pull/758#discussion_r1609624363?

### DIFF
--- a/packages/use-intl/src/core/MessageFormat.tsx
+++ b/packages/use-intl/src/core/MessageFormat.tsx
@@ -1,9 +1,5 @@
-import type {compile} from 'icu-to-json/compiler';
+import type {CompiledAst} from 'icu-to-json';
 
-type MessageFormat = Omit<
-  ReturnType<typeof compile>,
-  // TODO: Do we need the args?
-  'args'
->;
+type MessageFormat = CompiledAst;
 
 export default MessageFormat;
--- a/packages/use-intl/src/core/createBaseTranslator.tsx
+++ b/packages/use-intl/src/core/createBaseTranslator.tsx
@@ -1,6 +1,6 @@
 import {evaluateAst} from 'icu-to-json';
-import {compile} from 'icu-to-json/compiler';
-import {ReactElement} from 'react';
+import {compileToJson} from 'icu-to-json/compiler';
+import React, {Fragment, ReactElement} from 'react';
 import AbstractIntlMessages from './AbstractIntlMessages';
 import Formats from './Formats';
 import {InitializedIntlConfig} from './IntlConfig';
@@ -146,7 +146,7 @@ function createBaseTranslatorImpl<
     values?: RichTranslationValues,
     /** Provide custom formats for numbers, dates and times. */
     formats?: Partial<Formats>
-  ): string | ReactElement | Array<ReactElement> {
+  ): string | ReactElement {
     if (messagesOrError instanceof IntlError) {
       // We have already warned about this during render
       return getMessageFallback({
@@ -202,7 +202,7 @@ function createBaseTranslatorImpl<
       }
 
       try {
-        messageFormat = compile(message);
+        messageFormat = compileToJson(message);
       } catch (error) {
         return getFallbackFromErrorAndNotify(
           key,
@@ -219,7 +219,7 @@ function createBaseTranslatorImpl<
       // TODO: The return type seems to be a bit off, not sure if
       // this should be handled in `icu-to-json` or here.
       const evaluated = evaluateAst(
-        messageFormat.json,
+        messageFormat,
         locale,
         allValues,
         getFormatters(timeZone, formats, globalFormats)
@@ -234,7 +234,10 @@ function createBaseTranslatorImpl<
         formattedMessage = evaluated[0];
       } else {
         // Rich text
-        formattedMessage = evaluated;
+        formattedMessage = evaluated.map((part, index) => (
+          // @ts-expect-error TODO
+          <Fragment key={index}>{part}</Fragment>
+        ));
       }
 
       // TODO: Add a test that verifies when we need this
@@ -249,9 +252,7 @@ function createBaseTranslatorImpl<
       }
 
       // @ts-expect-error Verify return type (see comment above)
-      return Array.isArray(formattedMessage)
-        ? formattedMessage
-        : String(formattedMessage);
+      return formattedMessage;
     } catch (error) {
       return getFallbackFromErrorAndNotify(
         key,


packages/use-intl/test/react/useTranslations.test.tsx
```diff
@@ -1,9 +1,7 @@
 import {render, screen} from '@testing-library/react';
 import {parseISO} from 'date-fns';
-// eslint-disable-next-line import/no-named-as-default -- False positive
```
amannn, line 3: Not too bad!

Screenshot 2023-12-06 at 17 51 22

packages/use-intl/test/react/useTranslations.test.tsx
```diff
@@ -83,12 +81,14 @@ it('handles number formatting with percent', () => {
   screen.getByText('31%');
 });
 
-it('handles number formatting with a static currency', () => {
+// TODO: icu-to-json doesn't forward options to formatters
+it.skip('handles number formatting with a static currency', () => {
   renderMessage('{price, number, ::currency/EUR}', {price: 123394.1243});
   screen.getByText('€123,394.12');
 });
```
amannn, line 88: Needs to be addressed on the `icu-to-json` side I think.

packages/use-intl/test/react/useTranslations.test.tsx
```diff
@@ -83,12 +81,14 @@ it('handles number formatting with percent', () => {
...
 
-it('handles number formatting with defined decimals', () => {
+// TODO: icu-to-json doesn't forward options to formatters
+it.skip('handles number formatting with defined decimals', () => {
   renderMessage('{value, number, ::.#}', {value: 123394.1243});
   screen.getByText('123,394.1');
 });
```
amannn, line 94: Needs to be addressed on the `icu-to-json` side I think.

packages/use-intl/src/core/MessageFormat.tsx
```diff
@@ -0,0 +1,9 @@
+import type {compile} from 'icu-to-json/compiler';
+
+type MessageFormat = Omit<
+  ReturnType<typeof compile>,
+  // TODO: Do we need the args?
+  'args'
```
amannn, line 6: Could this be removed from the parsed result?

packages/use-intl/test/react/useTranslations.test.tsx
```diff
@@ -147,6 +147,22 @@ it('applies a time zone when using a built-in format', () => {
   expectFormatted('date', 'short', '5/8/23');
 });
 
+it('applies a time zone when using a date skeleton', () => {
+  const now = new Date('2024-01-01T00:00:00.000+0530');
+  renderMessage(`{now, date, ::yyyyMdHm}`, {now}, undefined, {
+    timeZone: 'Asia/Kolkata'
+  });
+  screen.getByText('1/1/2024, 00:00');
+});
```
amannn, line 156: This even fixes a bug that currently affects Format.JS: https://github.com/formatjs/formatjs/issues/4279

packages/use-intl/src/core/MessageFormat.tsx
jantimon: use `compileToJson` instead - args are mainly for types or to omptimize which formatters should be included

packages/use-intl/src/core/createBaseTranslator.tsx
```diff
@@ -276,14 +215,29 @@ function createBaseTranslatorImpl<
...
 
     try {
-      const formattedMessage = messageFormat.format(
-        // @ts-expect-error `intl-messageformat` expects a different format
-        // for rich text elements since a recent minor update. This
-        // needs to be evaluated in detail, possibly also in regards
-        // to be able to format to parts.
-        prepareTranslationValues({...defaultTranslationValues, ...values})
+      const allValues = {...defaultTranslationValues, ...values};
+      // TODO: The return type seems to be a bit off, not sure if
```
jantimon, line 219: there are two different ways to run the precompiled icu messages:

1. `run`
2. `evaluateAst`

the idea is that `run` will always return a string and `evaluateAst` will always return an array.

```tsx
const t = (key, args) => run(messages[key], lang, args);
```

`icu-to-json` is framework independent so it would require you to wrap a react Fragment around the result:

```tsx
t.rich = (key, args) => <>{evaluateAst(messages[key], lang, args)}</>;
```

packages/use-intl/src/core/createBaseTranslator.tsx
```diff
@@ -276,14 +215,29 @@ function createBaseTranslatorImpl<
...
+      // this should be handled in `icu-to-json` or here.
+      const evaluated = evaluateAst(
+        messageFormat.json,
+        locale,
+        allValues,
```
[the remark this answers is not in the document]
jantimon: the args are not needed for formatters however the args can tell you during compile time which formatters need to be available to run the code.

in your case there is a very special formatter wich you should add for full react support

it allows processing the children before they are passed to the given tag function:

```tsx
{
   tag: (children: Array<string | ReactNode>, locale: string) => 
         children.map((value, i) => typeof value === "string"
            ? value 
            : <Fragment key={`f-${i}`}>{value}</Fragment>
}
```

e.g. for a translation like `<foo>Hello <b>{name}</b></foo>`

packages/use-intl/src/core/getFormatters.tsx
```diff
@@ -0,0 +1,138 @@
...
+      timeZoneName: 'short'
+    }
+  }
+} as const;
+
+type FormatNameOrArgs<Options> =
+  | string
+  | {
+      type: number; // TODO: Unused, is this necessary?
+      tokens: Array<unknown>; // TODO: Unused, is this necessary?
```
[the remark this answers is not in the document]
jantimon: Not sure what you mean - Formatters are called here with `(value, locale, formatOptions)`

https://github.com/jantimon/icu-to-json/blob/83f24322bfd4ab66402e70d9727a650720501925/src/runtime.ts#L150-L152

packages/use-intl/test/react/useTranslations.test.tsx
```diff
@@ -412,7 +412,8 @@ describe('t.rich', () => {
     );
   });
 
-  it('handles nested rich text', () => {
+  // TODO: icu-to-json doesn't seem to handle this currently
+  it.skip('handles nested rich text', () => {
     const {container} = renderRichTextMessage(
       'This is <bold><italic>very</italic> important</bold>',
```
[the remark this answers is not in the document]
jantimon: this "should" work 😄

you might have to define a `tag` and/or a `baseTag` formatter

amannn:
https://next-intl.dev/blog/precompilation

PR #713, opened 2023-12-11
fix: Use .default for message imports in pages router example

pockata:
Hello and thanks for a great plugin! 
This is a small PR to help save someone 20 min of debugging. It adds `.default` to awaited imports of messages in the Advanced Pages Router example. This approach is used in every other example as far as I could tell.


amannn (approved):
Looks great, thank you for sharing your fix! 🙌

PR #721, opened 2023-12-13
feat: added ability to not include `x-default` alternate link 

lobotomoe:
# Why

> The reserved x-default value is used when no other language/region matches the user's browser setting.

If you want just redirect to defaultLocale language, you can use this approach: https://next-intl-docs.vercel.app/docs/environments/error-files#catching-non-localized-requests

Or, in my case: https://github.com/amannn/next-intl/discussions/720#discussioncomment-7844488

In my case I don't want to include `x-default` because this page is either 404 HTTP + JS redirect (as in docs approach) or 3xx HTTP redirect (as in my approach)

This is unnecessary work for crawlers and **potentially** damage SEO

### DIFF
--- a/docs/pages/docs/routing/middleware.mdx
+++ b/docs/pages/docs/routing/middleware.mdx
@@ -241,6 +241,11 @@ export default createMiddleware({
   // ... other config
 
   alternateLinks: false // Defaults to `true`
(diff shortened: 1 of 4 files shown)


lobotomoe:
@amannn Thank you! :)

PR #725, opened 2023-12-17
docs: remove ts any

themse:


amannn (commented):
Hey, thanks for the proposed docs change!

I'm a bit hesitant with this change as people have different opinions on how props should be typed (inline, separate type, interface, naming, …). 

There are also more cases that are not covered in this PR:

Screenshot 2023-12-18 at 17 25 09

My stance so far was to not include types for props, also leading to slimmer examples in the docs.

Also note that `locale` as a union in the root layout might be misleading, because it is important to [validate](https://next-intl-docs.vercel.app/docs/getting-started/app-router#applocalelayouttsx) the incoming param (which is a `string`).

What was the motivation for you to add this?

themse:
Thank you for the response. The main idea was a desire to avoid the use of 'any' in the code. 
How to write props is at the discretion of the person.  
Regarding validation, I agree, of course, that TypeScript doesn't check dynamic data and doesn't guarantee that the application won't break. It's more about avoiding `any` in the code.

themse:
I know that it's maybe out of scope, however sometimes such examples help the lifes :)

PR #730, opened 2023-12-19
docs: Update Next.js to 14.0.4 in examples

amannn:


examples/example-app-router-migration/package.json
```diff
@@ -10,7 +10,7 @@
     "start": "next start"
   },
   "dependencies": {
-    "next": "^14.0.3",
+    "next": "^14.0.4",
```
A7med3bdulBaset, line 13: Wouldn't this be better for examples?

```suggestion
    "next": "latest",
```
It's used in next.js examples

https://github.com/vercel/next.js/blob/c4adae89b17a6f64cc0abe8e697a88b9e1df1814/examples/api-routes/package.json#L9

amannn: Hi Ahmed! I think it's a good idea to pin the major version of Next.js so we can upgrade in a controlled fashion in case breaking changes are introduced.

PR #747, opened 2023-12-22
docs: add missing `notFound` import

stefanprobst:
this adds a missing import for `notFound` to the recently updated `i18n.ts` examples in the docs.


amannn (approved):
Good catch, thank you so much @stefanprobst!

PR #755, opened 2024-01-03
feat: Add full ESM support via an `import` export condition

amannn:
We currently support ESM partially. This PR investigates switching to bunchee in order to enable full ESM support.

**TODO**
- [ ] Make sure entries share code as expected
- [ ] Provide a separate development build with unminified errors (see https://twitter.com/jantimon/status/1745758514763469047). What are the side effects though? E.g. bundling with jest/vitest might get the prod build (to be verified). What if we rely on the `production` environment instead and ship full errors by default for other entry points? That would result in Next.js getting the smallest build, test runners getting unminified errors—and, users of `use-intl` (e.g. `react-native`) potentially also unminified errors. Should we ship a version where a condition like `typeof process !== 'undefined' && process.env.NODE_ENV === 'production'` is included in the bundle? Maybe ship `development`, `production` as well as `default` with  with an uncompiled condition like `typeof process !== 'undefined' && process.env.NODE_ENV === 'production' ? ` to support everything.
- [ ] Make sure turbo works


packages/next-intl/package.json
```diff
@@ -21,7 +21,7 @@
     "test": "TZ=Europe/Berlin vitest",
     "lint": "pnpm run lint:source && pnpm run lint:package",
     "lint:source": "eslint src test && tsc --noEmit",
-    "lint:package": "publint && attw --pack",
```
amannn, line 24: - [ ] Revert or use `.mjs`?

packages/use-intl/package.json
```diff
@@ -15,7 +15,7 @@
     "test": "TZ=Europe/Berlin vitest",
     "lint": "pnpm run lint:source && pnpm run lint:package",
     "lint:source": "eslint src test && tsc --noEmit",
-    "lint:package": "publint && attw --pack",
```
amannn, line 18: - [ ] Revert or use `.mjs`?

scripts/getBuildConfig.js
```diff
@@ -44,13 +44,16 @@ module.exports = function getConfig({
...
+
   /** @type import('rollup').RollupOptions */
   const config = {
     input,
     external: [/node_modules/, ...external],
     output: {
       dir: outDir + env,
-      format: 'cjs',
+      format,
+      entryFileNames: '[name]' + (format === 'es' ? '.mjs' : '.js'),
```
amannn, line 56: Remove this

packages/next-intl/package.json
```diff
@@ -97,6 +139,7 @@
     "@types/node": "^20.1.2",
     "@types/react": "18.2.34",
     "@types/react-dom": "^18.2.17",
+    "bunchee": "5.0.0-beta.1",
```
amannn, line 142: Code sharing notes:

```
# shared (react context)
dist/production/index.react-client.js
dist/production/navigation.react-client.js

# shared (memo cache)
dist/production/index.react-server.js
dist/production/server.react-server.js
dist/production/navigation.react-server.js

# separate
dist/production/middleware.js
dist/production/routing.js
dist/production/server.react-client.js
```

amannn:
Superseded by https://github.com/amannn/next-intl/pull/1115

PR #758, opened 2024-01-05
feat: Support ISO 8601 date strings with full precision for all formatting functions where dates can be passed

martinmunillas:
This PR adds support to format ISO 8601 directly, this is super useful for date strings coming directly from the backend, it makes sense as it is the [official format accepted in js](https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-date-time-string-format).

As of right now I find myself doing `format.dateTime(new Date(str), format)` which is pretty annoying.

I'm not sure if my changes are enough but I assume it should work, please provide any feedback.

**Example**

```tsx
import {useFormatter} from 'next-intl';

function BlogPost({post}) {
  const format = useFormatter();

  return (
    <article>
      <h1>{post.title}</h1>
      <time datetime={post.publishedAt}>
        {format.dateTime(post.publishedAt)}
      </time>
    </article>
  );
}
```

**Todo**
- [x] Date formatting
- [x] Date range formatting
- [x] Relative date formatting
- [x] Dev-only validation


packages/use-intl/src/core/createFormatter.tsx
```diff
@@ -153,6 +153,22 @@ export default function createFormatter({
           }
         }
 
+        if (typeof value === 'string') {
+          const str = value;
+          value = new Date(value);
```
amannn, line 158: There are two problems that I know of with date strings:

### 1) Optional parts of date strings

> Various components can be omitted, so the following are all valid:
> - Date-only form: YYYY, YYYY-MM, YYYY-MM-DD
> - Date-time form: one of the above date-only forms, followed by T, followed by HH:mm, HH:mm:ss, or HH:mm:ss.sss. Each combination can be followed by a time zone offset.

([source](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#date_time_string_format))

Omitting the time or the time zone part in the input is problematic because by calling the constructor, a time and time zone are implied and the user can format them:

```
// E.g. "1:00:00 AM Central European Standard Time"
formatter.dateTime('2020-11-20', {timeStyle: 'full'})
```

The time zone is especially tricky due to:

> When the time zone offset is absent, date-only forms are interpreted as a UTC time and date-time forms are interpreted as local time. This is due to a historical spec error that was not consistent with ISO 8601 but could not be changed due to web compatibility. See [Broken Parser – A Web Reality Issue](https://maggiepint.com/2017/04/11/fixing-javascript-date-web-compatibility-and-reality/).

Here's an interesting case for you that can be run in the test suite of this repo (uses `TZ=Europe/Berlin` in the Node.js environment when running a test):

```tsx
const formatter = createFormatter({locale: 'en'});
expect(
  formatter.dateTime('2020-11-20', {
    dateStyle: 'medium',
    timeZone: 'America/New_York'
  })
).toBe('Nov 20, 2020'); // ❌ Result: Nov 19, 2020
```

Fun, right? The reason is that the UTC time zone is assumed and the explicitly provided time zone moves the created date the the previous date.

### 2) Using non-standard date strings

The date constructor also [accepts non-standard date strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#non-standard_date_strings) with varying browser support.

Maybe due to this mess `Intl.DateTimeFormat.prototype.format()` only accepts `Date | number`. Hopefully [`Temporal`](https://tc39.es/proposal-temporal/docs/) will become the better alternative at some point in the future.

---

Don't get me wrong, I absolutely see your point and am trying to optimize for convenience with `next-intl`. By moving the date parsing into userland, we currently at least make it obvious how a string is turned into a date. Some users might prefer to [lint against this](https://github.com/amzn/eslint-plugin-no-date-parsing) even.

I think to move forward with this, we'd have to validate that the incoming date string conforms to ISO 8601 and includes all parts (year, month, date, hour, minute, seconds, milliseconds, timezone). That however will increase the bundle size slightly.

I'm honestly not confident about this change currently. Especially `Intl.DateTimeFormat.prototype.format()` not accepting strings makes me suspicious, maybe there are other problems I don't know about yet.

I think as an immediate todo, the `next-intl` docs should explain these problems in more detail in the [Formatting dates and times](https://next-intl-docs.vercel.app/docs/usage/dates-times#formatting-dates-and-times) docs. Currently, there's an expandable section "How can I parse dates or manipulate them?" there. I think it should be split in two and details about parsing should be included.

I'm not saying this will never make it in, but I'd say we should leave the PR open for a bit, give it more thought and maybe consider opinions from other users in case others chime in.

packages/use-intl/test/core/createFormatter.test.tsx
```diff
@@ -1,62 +1,74 @@
...
   });
   expect(
     formatter.dateTime(parseISO('2020-11-20T10:36:01.516Z'), {
-      dateStyle: 'medium'
+      dateStyle: 'medium',
+    })
+  ).toBe('Nov 20, 2020');
+});
+
+it('formats a ISO 8601 datetime string', () => {
```
amannn, line 17: nit:
```suggestion
it('formats an ISO 8601 datetime string', () => {
```

packages/use-intl/src/core/createFormatter.tsx
```diff
@@ -128,7 +128,7 @@ export default function createFormatter({
 
   function dateTime(
     /** If a number is supplied, this is interpreted as a UTC timestamp. */
```
amannn, line 130: This would have to be updated.

amannn (commented):
Thanks for getting the ball rolling here! I currently have some concern about this and would like the leave this open for now, please see the inline comment.

packages/use-intl/src/core/createFormatter.tsx
```diff
@@ -1,10 +1,10 @@
 import DateTimeFormatOptions from './DateTimeFormatOptions';
 import Formats from './Formats';
-import IntlError, {IntlErrorCode} from './IntlError';
+import IntlError, { IntlErrorCode } from './IntlError';
```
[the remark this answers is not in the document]
martinmunillas: yes! my bad

packages/use-intl/src/core/IntlError.tsx
```diff
@@ -5,7 +5,8 @@ export enum IntlErrorCode {
   INSUFFICIENT_PATH = 'INSUFFICIENT_PATH',
   INVALID_MESSAGE = 'INVALID_MESSAGE',
   INVALID_KEY = 'INVALID_KEY',
-  FORMATTING_ERROR = 'FORMATTING_ERROR'
+  INVALID_FORMAT = 'INVALID_FORMAT',
```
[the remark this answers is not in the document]
martinmunillas: I thought so at first but then I thought FORMATTING_ERROR was a formatting error but not an error because of the format while parsing. but if it still makes sense to you I'll revert it

packages/use-intl/src/core/createFormatter.tsx
martinmunillas: I understand point 1 but I don't see how that is any different to what exists now, if i create a date with `new Date(2020, 11, 20)` and format that, the error is the same, is not an issue of string parsing but on how the date object is created, if the date doesn't contain a

About point 2, yes, there is not much to argue there, js Dates are a mess, and there could be possibles bugs as everywhere, but this doesn't change anything for users use other parsing mechanisms, only adds an easier way for users that do use the `new Date(str)` one. I wouldn't also mind to include other date parser if necessary or change the error message and documentation if necessary.

I still understand your concerns and appreciate your feedback! I'm just think it would be super useful to directly accept strings

amannn: > if i create a date with new Date(2020, 11, 20) and format that, the error is the same

That's true. To me, the difference is that this is in userland and hopefully caught in review (or by a linter).

If the user calls `format.dateTime('2020-11-20', {dateStyle: 'medium'})` (from my example above) this looks correct. Even the output "Nov 19, 2020" _looks_ correct—but is off by a day. In this simplified example it's easy to spot but maybe less so in a complex app.

The possibility I see here is validating at runtime that a full ISO 8601 string is passed. I'm unsure if dev-only error handling is a good idea here, and I'd like to avoid increasing the bundle size due to this. We currently don't have this issue if the user calls `new Date` in the app code. Furthermore, this gives the user the chance to include runtime validation code if it's relevant for the app.

I think for the time being I'll wait with this and instead improve the docs on parsing dates. I hope you can understand.

amannn: A closely related note from the recently released `date-fns@3`:

Screenshot 2024-01-29 at 10 18 34

(https://blog.date-fns.org/v3-is-out/)

**Example:**

Screenshot 2024-01-29 at 10 19 15

Might be worth keeping an eye on this, maybe it works out well for `date-fns` or there are learnings for `next-intl` here.

For reference, here's [a blog post on a 2.0 prerelease for `date-fns`](https://blog.date-fns.org/300-bytes/) with more background on string parsing.

amannn: Another related update: [Tempo](https://tempo.formkit.com/) was just released and they support receiving [dates as strings](https://github.com/formkit/tempo/blob/b20d77775bc519f02188f4bdf8b476c3f54c64ba/src/date.ts#L30-L34). They include [runtime validation](https://github.com/formkit/tempo/blob/main/src/iso8601.ts#L13) that will throw in case an invalid date string is encountered. [I've asked](https://twitter.com/jamannnnnn/status/1757691525398413382) the library author for some background on this decision.

Maybe in the end we should really support this pattern. I'd still consider doing the runtime validation only during development to avoid increasing the bundle size.

martinmunillas: let me know if you decide to support it and how to move this pr forward

martinmunillas: Hi @amannn sorry for leaving this a little abandoned, I haven't really had time to get my hands on this again. I'm not sure when I'll be able to, so anyone feel free to take over if they have time, if not I'll finish this, I just don't know when.

amannn: I'll look into this!

amannn: I think the implementation is good now. One tradeoff I noticed: Dates in messages are required to be actual dates, strings will lead to a formatting error. So there's some slight divergence there.

amannn: Another tradeoff I just discovered: With https://next-intl.dev/blog/next-intl-4-0#strictly-typed-icu-arguments, we would then also allow any string to be passed to the `date` formatter.

It's a bit a tradeoff: Simpler (avoids parsing in userland), but maybe more error-prone (any string is accepted)

PR #765, opened 2024-01-09
fix: Change `numeric` option from `auto` to `always` for `format.relativeTime`

### LINKED ISSUE #757: formatter.relativeTime outputs wrong relative time
### Description

I just noticed on my blog the relative time is wrong.

In my case I have this date on my post: "2022-07-11"
The relativeTime output is: "Last year"

This is my i18n.ts:

```ts
import { type AbstractIntlMessages } from 'next-intl';
import { getRequestConfig } from 'next-intl/server';

export default getRequestConfig(async ({ locale }) => ({
  messages: (
    (await import(`./messages/${locale}.json`)) as {
      default: AbstractIntlMessages;
    }
  ).default,
  timeZone: 'Europe/Berlin',
  now: new Date(),
}));
```

My component looks like this
```tsx
export default async function Blog({
  params,
}: {
  params: { slug: string; locale: string };
}) {
  unstable_setRequestLocale(params.locale);

  const now = await getNow({ locale: params.locale });
  const formatter = await getFormatter({ locale: params.locale });

return (
...
<p className="ml-2 text-sm text-gray-700 dark:text-[#c2c2c2]">
            {formatter.dateTime(new Date(post.publishedAt), {
              year: 'numeric',
              month: 'short',
              day: 'numeric',
            })}{' '}
            ({formatter.relativeTime(new Date(post.publishedAt), now)})
          </p>
...)}
```

You can preview the page (here)[https://alexanderkonietzko.vercel.app/en/blog/vercel-deployment]

Did I make a mistake here?

https://github.com/alex289/Portfolio

### Reproduction description

Steps to reproduce:
1. Open reproduction
2. Go to the post page
3. See the wrong relative time

### Expected behaviour

Since "2022-07-11" is more than one year ago I expected something like "2 years ago"

amannn:
Fixes #757

Previously, [`numeric: 'auto'`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat/RelativeTimeFormat#using_the_auto_option) was used for formatting relative times. This occasionally produces incorrect output when values are rounded based on the unit, therefore it's changed to `always` now. The only exception are cases <1s which render as "now" via `numeric: 'auto'` (instead of "0 seconds ago").

**Previous behavior:**

```tsx
// Output: "Last year"
format.relativeTime(
  new Date("2022-07-11T00:00:00.000Z"),
  new Date("2024-01-17T00:00:00.000Z")
);
```

**New behavior:**

```tsx
// Output: "1 year ago"
format.relativeTime(
  new Date("2022-07-11T00:00:00.000Z"),
  new Date("2024-01-17T00:00:00.000Z")
);
```

Note that this is considered as "1 year" due to rounding. Once a difference of 1.5 years is reached, then "2 years ago" will be displayed.

### DIFF
--- a/docs/pages/docs/usage/configuration.mdx
+++ b/docs/pages/docs/usage/configuration.mdx
@@ -15,7 +15,7 @@ Depending on if you handle [internationalization in Client- or Server Components
 `i18n.ts` can be used to provide configuration for **Server Components**.
 
 ```tsx filename="i18n.ts"
-import {notFound} from "next/navigation";
+import {notFound} from 'next/navigation';
 import {getRequestConfig} from 'next-intl/server';
 
 // Can be imported from a shared config
@@ -462,9 +462,12 @@ function getMessageFallback({namespace, key, error}) {
   }
 }
 
-<NextIntlClientProvider onError={onError} getMessageFallback={getMessageFallback}>
+<NextIntlClientProvider
+  onError={onError}
+  getMessageFallback={getMessageFallback}
+>
   ...
-</NextIntlClientProvider>
+</NextIntlClientProvider>;
 ```
 
 </Tab>
@@ -504,11 +507,12 @@ return (
 As a convenience, there are a couple of hooks that allow you to read global configuration.
 
 ```tsx
-import {useLocale, useTimeZone, useMessages} from 'next-intl';
+import {useLocale, useTimeZone, useMessages, useNow} from 'next-intl';
 
 function Component() {
   const locale = useLocale();
   const timeZone = useTimeZone();
   const messages = useMessages();
+  const now = useNow();
 }
 ```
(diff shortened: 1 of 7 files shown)


brianguertin:
Hmmm, this is not a fix at all, it's disabling very useful behaviour. I think most (English) users would prefer "today" over "0 days ago". And "yesterday" to "1 day ago".

Are library users supposed to fall back to manually using the browser API to format relative times if they want numeric: auto?

amannn:
Can you provide details for which case you received "0 days ago"? There's an [extensive test suite](https://github.com/amannn/next-intl/blob/676debe9f577d1225599e889e9f1700acf460ef7/packages/use-intl/src/core/createFormatter.test.tsx#L162-L211) for this feature.

> And "yesterday" to "1 day ago"

I generally agree, however the motivation for this feature was that "yesterday" as produced with `auto` is unfortuantely not always "yesterday". Rounding gets in the way here, please see the PR description and the [code comment](https://github.com/amannn/next-intl/pull/765/files#diff-5d85f0efd15bf92cc19e164678e48f1ed4d4323a64a5dd6f3882ebe9c766e1c2).

Issue #540, opened 2023-10-02
Add support for number range formatting


vimtor:
### Is your feature request related to a problem? Please describe.

I want to format a range of numbers using the existing number range format API:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/formatRange

### Describe the solution you'd like

Either an option to the `.number` method in the `Formatter` object or a new `numberRange` method.

### Describe alternatives you've considered

I've considered using the API directly, but it is only supported in Node.js 19. The latest version available in Vercel is Node.js 18.

vimtor:
Yes, @amannn! That would be a great addition :)

Regarding the polyfill, I assume the polyfill will be included with the library, right? If that is the case, it would work 100%

vimtor:
Yes! I would love to contribute to this feature.

Would it be better to wait until the RSC integration is merged, right?

amannn:
Superseded by https://github.com/amannn/next-intl/issues/774. We can still add this if there's need for this functionality.

PR #769, opened 2024-01-10
feat: Add `format.dateTimeRange`

martinmunillas:
Contributes to https://github.com/amannn/next-intl/issues/774

Add support for [Intl.DateTimeFormat.prototype.formatRange()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/formatRange)


packages/use-intl/src/core/createFormatter.tsx
```diff
@@ -158,6 +158,40 @@ export default function createFormatter({
...
+            onError(
+              new IntlError(
+                IntlErrorCode.ENVIRONMENT_FALLBACK,
+                process.env.NODE_ENV !== 'production'
+                  ? `The \`timeZone\` parameter wasn't provided and there is no global default configured. Consider adding a global default to avoid markup mismatches caused by environment differences. Learn more: https://next-intl-docs.vercel.app/docs/configuration#time-zone`
+                  : undefined
+              )
+            );
+          }
+        }
```
amannn, line 188: Maybe we can share this part with `dateTime` to reduce bundle size.

```suggestion
        if (!options?.timeZone) {
          options = applyGlobalTimeZone(options);
        }
```

… with:

```tsx
  function applyGlobalTimeZone(options?: DateTimeFormatOptions) {
    if (globalTimeZone) {
      options = {...options, timeZone: globalTimeZone};
    } else {
      onError(
        new IntlError(
          IntlErrorCode.ENVIRONMENT_FALLBACK,
          process.env.NODE_ENV !== 'production'
            ? `The \`timeZone\` parameter wasn't provided and there is no global default configured. Consider adding a global default to avoid markup mismatches caused by environment differences. Learn more: https://next-intl-docs.vercel.app/docs/configuration#time-zone`
            : undefined
        )
      );
    }
    return options;
  }
```

packages/use-intl/test/core/createFormatter.test.tsx
```diff
@@ -253,6 +253,42 @@ describe('relativeTime', () => {
...
+        new Date(Date.UTC(1906, 0, 10, 11, 0, 0)), // Wed, 10 Jan 1906 11:00:00 GMT
+        {
+          year: '2-digit',
+          month: 'numeric',
+          day: 'numeric',
+          hour: 'numeric',
+          minute: 'numeric'
+        }
+      )
+    // 1 hour more given that the timezone is Europe/Berlin and the date is in UTC
```
amannn, line 287: The linter will probably show an error here.

packages/use-intl/src/core/createFormatter.tsx
```diff
@@ -158,6 +158,40 @@ export default function createFormatter({
     );
   }
 
+  function dateTimeRange(
```
amannn, line 161: Would you like to add docs for this too? See https://github.com/amannn/next-intl/blob/main/docs/pages/docs/usage/dates-times.mdx

packages/use-intl/src/core/createFormatter.tsx
```diff
@@ -158,6 +158,40 @@ export default function createFormatter({
...
+    /** If a number is supplied, this is interpreted as a UTC timestamp. */
+    start: Date | number,
+    /** If a number is supplied, this is interpreted as a UTC timestamp. */
+    end: Date | number,
+    /** If a time zone is supplied, the `value` is converted to that time zone.
+     * Otherwise the user time zone will be used. */
+    formatOrOptions?: string | DateTimeFormatOptions
+  ) {
+    return getFormattedValue(
+      [start, end],
```
amannn, line 171: This argument is used by `getFormattedValue` to produce a fallback value in case formatting fails (e.g. due to an unknown format that is supplied).

Maybe we can add a test here to verify the output:

```tsx
  it('returns a reasonable fallback if an invalid format is provided', () => {
    const formatter = createFormatter({
      locale: 'en',
      timeZone: 'Europe/Berlin'
    });
    expect(
      formatter.dateTimeRange(
        new Date(2007, 0, 10, 10, 0, 0),
        new Date(2008, 0, 10, 11, 0, 0),
        'unknown'
      )
    ).toBe('TODO');
  });
```

Currently, on my machine it outputs:

```
Wed Jan 10 2007 10:00:00 GMT+0100 (Mitteleuropäische Normalzeit),Thu Jan 10 2008 11:00:00 GMT+0100 (Mitteleuropäische Normalzeit)
```

It uses `process.env.LANG` and `process.env.TZ` to produce this output.

I think it's moderately ok for a fallback, but an issue here is that it could lead to hydration mismatches across the server and client if the locale and time zone is different. I think we should settle for something stable.

Maybe we could reuse `dateTime` and use this:

```tsx
// Prints "1/10/2007 – 1/10/2008" in my test from above
[dateTime(start), dateTime(end)].join(' – '),
```

Another optimization on top could be to avoid computing the fallback upfront here and change this argument of `getFormattedValue` to a function to lazily execute on demand. We can also reorder the arguments of `getFormattedValue` to move the fallback function last.

Does this make sense? What do you think?

packages/use-intl/test/core/createFormatter.test.tsx
```diff
@@ -253,6 +253,42 @@ describe('relativeTime', () => {
...
+        new Date(2007, 0, 10, 10, 0, 0),
+        new Date(2008, 0, 10, 11, 0, 0),
+        {
+          weekday: 'long',
+          year: 'numeric',
+          month: 'long',
+          day: 'numeric'
+        }
+      )
+    ).toBe('Wednesday, January 10, 2007 – Thursday, January 10, 2008');
```
amannn, line 273: Thanks for adding tests!

amannn (commented):
Thanks for the great PR! I think this is a good idea.

There was related discussion about number range formatting in https://github.com/amannn/next-intl/issues/540. I've now added https://github.com/amannn/next-intl/issues/774 to get an overview about the formatting APIs from `Intl` that `next-intl` currently supports.

martinmunillas:
sorry I haven't looked back at this pr

packages/use-intl/src/core/createFormatter.tsx
martinmunillas: yeah sure!

packages/use-intl/test/core/createFormatter.test.tsx
martinmunillas: what is the correct way to add comments?

martinmunillas: or do you want me to just delete it?

packages/use-intl/src/core/createFormatter.tsx
martinmunillas: yess

packages/use-intl/test/core/createFormatter.test.tsx
amannn: The indent is off and therefore [the build fails](https://github.com/amannn/next-intl/actions/runs/7479432434/job/20388966214?pr=769), everything else is fine!

packages/use-intl/src/core/createFormatter.tsx
amannn: There's been an update on the `main` branch in regard to this: `getFormattedValue` now no longer accepts a `value` but instead a `getFallback` function. You might want to rebase on the latest changes when continuing with this.

PR #772, opened 2024-01-11
fix: update vscode-integration

NilsJacobsen:
### What I changed

- Changed `Inlang` to `inlang` according to brand guideline
- Added `?ref=next-intl` to the links that point to inlang.com
- Added inlang CLI as an ecosystem example


amannn (approved):
Looks good, thanks!

Issue #756, opened 2024-01-03
Share message format cache between `getTranslations` and `useTranslations` in RSC


amannn:
### Is your feature request related to a problem? Please describe.

The RSC implementation of `useTranslations` already uses it:

https://github.com/amannn/next-intl/blob/e90eda6d85867f87d941ebba10d0bcf6170c5e81/packages/next-intl/src/react-server/getTranslator.tsx#L109

This should be used for `getTranslations` too:

https://github.com/amannn/next-intl/blob/e90eda6d85867f87d941ebba10d0bcf6170c5e81/packages/next-intl/src/server/react-server/getTranslations.tsx#L216

### Describe the solution you'd like

Add a test that verifies the cache is shared and add the property for the `getTranslations`.

PR #798, opened 2024-01-21
feat: Share message format cache between `getTranslations` and `useTranslations`

fkapsahili:
I thought this would be a good starting point as a first time contributor. Just let me know if you had a different idea of how to test the shared cache 🙂. I just made sure that the cache is actually shared and the newly added test would fail without the `messageFormatCache` property.

### Prerequisites

- [x] All commits follow the [conventional commits specification](https://www.conventionalcommits.org/en/v1.0.0/)

### Content
- Added a test that verifies the cache is shared between `useTranslations` and `getTranslations`
- Added the `messageFormatCache` to the `getTranslations`

#### Tests
- Added a test that veriifes the cache is shared

Closes https://github.com/amannn/next-intl/issues/756


packages/next-intl/test/react-server/index.test.tsx
```diff
@@ -57,4 +71,22 @@ describe('performance', () => {
     // Render 3: Recursive call that renders synchronously as well
     expect(renderCount).toBe(3);
   });
+
+  it('shares message format cache between useTranslations and getTranslations', async () => {
+    // First invocation
+    useTranslations('Component');
+    const firstCallCache =
+      vi.mocked(createTranslator).mock.calls[0][0].messageFormatCache;
```
amannn, line 79: I think the test strikes a good balance by testing the instance of the `messageFormatCache`. Actually rendering a translation will be a bit more involved. From what I've seen async components can't be rendered in React 18.2, we'd have to upgrade to canary in the dev dependencies of `next-intl` (which seems to require other changes).

So tldr;—yep, I think this test is a good idea!

packages/next-intl/test/react-server/index.test.tsx
```diff
@@ -57,4 +71,22 @@ describe('performance', () => {
...
+
+    // Second invocation with a different namespace
+    await getTranslations('Component2');
+    const secondCallCache =
+      vi.mocked(createTranslator).mock.calls[1][0].messageFormatCache;
+
+    // Verify that the same cache instance is used in both invocations
+    expect(firstCallCache).toBe(secondCallCache);
```
amannn, line 87: It seems like the test accidentally passes currently despite us having two separate caches:

1. https://github.com/fkapsahili/next-intl/blob/f1fc3ddd9f1b2919ca0b750c0eda682477bcd4f7/packages/next-intl/src/server/react-server/getTranslations.tsx#L17

2. https://github.com/fkapsahili/next-intl/blob/f1fc3ddd9f1b2919ca0b750c0eda682477bcd4f7/packages/next-intl/src/react-server/getTranslator.tsx#L15

That is due to way the mock implementation of `cache` currently works:

https://github.com/amannn/next-intl/blob/716d71605c2b855c6c6004e0c233bbc20ddeaf74/packages/next-intl/__mocks__/react.tsx#L29-L31

Maybe we should use a `WeakMap` for the cache key there to make sure we use a different cache for different functions?

Based on this, this test should hopefully fail then and we can fix it by actually using the same message format cache instance.

Does that make sense?

amannn (commented):
Thank you so much for looking into this @fkapsahili! 🙌

packages/next-intl/test/react-server/index.test.tsx
fkapsahili: Thanks for pointing this out - this definitely makes sense! The current mock setup indeed leads to the situation where the test passes despite the usage of separate cache implementations.

Using a WeakMap for the cache key, worked perfectly as you suggested. I think this will help ensure that the cache behaves more realistically in the test environment. This change also made the test fail as expected 🙂.

I've now added an implementation using the `WeakMap` in the mock cache and added a shared cache instance using `messageFormatCache` to make sure both `getTranslations` and `useTranslations` actually use the same message format cache instance, as intended. I've updated the PR with these improvements. I thought it would make sense to locate the shared cache in the new file `messageFormatCache.tsx` in the `shared` directory.

Let me know if there is anything to improve! 🙂

amannn (approved):
Excellent PR, anytime again! 🙌

Issue #454, opened 2023-08-14
Opt out of using cookies


Kinbaum:
### Is your feature request related to a problem? Please describe.

Similar to #366, but I would like to have the option to disable the use of cookies for the locale.

### Describe the solution you'd like

Adding an option to ﻿`createMiddleware` in `﻿next-intl` to disable the use of cookies for locale and derive the value only from the Accept-Language header or URL itself could have several benefits:

1. Enhanced Privacy: Some users or locales may have stringent privacy regulations around the use of cookies and may prefer or require a cookie-less solution. This feature will respect those privacy boundaries.
1. Flexibility: It gives developers greater control over how locales are determined and how user data is handled. Some applications may prefer to base localization purely on URL structure or Accept-Language headers. This avoids potential discrepancies between cookie locale data and those other factors.
1. Security: Reducing reliance on cookies can help mitigate certain security risks associated with them, such as Cross-Site Scripting (XSS) and Cross-Site Request Forgery (CSRF).

### Describe alternatives you've considered

Attempt to delete the cookie myself in the middleware with each request.

amannn:
Thanks for raising this!

As you've correctly mentioned, this can currently be implemented by composing the middleware:

```tsx
import createIntlMiddleware from 'next-intl/middleware';
import {NextRequest} from 'next/server';
 
export default async function middleware(request: NextRequest) {
  const handleI18nRouting = createIntlMiddleware({
    locales: ['en', 'de'],
    defaultLocale: 'en'
  });
  const response = handleI18nRouting(request);
 
  if (response.cookies.get('NEXT_LOCALE')) {
    response.cookies.delete('NEXT_LOCALE');
  }
 
  return response;
}
 
export const config = {
  matcher: ['/((?!_next|.*\\..*).*)']
};
```

In case there's a growing demand for this, I think we could consider built-in support. I haven't yet thought out all implications this might have though (e.g. we need additional guardrails when someone tries to use this together with `localePrefix: 'never'`).

fernandojbf:
the delete next locale will continue to send the cookie empty header. Would be nice to have the opt out feature.

![image](https://github.com/amannn/next-intl/assets/9076868/0cf46d2f-0ad0-4606-b8dd-9fb888db27b3)

amannn:
Setting the `NEXT_LOCALE` cookie on the response will now be disabled if `localeDetection: false` is set as of `next-intl@3.1.3` (see https://github.com/amannn/next-intl/pull/654).

[...]

PR #817, opened 2024-01-25
fix: Set cookie to `SameSite: Lax`

### LINKED ISSUE #527: Always resetting the cookie when a PWA is installed
### Description

Based on the [docs](https://next-intl-docs.vercel.app/docs/routing/middleware#locale-detection), the main strategy to detect the locale is:

1. A locale prefix is present in the pathname (e.g. /de/about)
2. **A cookie is present that contains a previously detected locale**
3. The [`accept-language header`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language) is matched against the available locales
5. The `defaultLocale` is used

This happens very well when accessing a browser by URL.

But the strategy fails if I access my app, **which is a PWA**, from installed on mobile.
The `NEXT_LOCALE` cookie is resetting when the app opens based on the `accept-language header`, even if I have this cookie value already. 

**Debugging my phone**
If I access the root URL `/`, the previous `NEXT_LOCALE` is ignored. The locale is always `en` because I'm using it as the default language on my phone.

There is no problem with the app, only the `NEXT LOCALE` cookie is resetting. Other cookies are preserved and retrieved. In case I change my cellphone locale to Portuguese, the cookie is always `pt` no matter what.

Screenshot 2023-09-22 at 20 15 13

**Important**
- `"next": "14.1.0"`
- `"next-intl": "3.5.1"`
- `"@serwist/next': '8.4.3"`
- `'@serwist/sw': '8.4.3'`
- It was tested with 2 Android phones with Chrome with the same result.
- Before some `next-intl` update (I don't know which version) it did not happen. 

https://github.com/maiconsanson/next-intl-pwa-ignoring-cookie

### Reproduction description

Steps to reproduce:
1. Configure your mobile device to the English locale `en`.
2. **Install the PWA (mobile) on Chrome**.
3. Open the app and change the locale by clicking on the link "Mude para Português".
3. Close the PWA.
4. Open the app again and notice that the locale is `en`, not the previous `pt` selected.

Repeat these steps directly on the browser and you seed that works normally.

Here is the production link to see in action:
https://intl-unable-find-locale.vercel.app/

### Expected behaviour

The expected behavior is that the saved `NEXT_LOCALE` cookie value be used to define language when the PWA is installed.

### LINKED ISSUE #811: Locale is reseted from selected to accept-language's locale when localePrefix is 'never'
### Description

When user navigates from external site to our site then locale is reseted to accept-language's locale because NEXT_LOCALE cookie has a `SameSite: Strict` and browser don't send cookie in this case. Also `localePrefix` is `never` and thats why next-intl set locale to accept-language's locale (or to default locale).

This is not reproducible if `Lax` is used.

-

### Reproduction description

I can't give you reproduction example, I've tried but it's not working (basic functional). Unfortunately, I don't have time to figure out why.

### Expected behaviour

Shouldn't reset  the locale.

amannn:
See https://web.dev/articles/samesite-cookies-explained#explicitly_state_cookie_usage_with_the_samesite_attribute for the rationale behind this.

Fixes #527
Fixes #811


corolair-hazem:
Is there a way i can set it to none, i need this because my app is used in an iframe with language switcher, please help

amannn:
@corolair-hazem Please see https://github.com/amannn/next-intl/issues/454

PR #818, opened 2024-01-26
Update app-router.mdx

shumakmanohar:
Update docs For without src folder ,


amannn (commented):
Thanks for the suggestion! This is addressed as "Can I move this file somewhere else?" in the next section, I believe it's easier to explain there since we're introducing the `i18n.ts` file in that section too. Hope that makes sense!

amannn:
Closing due to inactivity …

PR #853, opened 2024-02-09
docs: Update `LocaleSwitcherSelect` in `example-app-router` to match docs

### LINKED ISSUE #581: Typescript errors using new localized path apis
### Description

I have implemented the new localized pathnames for server components in my nextjs app, which is 👏 . But the one thing that I am having trouble with is that I am getting typescript errors. 

From the error, it seems that the library is expecting a specific set of params, but is receiving generics. 

For the record, the code still works as expected, so the only problem is the error (which I am just ignoring for now).

Here is my abbreviated code

```
  const pathname = usePathname();
  const router = useRouter();
  const params = useParams();
  
  return (
    <a href="#"  onClick={() => router.replace({ pathname, params }, { locale: "de" })}>Switch to German </a>
  )
```

And the TS error

> Argument of type '{ pathname: "/" | "/contact" | "/cookies" | "/depression-test" | "/find-a-therapist" | "/find-a-therapist/[city_slug]" | "/find-a-therapist/[city_slug]/[profile_slug]" | "/find-help-with/[category_slug]" | ... 13 more ... | "/profile/[uri]"; params: Record<...> | null; }' is not assignable to parameter of type '"/" | "/contact" | "/cookies" | "/depression-test" | "/find-a-therapist" | "/for-companies" | "/for-therapists" | "/imprint" | "/our-story" | "/press" | "/privacy" | "/resources" | ... 26 more ... | ({ ...; } & { ...; })'.
  Types of property 'params' are incompatible.
    Type 'Record<string, string | string[]> | null' is not assignable to type '{ city_slug: ParamValue; } | { city_slug: ParamValue; profile_slug: ParamValue; } | { category_slug: ParamValue; } | { language: ParamValue; } | { ...; } | undefined'.
      Type 'null' is not assignable to type '{ city_slug: ParamValue; } | { city_slug: ParamValue; profile_slug: ParamValue; } | { category_slug: ParamValue; } | { language: ParamValue; } | { ...; } | undefined'.ts(2345)
      
      
      

https://codesandbox.io/p/sandbox/next-intl-bug-template-app-forked-zcymvq

### Reproduction description

Implement the new localized path navigation and make sure to have wildcard params.

Implement the recommended method for changing the locale and maintaining the current page

```
import {useParams} from 'next/navigation';
import {usePathname, useRouter} from '../navigation';
 
const pathname = usePathname();
const router = useRouter();
const params = useParams();
 
router.replace({pathname, params}, {locale: 'de'});
```

Notice the typescript warning for params.

### Expected behaviour

I would think we would not receive typescript warnings for a recommended implementation. but if there is no easy way aound it, then perhaps its worth mentioning it in the docs?

CodyDuff:
Updates LocaleSwitcherSelect.tsx in examples/example-app-router to match the doc changes in commit [de285f6](https://github.com/amannn/next-intl/commit/de285f6f78de15c51a62e2cdd29e0debcc2d415b), which fixed #581.

### DIFF
--- a/examples/example-app-router/src/components/LocaleSwitcherSelect.tsx
+++ b/examples/example-app-router/src/components/LocaleSwitcherSelect.tsx
@@ -25,13 +25,10 @@ export default function LocaleSwitcherSelect({
     const nextLocale = event.target.value;
     startTransition(() => {
       router.replace(
-        {
-          pathname,
-          // TypeScript validates that only known `params` are used in combination
-          // with a given `pathname`. Since the two will always match for the current
-          // route, we can skip runtime checks.
-          params: params as any
-        },
+        // @ts-expect-error -- TypeScript will validate that only known `params`
+        // are used in combination with a given `pathname`. Since the two will
+        // always match for the current route, we can skip runtime checks.
+        { pathname, params },
         {locale: nextLocale}
       );
     });
--- a/examples/example-app-router/src/components/LocaleSwitcherSelect.tsx
+++ b/examples/example-app-router/src/components/LocaleSwitcherSelect.tsx
@@ -1,8 +1,8 @@
 'use client';
 
 import clsx from 'clsx';
-import {ChangeEvent, ReactNode, useTransition} from 'react';
 import {useParams} from 'next/navigation';
+import {ChangeEvent, ReactNode, useTransition} from 'react';
 import {useRouter, usePathname} from '../navigation';
 
 type Props = {
@@ -28,7 +28,7 @@ export default function LocaleSwitcherSelect({
         // @ts-expect-error -- TypeScript will validate that only known `params`
         // are used in combination with a given `pathname`. Since the two will
         // always match for the current route, we can skip runtime checks.
-        { pathname, params },
+        {pathname, params},
         {locale: nextLocale}
       );
     });
--- a/examples/example-app-router/src/components/LocaleSwitcherSelect.tsx
+++ b/examples/example-app-router/src/components/LocaleSwitcherSelect.tsx
@@ -2,6 +2,7 @@
 
 import clsx from 'clsx';
 import {ChangeEvent, ReactNode, useTransition} from 'react';
+import {useParams} from 'next/navigation';
 import {useRouter, usePathname} from '../navigation';
 
 type Props = {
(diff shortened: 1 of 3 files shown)


examples/example-app-router/src/components/LocaleSwitcherSelect.tsx
```diff
@@ -18,11 +19,21 @@ export default function LocaleSwitcherSelect({
...
     const nextLocale = event.target.value;
     startTransition(() => {
-      router.replace(pathname, {locale: nextLocale});
+      router.replace(
+        {
+          pathname,
+          // TypeScript validates that only known `params` are used in combination
+          // with a given `pathname`. Since the two will always match for the current
+          // route, we can skip runtime checks.
+          params: params as any
```
amannn, line 33: The situation is slightly different here: Since there are no dynamic params in this example, there can never be any `params`. However, I'd agree that it's beneficial if we show a sample that works for either case.

This would work:

```suggestion
          // @ts-expect-error -- TypeScript validates that only known `params` are used in combination
          // with a given `pathname`. Since the two will always match for the current
          // route, we can skip runtime checks.
          params
```

That would work in the example in the docs too. We could update the docs accordingly so these are in sync.

What do you think?

amannn (commented):
Thanks for the suggestion @CodyDuff!

examples/example-app-router/src/components/LocaleSwitcherSelect.tsx
CodyDuff: Oh ok, I understand. Whatever you want also works for me. I had just stumbled onto this issue because I started a project using example-app-router as my starting point and eventually created a dynamic route causing the error.

Your suggested placement of the `@ts-expect-error` did not dismiss the error. I actually had to place the `@ts-expect-error` comment above the entire `{ pathname, params }` object like this:
```
startTransition(() => {
  router.replace(
    // @ts-expect-error -- TypeScript will validate that only known `params`
    // are used in combination with a given `pathname`. Since the two will
    // always match for the current route, we can skip runtime checks.
    { pathname, params },
    { locale: nextLocale },
  );
});
```
Does this work for you?

Also, when you say "We could update the docs accordingly so these are in sync", are you saying the docs at `docs/pages/docs/routing/navigation.mdx` should include the `@ts-expect-error` comment as well?

amannn: Sure, that looks good! Yep, would you mind updating `navigation.mdx` to match the updated implementation in `example-app-router`?

amannn (approved):
Thank you so much! 🙌

PR #856, opened 2024-02-11
docs: Update Supabase auth example to use `@supabase/ssr`

flixlix:
created an example for usage with the new supabase ssr package. The middleware I created is working in development as well as in production. The matcher in `middleware.ts` Was adapted based on multiple metadata files that needed to be ignored.

Suggestions on changes are welcome.

image

see `https://supabase.com/docs/guides/auth/auth-helpers/nextjs?language=ts` for the recommendation from supabase.

### DIFF
--- a/docs/pages/docs/routing/middleware.mdx
+++ b/docs/pages/docs/routing/middleware.mdx
@@ -549,29 +549,113 @@ export const config = {
 
 ### Example: Integrating with Supabase Authentication
 
-[`@supabase/auth-helpers-nextjs`](https://supabase.com/docs/guides/auth/auth-helpers/nextjs?language=ts#managing-session-with-middleware) exports `createMiddlewareClient` that accepts a `req` and `res` object. We can pass the incoming `req` and the `res` created with `createIntlMiddleware` to it.
+[`@supabase/ssr`](https://supabase.com/docs/guides/auth/server-side/creating-a-client) exports a `createServerClient` function. First we're going to create a utils function that invokes the `createServerClient` function and returns the `supabase` object.
+
+```ts filename="utils/supabase/middleware.ts"
+import {createServerClient, type CookieOptions} from '@supabase/ssr';
+import {type NextRequest, NextResponse} from 'next/server';
+
+export const createClient = (request: NextRequest) => {
+  // Create an unmodified response
+  let response = NextResponse.next({
+    request: {
+      headers: request.headers
+    }
+  });
+
+  const supabase = createServerClient(
+    process.env.NEXT_PUBLIC_SUPABASE_URL!,
+    process.env.NEXT_PUBLIC_SUPABASE_ANON_KEY!,
+    {
+      cookies: {
+        get(name: string) {
+          return request.cookies.get(name)?.value;
+        },
+        set(name: string, value: string, options: CookieOptions) {
+          // If the cookie is updated, update the cookies for the request and response
+          request.cookies.set({
+            name,
+            value,
+            ...options
+          });
+          response = NextResponse.next({
+            request: {
+              headers: request.headers
+            }
+          });
+          response.cookies.set({
+            name,
+            value,
+            ...options
+          });
+        },
+        remove(name: string, options: CookieOptions) {
+          // If the cookie is removed, update the cookies for the request and response
+          request.cookies.set({
+            name,
+            value: '',
+            ...options
+          });
+          response = NextResponse.next({
+            request: {
+              headers: request.headers
+            }
+          });
+          response.cookies.set({
+            name,
+            value: '',
+            ...options
+          });
+        }
+      }
+    }
+  );
+
+  return {supabase, response};
+};
+```
+
+Next, we're going to set up two required environment variables in a `.env.local` file. The `NEXT_PUBLIC_SUPABASE_URL` and `NEXT_PUBLIC_SUPABASE_ANON_KEY` are required to create the `supabase` object.
+
+```txt filename=".env.local"
+NEXT_PUBLIC_SUPABASE_URL=your_supabase_project_url
+NEXT_PUBLIC_SUPABASE_ANON_KEY=your_supabase_anon_key
+```
+
+Finally, we're going to create a middleware file that uses the `createClient` function that we created and merges it with the `next-intl` middleware.
 
 ```tsx filename="middleware.ts"
-import {createMiddlewareClient} from '@supabase/auth-helpers-nextjs';
+import {NextResponse, type NextRequest} from 'next/server';
+import {createClient} from '@/utils/supabase/middleware';
 import createIntlMiddleware from 'next-intl/middleware';
-import type {NextRequest} from 'next/server';
-import type {Database} from '@/lib/database.types';
 
-export default async function middleware(req: NextRequest) {
+const defaultLocale = 'en';
+const locales = ['en', 'de'];
+
+export async function middleware(request: NextRequest) {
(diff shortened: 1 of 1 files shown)


docs/pages/docs/routing/middleware.mdx
```diff
@@ -549,29 +549,113 @@ export const config = {
...
 ### Example: Integrating with Supabase Authentication
 
-[`@supabase/auth-helpers-nextjs`](https://supabase.com/docs/guides/auth/auth-helpers/nextjs?language=ts#managing-session-with-middleware) exports `createMiddlewareClient` that accepts a `req` and `res` object. We can pass the incoming `req` and the `res` created with `createIntlMiddleware` to it.
+[`@supabase/ssr`](https://supabase.com/docs/guides/auth/server-side/creating-a-client) exports a `createServerClient` function. First we're going to create a utils function that invokes the `createServerClient` function and returns the `supabase` object.
+
+```ts filename="utils/supabase/middleware.ts"
+import {createServerClient, type CookieOptions} from '@supabase/ssr';
+import {type NextRequest, NextResponse} from 'next/server';
+
+export const createClient = (request: NextRequest) => {
```
amannn, line 558: Based on the Supabase guide, I think we should accept a `response` here, where we can can assign cookies on directly (or if we inline this function, just use the `response` from the closure).

docs/pages/docs/routing/middleware.mdx
```diff
@@ -549,29 +549,113 @@ export const config = {
...
+  // Create an unmodified response
+  let response = NextResponse.next({
```
amannn, line 560: This shouldn't be necessary, we should use the `response` from `handleI18nRouting`.

docs/pages/docs/routing/middleware.mdx
```diff
@@ -549,29 +549,113 @@ export const config = {
...
+          return request.cookies.get(name)?.value;
+        },
+        set(name: string, value: string, options: CookieOptions) {
+          // If the cookie is updated, update the cookies for the request and response
+          request.cookies.set({
+            name,
+            value,
+            ...options
+          });
+          response = NextResponse.next({
```
amannn, line 581: This looks a bit suspicious to me. I saw it's mentioned in the Supabase docs too, but I'm wondering why this is necessary and also if it could break something. Do you think we can remove this?

docs/pages/docs/routing/middleware.mdx
```diff
@@ -549,29 +549,113 @@ export const config = {
...
+          });
+        },
+        remove(name: string, options: CookieOptions) {
+          // If the cookie is removed, update the cookies for the request and response
+          request.cookies.set({
+            name,
+            value: '',
+            ...options
+          });
+          response = NextResponse.next({
```
amannn, line 599: Similar question as above.

docs/pages/docs/routing/middleware.mdx
```diff
@@ -549,29 +549,113 @@ export const config = {
...
-import {createMiddlewareClient} from '@supabase/auth-helpers-nextjs';
+import {NextResponse, type NextRequest} from 'next/server';
+import {createClient} from '@/utils/supabase/middleware';
 import createIntlMiddleware from 'next-intl/middleware';
-import type {NextRequest} from 'next/server';
-import type {Database} from '@/lib/database.types';
 
-export default async function middleware(req: NextRequest) {
+const defaultLocale = 'en';
+const locales = ['en', 'de'];
```
amannn, line 633: If we don't need these values anywhere else except from `createIntlMiddleware`, I'd inline it their for simplicity.

amannn (commented):
Thank you so much for proposing an update here! 🙌

docs/pages/docs/routing/middleware.mdx
```diff
@@ -549,29 +549,113 @@ export const config = {
...
+        headers: request.headers
+      }
+    });
+  }
 }
 
 export const config = {
-  matcher: ['/', '/(de|en)/:path*']
+  matcher: [
+    '/((?!_next/static|_next/image|favicon.ico|assets|icon|sitemap.xml|robots.txt|auth|api).*)'
```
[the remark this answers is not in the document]
flixlix: the only problem I found with this approach is that, when adding a new language, this matcher would need to be upadted, although the language options are already defined elsewhere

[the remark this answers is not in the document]
flixlix: For now I'll add this default matcher

docs/pages/docs/routing/middleware.mdx
flixlix: tried without, is working in development as well as in production, so I'd say lets keep it out

docs/pages/docs/routing/middleware.mdx
flixlix: refactored accordingly

docs/pages/docs/routing/middleware.mdx
```diff
@@ -549,29 +549,113 @@ export const config = {
...
-  await supabase.auth.getSession();
+  const response = handleI18nRouting(request);
 
-  return res;
+  try {
+    const {supabase} = createClient(request);
+    await supabase.auth.getSession();
+    return response;
+  } catch (e) {
+    return NextResponse.next({
```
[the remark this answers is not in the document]
flixlix: adjusted accordingly

docs/pages/docs/routing/middleware.mdx
```diff
@@ -549,29 +549,113 @@ export const config = {
...
+            name,
+            value: '',
+            ...options
+          });
+        }
+      }
+    }
+  );
+
+  return {supabase, response};
```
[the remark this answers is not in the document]
flixlix: just put it all in one `middleware.ts` file, the `createClient` function now returns only the supabase object, since the response isn't needed.

amannn:
Hey, I took the liberty to adapt the section a bit in https://github.com/amannn/next-intl/pull/856/commits/a780a555322ffa10b40a3220c79751c84abf5267. I tried to make it a bit more compact by relying more on the guide from Supabase (e.g. in regard to setting up env params).

Let me know if this looks good to you!

docs/pages/docs/routing/middleware.mdx
```diff
@@ -549,73 +549,43 @@ export const config = {
...
-
-Finally, we're going to create a middleware file that uses the `createClient` function that we created and merges it with the `next-intl` middleware.
+You can do so by following the [setup guide from Supabase](https://supabase.com/docs/guides/auth/server-side/nextjs?router=app) and adapting the middleware as follows:
 
 ```tsx filename="middleware.ts"
 import {type NextRequest} from 'next/server';
 import {createServerClient, type CookieOptions} from '@supabase/ssr';
 import createIntlMiddleware from 'next-intl/middleware';
 
+const handleI18nRouting = createIntlMiddleware({
```
flixlix, line 561: Why put this function outside the middleware, if is not used anywhere else? Just curious

flixlix (commented):
Otherwise lgtm

docs/pages/docs/routing/middleware.mdx
amannn: I'd say it's really fine in either place. Since the creation of the middleware doesn't depend on any request-based information, I figured we can move it outside. But inside is absolutely ok too!

amannn (commented):
Thanks again for your help here! 🙌

flixlix:
Thank you for merging this PR! 🎉

Issue #775, opened 2024-01-11
Make `locales` case insensitive. Allow automatic resolving to the URL with the casing provided in the `locales` property.


codesfromshad:
### Is your feature request related to a problem? Please describe.

When locales are specified according to the [IETF BCP 47 specification](https://en.wikipedia.org/wiki/IETF_language_tag), a mismatch in character casing in a URL request can lead to issues. For instance, if the locales are defined as `[bn-BD, en-UK, ...]` but a request is made using a URL with an incorrect casing, such as `bn-bd`, this discrepancy results in a 404 error.

### Describe the solution you'd like

To address the issue described, the solution should involve implementing a case-insensitive matching mechanism for locale identifiers in the package. This means that when a URL request is made, the package should recognize and correctly handle locales regardless of their character casing. For example, if the locales are defined as `[bn-BD, en-UK, ...]`, the package should treat a request with `bn-bd` as valid and correctly map it to `bn-BD`, also rewrite the URL to reflect the correct case, changing `bn-bd` to `bn-BD`. This approach will prevent 404 errors due to case mismatches and ensure that the package adheres to the flexibility often expected in handling language tags as per the [IETF BCP 47 specification](https://en.wikipedia.org/wiki/IETF_language_tag). In summary, the desired outcome is a more forgiving and robust handling of locale identifiers, accommodating variations in character casing.

### Describe alternatives you've considered

This is my current workaround:
```TypeScript
// middleware.ts

import { getCanonicalLocales } from "@formatjs/intl-getcanonicallocales";
import type { I18nConfig, Locale } from "i18n.config";
import { defaultLocale, locales } from "i18n.config";
import type { NextRequest } from "next/server";
import { NextResponse } from "next/server";
import createMiddleware from "next-intl/middleware";

export default async function middleware(request: NextRequest) {
  const { pathname, search } = request.nextUrl;

  // Get the potential locale from the first path segment of `pathname`
  const potentialLocale = pathname.split("/")[1];

  // Initialize canonical locale as an empty string
  let canonicalLocale = "";

  try {
    // Attempt to get a validated and canonicalized version of the potential locale
    canonicalLocale = getCanonicalLocales(potentialLocale)[0];
  } 
  // Catch and ignore `RangeError` for invalid locales
  catch (error) {}

  // If the canonical locale is valid and different from the potential one,
  // and is included in our list of supported `locales`, perform a redirection
[...]

PR #861, opened 2024-02-15
feat: Add redirects for case mismatches in locale prefixes (e.g. `/EN` → `/en`)

fkapsahili:
I noticed the same in one of my projects and thought I could look into it 🙂 . This PR should implement a case-insensitive matching mechanism for the given locales. This means that the package should treat a request with an invalid case locale, such as `de-at` or `DE`, as valid and map it correctly to `de-AT` and `de` respectively; it should also rewrite the URL to reflect the correct case, for example by rewriting `de-at` to `de-AT`.

I have done a little bit research on this topic and I think case-insensitive matching for locales makes sense to enhance the UX and avoid people mixing up uppercase and lowercase when typing in locale codes. Also, it seems to be a common practice on a lot of popular websites. Maybe it would still make sense to be able to control the behavior via a property in the middleware. What do you think about this?

Also, in the Playwright test I added, I noticed that the matcher `toHaveURL` still expects `/DE` as the URL. In my browser, however, the redirect works as expected from `/DE` to `/de`. Just let me know if I have overlooked something in the implementation, or if I can improve something 🙂.

### Prerequisites

### Content
- Added tests to the `middleware` that verify the `pathLocale` is converted correctly in a case-insensitive manner
- Changed `getNormalizedPathname` to handle the locale in a case-insensitive manner
- Changed `getLocaleFromPathname` to get a validated version of the locale with case-insensitive matching

#### Tests
- Added tests to verify the redirect for invalid cased locales works as intended
- Added a playwright test to the `example-app-router-playground` to make sure the redirect works

Closes https://github.com/amannn/next-intl/issues/775

### DIFF
--- a/examples/example-app-router-playground/tests/main.spec.ts
+++ b/examples/example-app-router-playground/tests/main.spec.ts
@@ -57,7 +57,7 @@ it('redirects to a matched locale at the root for non-default locales', async ({
   page.getByRole('heading', {name: 'Start'});
 });
 
-it('redirects to a matched locale for invalid cased locales', async ({
+it('redirects to a matched locale for an invalid cased non-default locale', async ({
   browser
   }) => {
   const context = await browser.newContext({locale: 'de'});
@@ -66,6 +66,39 @@ it('redirects to a matched locale for invalid cased locales', async ({
   await page.goto('/DE');
   await expect(page).toHaveURL('/de');
   page.getByRole('heading', {name: 'Start'});
+});
+
+it('redirects to a matched locale for an invalid cased non-default locale in a nested path', async ({
+  browser
+  }) => {
+  const context = await browser.newContext({locale: 'de'});
+  const page = await context.newPage();
+
+  await page.goto('/DE/verschachtelt');
+  await expect(page).toHaveURL('/de/verschachtelt');
+  page.getByRole('heading', {name: 'Verschachtelt'});
+});
+
+it('redirects to a matched locale for an invalid cased default locale', async ({
+  browser
+  }) => {
+  const context = await browser.newContext({locale: 'en'});
+  const page = await context.newPage();
+
+  await page.goto('/EN');
+  await expect(page).toHaveURL('/');
+  page.getByRole('heading', {name: 'Home'});
+});
+
+it('redirects to a matched locale for an invalid cased default locale in a nested path', async ({
+  browser
+  }) => {
+  const context = await browser.newContext({locale: 'en'});
+  const page = await context.newPage();
+
+  await page.goto('/EN/nested');
+  await expect(page).toHaveURL('/nested');
+  page.getByRole('heading', {name: 'Nested'});
 });
 
 it('redirects a prefixed pathname for the default locale to the unprefixed version', async ({
--- a/examples/example-app-router-playground/tests/main.spec.ts
+++ b/examples/example-app-router-playground/tests/main.spec.ts
@@ -64,7 +64,7 @@ it('redirects to a matched locale for invalid cased locales', async ({
   const page = await context.newPage();
 
   await page.goto('/DE');
-  await expect(page).toHaveURL('/DE');
+  await expect(page).toHaveURL('/de');
   page.getByRole('heading', {name: 'Start'});
 });
 
--- a/examples/example-app-router-playground/tests/main.spec.ts
+++ b/examples/example-app-router-playground/tests/main.spec.ts
@@ -57,6 +57,17 @@ it('redirects to a matched locale at the root for non-default locales', async ({
   page.getByRole('heading', {name: 'Start'});
 });
 
+it('redirects to a matched locale for invalid cased locales', async ({
+  browser
+  }) => {
+  const context = await browser.newContext({locale: 'de'});
+  const page = await context.newPage();
+
+  await page.goto('/DE');
+  await expect(page).toHaveURL('/DE');
+  page.getByRole('heading', {name: 'Start'});
+});
+
 it('redirects a prefixed pathname for the default locale to the unprefixed version', async ({
   request
 }) => {
--- a/packages/next-intl/src/middleware/utils.tsx
+++ b/packages/next-intl/src/middleware/utils.tsx
@@ -10,14 +10,15 @@ export function getLocaleFromPathname<Locales extends AllLocales>(
   locales: Locales
 ) {
   // Get the potential locale from the pathname
-  const potentialLocale = pathname.split('/')[1];
+  const potentialLocale = pathname.split('/')[1].toLowerCase();
 
-  // Try to get a validated version of the potential locale in a case-insensitive manner
-  return (
-    locales.find(
-      (locale) => locale.toLowerCase() === potentialLocale.toLowerCase()
-    ) ?? ''
+  // Try to get a validated version of the potential locale with case-insensitive matching
+  const matchedLocale = locales.find(
+    (locale) => locale.toLowerCase() === potentialLocale
   );
+
+  // Return matched locale or default to an empty string if not found
+  return matchedLocale ?? '';
 }
 
 export function getInternalTemplate<
--- a/packages/next-intl/src/middleware/utils.tsx
+++ b/packages/next-intl/src/middleware/utils.tsx
@@ -5,8 +5,19 @@ import {
   MiddlewareConfigWithDefaults
 } from './NextIntlMiddlewareConfig';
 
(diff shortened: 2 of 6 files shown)


packages/next-intl/test/middleware/middleware.test.tsx
```diff
@@ -531,6 +537,26 @@ describe('prefix-based routing', () => {
         );
       });
 
+      it('redirects an invalid, upper-cased request for a localized route to the case-sensitive format', () => {
+        middlewareWithPathnames(createMockRequest('/DE-AT', 'de-AT'));
+        expect(MockedNextResponse.rewrite).toHaveBeenCalled();
+        expect(MockedNextResponse.next).not.toHaveBeenCalled();
+        expect(MockedNextResponse.redirect).not.toHaveBeenCalled();
+        expect(MockedNextResponse.rewrite.mock.calls[0][0].toString()).toBe(
+          'http://localhost:3000/de-AT'
```
amannn, line 546: I'm not sure if we're on the same page, but this test doesn't seem to assert for what the description says.

Let's clear up the terminology:
- **Redirect**: A status code in the 3xx range is returned that points the user to another location.
- **Rewrite**: The incoming request is mapped to another internal path.

In the context of `next-intl`, we e.g. have a _redirect_ for the root `/` to the default locale (e.g. `/en`). On the other hand, we use _rewrites_ e.g. for [localized pathnames](https://next-intl-docs.vercel.app/docs/routing/middleware#localizing-pathnames) to handle them via a single internal pathname.

The way I'd imagine this feature and how [I've observed it e.g. on MDN](https://github.com/amannn/next-intl/issues/775#issuecomment-1930415643) is that a redirect is returned when there's a case mismatch.

Due to this, the test should read like this:

```suggestion
        expect(MockedNextResponse.rewrite).not.toHaveBeenCalled();
        expect(MockedNextResponse.next).not.toHaveBeenCalled();
        expect(MockedNextResponse.redirect).toHaveBeenCalled();
        expect(MockedNextResponse.redirect.mock.calls[0][0].toString()).toBe(
          'http://localhost:3000/de-AT'
```

examples/example-app-router-playground/tests/main.spec.ts
```diff
@@ -57,6 +57,17 @@ it('redirects to a matched locale at the root for non-default locales', async ({
...
 });
 
+it('redirects to a matched locale for invalid cased locales', async ({
+  browser
+  }) => {
+  const context = await browser.newContext({locale: 'de'});
+  const page = await context.newPage();
+
+  await page.goto('/DE');
+  await expect(page).toHaveURL('/DE');
```
amannn, line 67: > Also, in the Playwright test I added, I noticed that the matcher toHaveURL still expects /DE as the URL.

See the comment below, that should help to fix this test too.

packages/next-intl/test/middleware/middleware.test.tsx
```diff
@@ -531,6 +537,26 @@ describe('prefix-based routing', () => {
...
+      });
+
+      it('redirects an invalid, lower-cased request for a localized route to the case-sensitive format', () => {
+        middlewareWithPathnames(createMockRequest('/de-at', 'de-AT'));
+        expect(MockedNextResponse.rewrite).toHaveBeenCalled();
+        expect(MockedNextResponse.next).not.toHaveBeenCalled();
+        expect(MockedNextResponse.redirect).not.toHaveBeenCalled();
+        expect(MockedNextResponse.rewrite.mock.calls[0][0].toString()).toBe(
+          'http://localhost:3000/de-AT'
+        );
```
amannn, line 557:
```suggestion
        middlewareWithPathnames(createMockRequest('/de-at', 'de-AT'));
        expect(MockedNextResponse.rewrite).not.toHaveBeenCalled();
        expect(MockedNextResponse.next).not.toHaveBeenCalled();
        expect(MockedNextResponse.redirect).toHaveBeenCalled();
        expect(MockedNextResponse.redirect.mock.calls[0][0].toString()).toBe(
          'http://localhost:3000/de-AT'
        );
```

amannn (commented):
Another great PR, thank you so much @fkapsahili! 🙌

Let me know if my inline comment is helpful, it seems to me like there might have been a slight misunderstanding. 

> Maybe it would still make sense to be able to control the behavior via a property in the middleware. What do you think about this?

Let's make this non-configurable for the time being, I can't think of a good use case where you wouldn't want this and [composing the middleware](https://next-intl-docs.vercel.app/docs/routing/middleware#composing-other-middlewares) generally gives users more options for edge cases than what the next-intl middleware assumes as defaults.

packages/next-intl/test/middleware/middleware.test.tsx
fkapsahili: Ah yes, that makes perfect sense. I also noticed during my research that a redirect is returned if there is a mismatch in the locale - I got something mixed up when writing the tests 😄. I have now adjusted this according to your suggestions and the tests failed as expected, but after I undid the changes in `resolveLocale`, everything works. And as a result, only very few changes are necessary in the middleware 🙏🏼.

fkapsahili:
Let me know if there is anything to improve or if I have overlooked something! 🙂

amannn (commented):
Thanks for the update, this already looks pretty good!

> only very few changes are necessary in the middleware

These are the best PRs, a few new tests and a minimal change!

Would you mind adding these tests too:
 - Redirects for nested paths (e.g. `/about`) for the default and secondary locales
 - Redirects for `localePrefix: 'as-necessary'` for a secondary locale
 - Redirects for nested paths when using non-localized pathnames (default and secondary locales)

Btw. my apologies if the middleware code is a bit hard to follow, I'd really like to clean it up into more grouped code paths at some point. I think for now it's important to have a comprehensive test suite to rely on, this would enable a potential refactor at some point.

> I undid the changes in `resolveLocale`

I really appreciate how careful you are with changes in PRs 🙏

fkapsahili:
Thanks for the review! Also thanks for your tip, I noticed that nested paths were still rewritten for the default locale - I have now fixed this with a small change in `getKnownLocaleFromPathname` so that the redirect works in all cases.
I also added a few more Playwright tests to make sure that all redirects work correctly in the browser. The test cases you mentioned should also all be covered now. I assumed that by `as-necessary` you meant the `as-needed` strategy in the middleware 🙂.

packages/next-intl/src/middleware/utils.tsx
```diff
@@ -86,7 +88,9 @@ export function getKnownLocaleFromPathname<Locales extends AllLocales>(
   locales: Locales
 ): Locales[number] | undefined {
   const pathLocaleCandidate = getLocaleFromPathname(pathname);
-  const pathLocale = locales.includes(pathLocaleCandidate)
+  const pathLocale = locales.find(
+    (locale) => locale.toLowerCase() === pathLocaleCandidate.toLowerCase()
+  )
```
amannn: The value returned from this function is assigned to a variable here:

https://github.com/amannn/next-intl/blob/5cc264ddd769a57ea09440f25c26325c142d0c25/packages/next-intl/src/middleware/middleware.tsx#L137

… and since it doesn't equal the detected locale at 

https://github.com/amannn/next-intl/blob/5cc264ddd769a57ea09440f25c26325c142d0c25/packages/next-intl/src/middleware/middleware.tsx#L219

… a redirect is invoked:

https://github.com/amannn/next-intl/blob/5cc264ddd769a57ea09440f25c26325c142d0c25/packages/next-intl/src/middleware/middleware.tsx#L246

That is correct.

However, the `locale` that is returned from `resolveLocale` here:

https://github.com/amannn/next-intl/blob/5cc264ddd769a57ea09440f25c26325c142d0c25/packages/next-intl/src/middleware/middleware.tsx#L47

… doesn't pick up the prefix since it used to be case senstive.

Since we redirect based on `locale` (and not `pathLocale`), I noticed that that request like `createMockRequest('/EN', 'de')` in the test would redirect to `/de`.

I've added a small fix in https://github.com/amannn/next-intl/pull/861/commits/f05395bfa86c94bda0c03666180b1417db78c585 that uses the locale from the pathname instead in this case—hope this is ok for you!

amannn (approved):
Thanks for adding more tests, this looks really solid to me now! 💯 

I'll just wait for the tests to pass …

> I assumed that by `as-necessary` you meant the `as-needed` strategy in the middleware 🙂.

Yeah, definitely—I keep confusing these two 🤦‍♂️. In retrospect, maybe it should have been called `non-default` 🙂 

Many thanks again for this super clean PR! 👏👏

PR #885, opened 2024-02-25
fix: Improve support for older browsers by switching from `replaceAll` to `replace`

### LINKED ISSUE #884: t.replace(...).replaceAll is not a function
### Description

Hi!

People using slightly older browsers trigger the following error in my Sentry:

`t.replace(...).replaceAll is not a function`

that's because `replaceAll` is kinda new function. 

Would it be possible to change the usage of this in the project? It appears only at two places:

1. https://github.com/amannn/next-intl/blob/cb74c29e9a494691f27445f4b2043c924a8713c0/packages/next-intl/src/shared/utils.tsx#L114 (the one that is relevant to my current error I assume)
2. https://github.com/amannn/next-intl/blob/cb74c29e9a494691f27445f4b2043c924a8713c0/packages/next-intl/src/middleware/utils.tsx#L123 (have not seen error with this one yet but I think it might come if the first place gets changed and this not)

My users are hitting error boundary and cannot use the web at the moment.

## Full Error Stack

```
{
message: t.replace(...).replaceAll is not a function, 
name: TypeError, 
stack: 
TypeError: t.replace(...).replaceAll is not a function
    at a (https://www.vasekvety.sk/_next/static/chunks/296-ae496176be3b700e.js:1:7719)
    at Object.e.matchesPathname (https://www.vasekvety.sk/_next/static/chunks/296-ae496176be3b700e.js:1:8149)
    at https://www.vasekvety.sk/_next/static/chunks/296-ae496176be3b700e.js:1:6444
    at Array.find (<anonymous>)
    at Object.e.getRoute (https://www.vasekvety.sk/_next/static/chunks/296-ae496176be3b700e.js:1:6388)
    at usePathname (https://www.vasekvety.sk/_next/static/chunks/296-ae496176be3b700e.js:1:2092)
// ... the rest is not relevant ...
```

Not needed really

### Reproduction description

Steps to reproduce:
- I don't think this is necessary; both the error and the docs mentioning replaceAll is a new function are descriptive enough 

### Expected behaviour

I expect that the `templateToRegex` function is run without any error at `replaceAll()` line

MichalMoravik:
Fixes #884


amannn (commented):
Thanks for proposing this change! I'd generally be ok with this, as long as bundle size doesn't grow.

Seems like the regex is invalid currently: https://github.com/amannn/next-intl/actions/runs/8041481638/job/21968736617?pr=885

Can you share which browser you saw in your bug report?

MichalMoravik:
@amannn I replied in the issue. Thank you!!

amannn:
Thanks! Yep, as mentioned I'd be ok with this change as long as the bundle size doesn't grow. Can you fix the invalid regex?

MichalMoravik:
Hey @amannn, sorry for getting back so late. I just changed `.replace` to `.replaceAll` again. Can you check if everything is correct and let me know if this would be possible? Thanks!

amannn (approved):
Perfect, thanks a lot @MichalMoravik!

PR #896, opened 2024-02-27
docs: add example of next auth v5

Robin-Ln:


Robin-Ln:
https://github.com/nextauthjs/next-auth/issues/9536

PR #899, opened 2024-02-28
docs: add example of next auth v5 

### LINKED ISSUE #596: Support for NextAuth v5 multiple middleware

Robin-Ln:
closes #596


amannn:
Hey @Robin-Ln, sorry I wasn't able to look into this PR deeper yet. Upgrading to Auth.js v5 is certainly something I'm looking for, but since the stable release hasn't been published yet, this hasn't been a priority so far. There was also an initial attempt at this in https://github.com/amannn/next-intl/pull/598. Ideally we keep as much of the previous example in tact and upgrade to v5 once that is stable. I'll need to have a closer look at this PR and #598 to see about which parts work best once v5 has been released.

Thank you very much for your work here!

Issue #904, opened 2024-03-01
Allow only translated pathnames


amannn:
### Discussed in https://github.com/amannn/next-intl/discussions/902

<div type='discussions-op-text'>

<sup>Originally posted by **mnemosx** March  1, 2024</sup>
Hi!
Using your middeware with pathnames set up like this:
```js
const handleI18nRouting = createIntlMiddleware({
  locales,
  defaultLocale: SupportedLangs.LV,
  localeDetection: false,

  pathnames: {
    '/': '/',
    '/about': {
      lv: '/par-mums',
      ru: '/o-nas',
    },
    '/contacts': {
      lv: '/kontakti',
      ru: '/kontakti',
    },
    '/faq': {
      lv: '/jautajumi',
      ru: '/voprosi',
    },
  },
});
```
works really well when I'm trying to access translated paths like `http://localhost:3000/lv/jautajumi/` 🙌 

There is just one issue -- `http://localhost:3000/lv/faq/` remains accessible too! This is not ideal because there are duplicate pages from the SEO standpoint.

What I'd like to see is that only the translated `lv` and `ru` links are available, while the English versions that are used as directory names in the App router and as keys in code (`about`, `contacts`, `faq`, etc.) return 404. e.g.:

- `/ru/voprosi/` -> renders Russian content
- `/voprosi/` -> adds the default prefix which is lv so path `/lv/voprosi/` renders 404
- `/lv/jautajumi/` -> renders Latvian content
- `/jautajumi/` -> adds the default prefix which is `lv` and renders Latvian content
- `/lv/faq/` -> **renders 404**
- `/ru/faq/` -> **renders 404**
- `/faq/` -> either adds a default prefix which is `lv` and renders 404 or just **renders 404** right away  

Can this be done?

Adding my folder structure for reference in case it helps:
<img width="224" alt="Screenshot 2024-03-01 at 15 06 53" src="https://github.com/amannn/next-intl/assets/16293217/166cfb46-6e01-41f7-919d-0301cca9f240">
</div>

RomainGueffier:
Hi, I also noticed this possible feature, in my projects I currently handle myself unwanted original directory paths with 301 redirections but looks like 404 is better.

I would like to contribute to this awesome project and try submit a PR if nobody already works on it. Disclaimer, I will not ship it fast as it is my first "real" contribution 😃.

Should it be an opt-in feature?

amannn:
[...]

PR #914, opened 2024-03-05
feat: When using localized pathnames, allow access to internal pathnames only if they match an entry from a particular locale—otherwise redirect

RomainGueffier:
Fixes https://github.com/amannn/next-intl/issues/904

**Example:**

```tsx
createIntlMiddleware({
  locales: ['lv', 'ru'],
  defaultLocale: 'lv',
  pathnames: {
    '/': '/',
    '/about': {
      lv: '/par-mums',
      ru: '/o-nas',
    }
  }
});
```

In this case, the only available app routes are `/lv`, `/ru`, `/lv/par-mums` and `/ru/o-nas`.

If a request for `/lv/about` or `/ru/about` is initiated, a 302 redirect to the localized alternative will now be returned.

If `en: '/about'` is added to `pathnames`, then the internal pathname becomes available—but only for this particular locale.

This PR furthermore includes a minor fix for the case where multiple localized pathnames share the same entry, but there are still variations:

```js
'/about': {
  'en-UK': '/about',
  'en-US': '/about',
  'lv': '/par-mums'
}
```

Previously, `/en-US/about` could incorrectly redirect to `/en-UK/about`. Now, this is fixed.

### DIFF
--- a/packages/next-intl/src/middleware/middleware.tsx
+++ b/packages/next-intl/src/middleware/middleware.tsx
@@ -145,12 +145,12 @@ export default function createMiddleware<Locales extends AllLocales>(
 
     let pathname = request.nextUrl.pathname;
     if (configWithDefaults.pathnames) {
-      let resolvedTemplateLocale;
-      [resolvedTemplateLocale = locale, internalTemplateName] =
-        getInternalTemplate(configWithDefaults.pathnames, normalizedPathname);
-      const isDefaultLocale =
-        configWithDefaults.defaultLocale === locale ||
-        domain?.defaultLocale === locale;
+      let resolvedTemplateLocale: Locales[number] | undefined;
+      [resolvedTemplateLocale, internalTemplateName] = getInternalTemplate(
+        configWithDefaults.pathnames,
+        normalizedPathname,
+        locale
+      );
 
       if (internalTemplateName) {
         const pathnameConfig =
@@ -168,48 +168,32 @@ export default function createMiddleware<Locales extends AllLocales>(
             pathLocale
           );
         } else {
-          response = redirect(
-            getPathWithSearch(
-              formatTemplatePathname(
-                normalizedPathname,
-                typeof pathnameConfig === 'string'
-                  ? pathnameConfig
-                  : pathnameConfig[resolvedTemplateLocale],
-                localeTemplate,
-                pathLocale || !isDefaultLocale ? locale : undefined
-              ),
-              request.nextUrl.search
-            )
-          );
-        }
-      } else {
-        const segments = pathname.split('/');
-        if (segments[1] === locale) segments.splice(1, 1);
-        const pathnameWithoutLocale = segments.join('/');
+          let sourceTemplate;
+          if (resolvedTemplateLocale) {
+            // A localized pathname from another locale has matched
+            sourceTemplate =
+              typeof pathnameConfig === 'string'
+                ? pathnameConfig
+                : pathnameConfig[resolvedTemplateLocale];
+          } else {
+            // An internal pathname has matched that
+            // doesn't have a localized pathname
+            sourceTemplate = internalTemplateName;
+          }
 
-        const internalTemplateName = Object.keys(
-          configWithDefaults.pathnames
-        ).find((template) => matchesPathname(template, pathnameWithoutLocale));
-
-        if (internalTemplateName) {
-          const pathnameConfig =
-            configWithDefaults.pathnames[internalTemplateName];
-          const localeTemplate: string =
-            typeof pathnameConfig === 'string'
-              ? pathnameConfig
-              : pathnameConfig[locale];
-          const showLocale =
-            (configWithDefaults.localePrefix === 'as-needed' &&
-              !isDefaultLocale) ||
-            configWithDefaults.localePrefix === 'always';
+          const localePrefix =
+            (hasLocalePrefix || !hasMatchedDefaultLocale) &&
+            configWithDefaults.localePrefix !== 'never'
+              ? locale
+              : undefined;
 
           response = redirect(
             getPathWithSearch(
               formatTemplatePathname(
                 normalizedPathname,
-                internalTemplateName,
+                sourceTemplate,
                 localeTemplate,
-                showLocale ? locale : undefined
+                localePrefix
               ),
               request.nextUrl.search
             )
--- a/packages/next-intl/src/middleware/middleware.tsx
+++ b/packages/next-intl/src/middleware/middleware.tsx
@@ -148,6 +148,9 @@ export default function createMiddleware<Locales extends AllLocales>(
       let resolvedTemplateLocale;
       [resolvedTemplateLocale = locale, internalTemplateName] =
         getInternalTemplate(configWithDefaults.pathnames, normalizedPathname);
+      const isDefaultLocale =
+        configWithDefaults.defaultLocale === locale ||
+        domain?.defaultLocale === locale;
 
       if (internalTemplateName) {
         const pathnameConfig =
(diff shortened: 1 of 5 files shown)


amannn:
Hey @RomainGueffier, 

thank you so much for this PR! 🙏 

I currently have quite a full schedule and need to prioritize a little bit. Just wanted to let you know that I have this PR on my list, I'll provide a review here as soon I find time for it!

Thanks again and have a good weekend! 🙌

packages/next-intl/src/middleware/middleware.tsx
```diff
@@ -183,6 +182,37 @@ export default function createMiddleware<Locales extends AllLocales>(
             )
           );
         }
+      } else {
+        const pathnameWithoutLocale =
+          pathname.split(`/${locale}`)[1] ?? pathname;
```
vordgi, line 187: Be careful with these matchings. This will be valid for example for the `/enterprise` page when locale is `en`.

May be would be better to do it like this:

```js
const pathnameWithoutLocale = pathname.split(new RegExp(`\/${lang}(?:\/|$)`))[1] ?? pathname;
```
or
```js
const matchedPath = pathname.match(new RegExp(`\/${lang}(?<path>\/.*|$)`));
const pathnameWithoutLocale = matchedPath?.groups.path ?? pathname
```
or
```js
const segments = pathname.split('/');
if (segments[1] === locale) segments.splice(1, 1);
const pathnameWithoutLocale = segments.join('/');
```

packages/next-intl/test/middleware/middleware.test.tsx
```diff
@@ -737,6 +749,48 @@ describe('prefix-based routing', () => {
           'http://localhost:3000/external?hello'
         );
       });
+
+      describe('redirects an internal route with localization to its localized pathname but removes the locale prefix while keeping search params', () => {
```
vordgi, line 753: Add tests on paths starting from the current locale, from the default locale, containing two locales and two default locales. next.js had a lot of problems with this in their i18n routing - don't repeat their mistakes!)

packages/next-intl/src/middleware/middleware.tsx
RomainGueffier: @vordgi I don't understand, what is this `/enterprise` page? It's nowhere in tests nor exemples. This line of code is not about matching but rather removing locale from the url.

The matching is done below, by checking if the actual path exists in next intl paths mapping.

In any case, last suggestion is good !

packages/next-intl/test/middleware/middleware.test.tsx
RomainGueffier: What am I supposed to do in such scenarios ? `/en/en/internal` => `/en/external` ??
I think this is outside of the scope of this feature, or not sure to understand...

packages/next-intl/src/middleware/middleware.tsx
vordgi: Just imagine that the locale is `en` and we got it from the domain (e.g. `en.example.com`), not from the path. The current path is `/enterprise` (`en.example.com/enterprise`).

Your logic (```pathname.split(`/${locale}`)[1] ?? pathname```) will remove `/en` from the path and keep `terprise` (`'/enterprise'.split('/en')[1] ?? '/enterprise'`)

packages/next-intl/test/middleware/middleware.test.tsx
vordgi: Next.js, when you go to the` /en/en/internal` page, will show you the `/en/internal` page as if everything is fine. But formally it is non-existent path and should response with 404 status

It was also funny other time that next.js from this path redirected to /en/internal and showed this page, and after reloading it redirected again - to /internal and again showed as if everything was okay.

vordgi: _Perhaps this is a comment not so much for you as for amannn_

packages/next-intl/src/middleware/middleware.tsx
RomainGueffier: 😮 I see now... thanks for explanation and suggestion!

amannn:
Hey @RomainGueffier,

I've finally found time to look into this, thank you for your patience! While reviewing the code, I've cleaned up some things along the way and added a few more tests, but I think everything should be fine now.

Let's see if CI passes …

packages/next-intl/test/middleware/middleware.test.tsx
amannn: @vordgi We had that topic with pathnames that overlap with a locale just recently! 😄 See https://github.com/amannn/next-intl/pull/698.

As far as I know that has been entirely addressed and should no longer be an issue.

amannn:
Yep, seems like all tests pass on CI! 🙌 

@RomainGueffier would you mind having another look from your side at the code to see if everything matches your understanding?

packages/next-intl/test/middleware/middleware.test.tsx
vordgi: A-ha-ha, wow, it means you solved this problem much faster than next.js team 😁

RomainGueffier:
Hello @amannn thanks for the feedback 😃 and little refactoring of my tests ! Looks good.
I would like to address these questions I had initially ⬇️ 

> 
> To discuss:
> 
> * [ ]  why not using 301 redirections or 404
> * [ ]  how to test internal pathname not inside Next Intl config, not found, etc. to avoid breaking Nextjs default features
> * [ ]  code quality and refactoring to make code more readdable
> * [ ]  add more tests (catch-all paths, etc.)
> * [ ]  documentation
> * [ ]  update exemples with internal pathnames

I feel like the code is a bit noisy, maybe I could clarify it in dedicated functions ?
What about using 301 instead of 302 redirections ?
Also I did not run a real next js application to try out modifications and internal paths not covered by next intl config, such as sitemap, robots, or other ones. Should I add more paths into exemples?

amannn:
> I feel like my code is a bit noisy, maybe I could clarify it in dedicated functions ?

I think for the time being it's ok. I'd eventually like to do a cleanup pass on the whole middleware logic to split it into chunks, but I'm waiting a bit with this currently since it's a bigger task and I think there are currently other priorities.

> What about using 301 instead of 302 redirections ?

I think 301 is a bit more risky since it can be cached. What we're handling here should be an edge case and not a URL that is available in a sitemap for example. I think sticking with 302 can be helpful in case a user adds a localized route that matches the internal one and this route suddenly becomes valid.

> Also I did not run a real next js application to try out modifications

I ran the example in the repo and the result looked good to me!

> internal paths not covered by next intl config, such as sitemap, robots, or other ones. Should I add more paths into exemples?

Maybe another test where a request is invoked that matches neither an internal nor an external pathname could be a good idea, yes! I think it should work as expected, but better safe than sorry :). So definitely add one if you feel like it!


---

Let me know what you think!

RomainGueffier:
Ok sounds good!

> Maybe another test where a request is invoked that matches neither an internal nor an external pathname could be a good idea, yes! I think it should work as expected, but better safe than sorry :). So definitely add one if you feel like it!

I tried to write a test, but it always rewrites or redirects to locale:
```ts
it('metadata paths should not redirect to a locale', () => {
      middlewareWithPathnames(
        createMockRequest('/sitemap.xml') // seems to consider this path as a internal route.
      );
      expect(MockedNextResponse.next).not.toHaveBeenCalled();
      expect(MockedNextResponse.rewrite).not.toHaveBeenCalled();
      expect(MockedNextResponse.redirect).not.toHaveBeenCalled();
      expect(MockedNextResponse.[??].toBe(
        'http://localhost:3000/sitemap.xml'
      );
});
```

The mock function does not seems to allow to configure an i18n layout further down the tree or support the next js metadata paths such as `sitemap.xml` or `robots.txt`. Do you have an idea? I'm ready for merge if this is complicated and not a critical test.

amannn:
@RomainGueffier A rewrite being triggered in this case is correct. But for the `/sitemap.xml` request the middleware shouldn't run in the first place, based on [the matcher config](https://next-intl-docs.vercel.app/docs/routing/middleware#matcher-config).

I think this is good to go then!

amannn (approved):
Many thanks for this PR @RomainGueffier! 👏👏

Issue #444, opened 2023-08-07
Improve initial render of `Link` for `localePrefix: 'as-necessary'` and make `getPathname` more useful


amannn:
### Is your feature request related to a problem? Please describe.

The `localePrefix: 'as-necessary'` strategy currently has these drawbacks:
1. The initial render of `Link` in RSC always includes a prefix, which we potentially remove during hydration if we detect that no locale prefix should be used.
2. `getPathname` is somewhat hard to use, because users have to specify their own logic to check if a prefix should be added.

### Describe the solution you'd like

1. The navigation APIs should accept `localePrefix`, `defaultLocale` and `domains`, all being expected to be provided in case you use them. We should move all routing config to a shared type that can be shared with the middleware. **Update**: Done in https://github.com/amannn/next-intl/pull/1299
2. `Link` should consider the new options for the initial render (see https://github.com/amannn/next-intl/issues/444#issuecomment-1832223775). If `Link` receives `locales`, we should be able to use `usePathname` to see if the current pathname is prefixed and use this information to get the initial SSR render right.
3. `getPathname` should consider the options too and return a final pathname (with/without a locale prefix, with/without domain). Furthermore, `getPathname` should be supported for shared pathnames too.
4. The remaining navigation APIs should also be checked for which improvements we could make based on the new configuration.
5. Navigation APIs could accept a `domain` if a user wants to link/redirect to another domain.

This will require a major version to ensure users supply the new options (related to https://github.com/amannn/next-intl/issues/779).

In regard to (2) there's an edge case for `localePrefix: 'as-necessary'` and `domains`. The domain is only available at request time, therefore we have to either a) support SSG and potentially [patch the `href` of `Link` on the client side](https://github.com/amannn/next-intl/issues/444#issuecomment-1832223775) or b) require dynamic rendering and return a final URL. This is TBD. Note that `getPathname` only has "one shot" to return a final URL, which would certainly require reading from headers. An idea could be to require dynamic rendering in case `domains` are configured, but to point users towards the alternative of building a separate app per domain if SSG is important.

Make sure to remove the note in the `localePrefix: 'as-needed'` docs once this is implemented.

### Describe alternatives you've considered

The problems outlined above are mostly related to `localePrefix: 'as-necessary'`.

You can avoid these by either:
1. Using `localePrefix: 'always'`
2. Using `localePrefix: 'never'` and supplying this to the navigation factory function (already supported)

amannn:
[...]

PR #926, opened 2024-03-07
fix: Handle changing href for `<Link />` correctly when using `localePrefix: 'never'`

### LINKED ISSUE #918: Link `href` doesn't update when search params change when `localePrefix` is set to `never`.
### Description

If a `Link` is present in a client component and the url depends on the existing `searchParams`, the url is not updated when the `searchParams` change. For some reason this happens when `createSharedPathnamesNavigation` is invoked with `localePrefix: "never"`, default value seems to work fine.

https://codesandbox.io/p/github/jorgemoya/next-intl-bug-repro-app-router/main?import=true

### Reproduction description

Steps to reproduce:
1. Open reproduction
2. Click on `Foo` link
3. Hover over `Broken Link` and notice it doesn't include `foo=bar` search params when it should.

Alternate:
1. Open reproduction
2. Click on `Foo` link
3. Refresh page
4. Click on `Home` link
5. Hover over `Broken Link` and notice it does include `foo=bar` search params when it shouldn't.

### Expected behaviour

The expected behavior would be for the `Broken Link` link the match the search params present, even after they update.

amannn:
Fixes #918

### DIFF
--- a/packages/next-intl/src/navigation/shared/BaseLink.tsx
+++ b/packages/next-intl/src/navigation/shared/BaseLink.tsx
@@ -27,8 +27,8 @@ function BaseLink(
   // `useParams` can be called, but the return type is `null`.
   const pathname = usePathname() as ReturnType<typeof usePathname> | null;
 
-  const defaultLocale = useLocale();
-  const isChangingLocale = locale !== defaultLocale;
+  const curLocale = useLocale();
+  const isChangingLocale = locale !== curLocale;
 
   const [localizedHref, setLocalizedHref] = useState<typeof href>(() =>
     isLocalHref(href) && (localePrefix !== 'never' || isChangingLocale)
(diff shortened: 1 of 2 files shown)


packages/next-intl/src/navigation/shared/BaseLink.tsx
```diff
@@ -52,12 +52,12 @@ function BaseLink(
...
 
   useEffect(() => {
-    if (!pathname || localePrefix === 'never') return;
+    if (!pathname) return;
 
     setLocalizedHref(
       localizeHref(href, locale, defaultLocale, pathname ?? undefined)
     );
-  }, [defaultLocale, href, locale, localePrefix, pathname]);
+  }, [defaultLocale, href, locale, pathname]);
```
amannn, line 60: Why do we need this effect in the first place? Is this outdated and can we remove it?

amannn: Added some notes to https://github.com/amannn/next-intl/issues/444#issue-1839328048, let's continue there …

Issue #913, opened 2024-03-05
Automatically prefer localized pathnames that are more specific


Hdanzo:
### Description

When using localized pathnames and trying to access a page (/products/add) that is in the same folder as a dynamic page (/products/[productId]), the user is redirected to the dynamic page with the name of the static page as the param (productId = "add"). 

This however is fixable by ordering "/products/add" before "/products/[productId]" in the pathnames object.

If this behavior is the expected one and not something that's planned to be changed, I suggest this behavior to be stated in the documentation.

### Mandatory reproduction URL

https://codesandbox.io/p/devbox/xenodochial-hodgkin-3wj7ll?file=%2Fapp%2F%5Blocale%5D%2Fpage.tsx%3A94%2C11

### Reproduction description

Steps to reproduce:
1. Open reproduction.
2. Change locale to "es" in the URL.
3. Click on "Agregar" Link.
4. Get redirected to "/productos/[productId]" instead of "/productos/agregar"

### Expected behaviour

1. Open reproduction.
2. Change locale to "es" in the URL.
3. Click on "Agregar" Link.
4. Get redirected to "/productos/agregar" instead of "/productos/[productId]"

amannn:
Thanks for this report, that's a great point!

I wonder if `next-intl` could automatically do the right thing here, Next.js is certainly capable of that.

This would need some further research on how routes can be ordered in a meaningful way. If I'm not mistaken, this is how Next.js does it: [`sorted-routes.ts`](https://github.com/vercel/next.js/blob/415cd74b9a220b6f50da64da68c13043e9b02995/packages/next/src/shared/lib/router/utils/sorted-routes.ts#L16). There's another related note in [`normalize-catchall-routes.ts`](https://github.com/vercel/next.js/blob/415cd74b9a220b6f50da64da68c13043e9b02995/packages/next/src/build/normalize-catchall-routes.ts#L20).

And this is where we need to apply it in `next-intl`:

https://github.com/amannn/next-intl/blob/419a9cea2817949b65b1af9724927b95d7d3bf01/packages/next-intl/src/middleware/utils.tsx#L21

I'll add a note about this to the docs for the time being but would be open to reviewing a PR here in case someone is interested in diving deeper into this.

amannn:
Absolutely if you're interested, always a pleasure! 🙌

PR #983, opened 2024-04-06
feat: Automatically prefer localized pathnames that are more specific

fkapsahili:
This PR fixes the current behavior of the internal route handling of the `next-intl` middleware that does not prioritize static route paths over dynamic routes. This fix makes sure that static routes (e.g. `/products/new`) are preferred over dynamic routes such as `/products/[slug]`.

To make sure the current behavior does not lead to any unexpected issues, I thought it would be good if `next-intl` would prioritize the specified `pathnames` in the following order:
1. static pathnames, e.g. `/products/new`
2. dynamic routes (non-catch-all), e.g. `/products/[slug]`
3. catch-all routes, e.g. `/products/[...slug]` or optional catch-all routes, e.g. `/products/[[...slug]]`

### Prerequisites

### Content
- Added `getSortedPathnames` to handle the sorting of the specified pathnames in the `middleware`
- Updated the `docs` to make sure the changes are reflected in the documentation

#### Tests
- Added tests to the `middleware` that verify that static routes are prioritized over dynamic routes and catch-all routes
- Added tests to the middleware `utils` to make sure that `getSortedPathnames` handles the route prioritization as intended
- Added playwright tests to the `example-app-router-playground` to make sure the route prioritization works as intended 

Closes https://github.com/amannn/next-intl/issues/913

### DIFF
--- a/packages/next-intl/src/middleware/utils.tsx
+++ b/packages/next-intl/src/middleware/utils.tsx
@@ -67,7 +67,7 @@ export function comparePathnamePairs<Locales extends AllLocales>(
   }
 
   // Both pathnames are completely static
-  return 0
+  return 0;
 }
 
 export function getSortedPathnames<Locales extends AllLocales>(
--- a/packages/next-intl/src/middleware/utils.tsx
+++ b/packages/next-intl/src/middleware/utils.tsx
@@ -89,6 +89,7 @@ export function getInternalTemplate<
 ): [Locales[number] | undefined, keyof Pathnames | undefined] {
   // Sort pathnames by specificity
   const sortedPathnames = getSortedPathnames(pathnames);
+
   // Try to find a localized pathname that matches
   for (const [
     internalPathname,
--- a/packages/next-intl/src/middleware/utils.tsx
+++ b/packages/next-intl/src/middleware/utils.tsx
@@ -16,10 +16,6 @@ function getExternalPath<Locales extends AllLocales>(
   return typeof path === 'string' ? path : firstLocalePath;
 }
 
-function isCatchAllRoute(pathname: string) {
-  return isOptionalCatchAll(pathname) || isCatchAll(pathname);
-}
-
 function isOptionalCatchAll(pathname: string) {
   return pathname.includes('[[...');
 }
@@ -41,30 +37,37 @@ export function comparePathnamePairs<Locales extends AllLocales>(
   a: PathnamePair<Locales>,
   b: PathnamePair<Locales>
 ): number {
-  const pathA = getExternalPath(a[1]);
-  const pathB = getExternalPath(b[1]);
+  const pathA = getExternalPath(a[1]).split('/').filter(Boolean);
+  const pathB = getExternalPath(b[1]).split('/').filter(Boolean);
 
-  const dynamicA = isDynamicRoute(pathA);
-  const dynamicB = isDynamicRoute(pathB);
-  const catchAllA = isCatchAllRoute(pathA);
-  const catchAllB = isCatchAllRoute(pathB);
+  const maxLength = Math.max(pathA.length, pathB.length);
+  for (let i = 0; i < maxLength; i++) {
+    const segmentA = pathA[i];
+    const segmentB = pathB[i];
 
-  // Prioritize static routes over dynamic and catch-all routes
-  if (!dynamicA && dynamicB) return -1;
-  if (dynamicA && !dynamicB) return 1;
+    // If one of the paths ends, prioritize the shorter path
+    if (!segmentA && segmentB) return -1;
+    if (segmentA && !segmentB) return 1;
 
-  // Both paths are either dynamic or static, prioritize non-catch-all over catch-all
-  if (catchAllA && !catchAllB) return 1;
-  if (!catchAllA && catchAllB) return -1;
+    // Prioritize static segments over dynamic segments
+    if (!isDynamicRoute(segmentA) && isDynamicRoute(segmentB)) return -1;
+    if (isDynamicRoute(segmentA) && !isDynamicRoute(segmentB)) return 1;
 
-  // If both are the same type (static, dynamic non-catch-all, or catch-all), prioritize by depth
-  const depthA = pathA.split('/').length;
-  const depthB = pathB.split('/').length;
+    // Prioritize non-catch-all segments over catch-all segments
+    if (!isCatchAll(segmentA) && isCatchAll(segmentB)) return -1;
+    if (isCatchAll(segmentA) && !isCatchAll(segmentB)) return 1;
+
+    // Prioritize non-optional catch-all segments over optional catch-all segments
+    if (!isOptionalCatchAll(segmentA) && isOptionalCatchAll(segmentB)) {
+      return -1;
+    }
+    if (isOptionalCatchAll(segmentA) && !isOptionalCatchAll(segmentB)) return 1;
+
+    if (segmentA === segmentB) continue;
+  }
 
   // Deeper (more specific) paths first
-  if (depthA !== depthB) return depthB - depthA;
-
-  return 0;
+  return pathA.length - pathB.length;
 }
 
 export function getSortedPathnames<Locales extends AllLocales>(
--- a/packages/next-intl/src/middleware/utils.tsx
+++ b/packages/next-intl/src/middleware/utils.tsx
@@ -9,13 +9,6 @@ export function getFirstPathnameSegment(pathname: string) {
   return pathname.split('/')[1];
 }
 
-function getExternalPath<Locales extends AllLocales>(
-  path: string | {[Key in Locales[number]]: string}
-): string {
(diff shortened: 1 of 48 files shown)


amannn (commented):
Many thanks for looking into this and another great PR @fkapsahili! 👏👏

On a first glance, this looks like just the right implementation. I have a somewhat busy week ahead, but I'll likely be able to take a thorough look on Friday—hope that works for you!

I'll be in touch! 🙌

fkapsahili:
> Many thanks for looking into this and another great PR @fkapsahili! 👏👏
> 
> On a first glance, this looks like just the right implementation. I have a somewhat busy week ahead, but I'll likely be able to take a thorough look on Friday—hope that works for you!
> 
> I'll be in touch! 🙌

Thanks for the feedback! Yes of course, that works well for me 🙂. The tests are all working now, so I'm setting the PR to ready for review 😄.

packages/next-intl/src/middleware/utils.tsx
```diff
@@ -9,6 +9,71 @@ export function getFirstPathnameSegment(pathname: string) {
...
+  a: PathnamePair<Locales>,
+  b: PathnamePair<Locales>
+): number {
+  const pathA = getExternalPath(a[1]);
+  const pathB = getExternalPath(b[1]);
+
+  const dynamicA = isDynamicRoute(pathA);
+  const dynamicB = isDynamicRoute(pathB);
+  const catchAllA = isCatchAllRoute(pathA);
+  const catchAllB = isCatchAllRoute(pathB);
```
amannn, line 50: Diving a bit deeper into this, I'm not sure these flags are sufficient to be honest.

E.g. consider this case:
```
/articles/[category]/[articleSlug]
/articles/[category]/just-in
```

In this case, the routes would be considered equal in regard to priority, and probably the natural ordering would be preserved. However, the second route is more specific and should match.

The number of dynamic segments could be another indicator, but I think this won't solve the problem cleanly as e.g. `/foo/[d1]` and `/foo/bar/[d1]` have the same number of segments.

It seems like to be able to tell which route is really more specific, we have to consider routes in a hierarchical way, e.g. find the most specific route in `/articles`, then try to find the most specific route in `./[category]`, etc.

Do you by chance know if Next.js handles these cases correctly? Maybe we can take some inspiration from them.

What do you think?

packages/next-intl/test/middleware/utils.test.tsx
```diff
@@ -162,3 +163,85 @@ describe('getInternalTemplate', () => {
     ]);
   });
 });
+
+describe('getSortedPathnames', () => {
```
amannn, line 167: Good idea to add some unit tests for this function!

packages/next-intl/src/middleware/utils.tsx
amannn: Or maybe the number of dynamic segments could in fact suffice? 🤔 

In the example `/foo/[d1]` and `/foo/bar/[d1]`, the routes have the same priority, but only one of them matches eventually—so maybe this isn't an issue.

Do you think you could a few more tests for edge cases so we can gain some clarity about those?

fkapsahili: Ah yes, thanks for the hint! You were absolutely right, my solution didn't take into account the case where we have nested dynamic routes, e.g: `/articles/[category]/[slug]` and a static route in the same place 😅.

I'm not very familiar with the Next.js repo, but at first glance I believe the code you referenced in the corresponding issue is actually where Next does the route sorting:
https://github.com/vercel/next.js/blob/415cd74b9a220b6f50da64da68c13043e9b02995/packages/next/src/shared/lib/router/utils/sorted-routes.ts

When looking at Next's code I realized that it is necessary that the individual route segments are compared with each other and have now implemented this additionally so that the failed tests that I added at the beginning of this update now pass.

If there are any other edge cases you can think of, I'm happy to add more tests! 🙂

packages/next-intl/src/middleware/utils.tsx
```diff
@@ -19,10 +87,13 @@ export function getInternalTemplate<
   pathname: string,
   locale: Locales[number]
 ): [Locales[number] | undefined, keyof Pathnames | undefined] {
+  // Sort pathnames by specificity
+  const sortedPathnames = getSortedPathnames(pathnames);
```
amannn, line 91: Would suggest to add a line break before the next comment just to bikeshed a bit about code aesthetics 😄 

```suggestion
  const sortedPathnames = getSortedPathnames(pathnames);

```

packages/next-intl/test/middleware/middleware.test.tsx
```diff
@@ -642,6 +672,54 @@ describe('prefix-based routing', () => {
         );
       });
 
+      it('priotizes static routes over dynamic and catch-all routes for the non-default locale', () => {
```
amannn, line 675:
```suggestion
      it('prioritizes static routes over dynamic and catch-all routes for the non-default locale', () => {
```

packages/next-intl/test/middleware/middleware.test.tsx
```diff
@@ -642,6 +672,54 @@ describe('prefix-based routing', () => {
...
+        expect(MockedNextResponse.rewrite).toHaveBeenCalledTimes(2);
+        expect(MockedNextResponse.rewrite.mock.calls[0][0].toString()).toBe(
+          'http://localhost:3000/en/products/add'
+        );
+        expect(MockedNextResponse.rewrite.mock.calls[1][0].toString()).toBe(
+          'http://localhost:3000/en/categories/new'
+        );
+      });
+
+      it('priotizes static routes over dynamic and catch-all routes for the non-default locale', () => {
```
amannn, line 689:
```suggestion
      it('prioritizes static routes over dynamic and catch-all routes for the non-default locale', () => {
```

packages/next-intl/test/middleware/middleware.test.tsx
```diff
@@ -642,6 +672,54 @@ describe('prefix-based routing', () => {
...
+        expect(MockedNextResponse.rewrite).toHaveBeenCalledTimes(2);
+        expect(MockedNextResponse.rewrite.mock.calls[0][0].toString()).toBe(
+          'http://localhost:3000/de/products/add'
+        );
+        expect(MockedNextResponse.rewrite.mock.calls[1][0].toString()).toBe(
+          'http://localhost:3000/de/categories/new'
+        );
+      });
+
+      it('priotizes more specific, static routes over dynamic routes for the non-default locale', () => {
```
amannn, line 705:
```suggestion
      it('prioritizes more specific, static routes over dynamic routes for the non-default locale', () => {
```

packages/next-intl/test/middleware/middleware.test.tsx
```diff
@@ -1100,10 +1198,76 @@ describe('prefix-based routing', () => {
...
+        expect(MockedNextResponse.rewrite).toHaveBeenCalledTimes(2);
+        expect(MockedNextResponse.rewrite.mock.calls[0][0].toString()).toBe(
+          'http://localhost:3000/en/products/add'
+        );
+        expect(MockedNextResponse.rewrite.mock.calls[1][0].toString()).toBe(
+          'http://localhost:3000/en/categories/new'
+        );
+      });
+
+      it('priotizes more specific, static routes over dynamic routes for the default locale', () => {
```
amannn, line 1219:
```suggestion
      it('prioritizes more specific, static routes over dynamic routes for the default locale', () => {
```

amannn (commented):
Great refactor, this looks really solid to and I think this is the right approach 👏.

I've left some inline comments about minor nits, but I also think there might be a chance to simplify this function slightly.

Let me know what you think! 🙌

packages/next-intl/src/middleware/utils.tsx
```diff
@@ -9,6 +9,74 @@ export function getFirstPathnameSegment(pathname: string) {
...
+  string,
+  string | {[Key in Locales[number]]: string}
+];
+
+export function comparePathnamePairs<Locales extends AllLocales>(
+  a: PathnamePair<Locales>,
+  b: PathnamePair<Locales>
+): number {
+  const pathA = getExternalPath(a[1]).split('/').filter(Boolean);
+  const pathB = getExternalPath(b[1]).split('/').filter(Boolean);
```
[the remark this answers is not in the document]
fkapsahili: Ah yes, that makes sense! I think the `getExternalPath` comes from a first attempt where I tried to separate everything very strictly. Just taking the internal route name is a good idea! I'll try to clean that up a bit so we have an easier to read test suite and a more minimalistic approach in the utils 👍🏼.

fkapsahili:
@amannn I have now simplified the sorting function of `getSortedPathnames` so that only the internal pathnames are passed as arguments, which makes the tests look a bit easier. Just let me know if there is anything I can improve!

amannn (approved):
Looks fantastic, a really cool quality-of-life improvement! 👏 

I'll just add one commit with a minor renaming nit I spotted during the final review.

PR #999, opened 2024-04-17
fix: Correctly detect base path at the app root `/` when using a locale prefix strategy other than `always`. This ensures the locale cookie is set correctly.

### LINKED ISSUE #997: Locale change issue with basePath and localePrefix = 'as-needed'
### Description

Locale simply doesn't change, seems to be related with the incorrectly applied cookies.
Unfortunately I can't reproduce it in codesandbox, only when running localy. Tried in Chrome and Firefox, shouldn't be exclusive to my setup.
Another way to reproduce it to basically add `basePath: '/s',` to `next.config.js` and `localePrefix = 'as-needed'` to `src/config.ts` in your [examples/example-app-router/](https://github.com/amannn/next-intl/tree/main/examples/example-app-router/).

https://github.com/amannn/next-intl/assets/34808650/be7d3de8-9def-4aba-8280-6484dbde66fa

### Verifications

- [X] I've verified that the problem I'm experiencing isn't covered in [the docs](https://next-intl-docs.vercel.app/).
- [X] I've searched for similar, existing issues on GitHub and [Stack Overflow](https://stackoverflow.com/search?q=next-intl).
- [X] I've compared my app to [a working example](https://next-intl-docs.vercel.app/examples) to look for differences.

https://codesandbox.io/p/github/SpasiboKojima/next-intl-bug-repro-app-router/main

### Reproduction description

Steps to reproduce:
1. Open reproduction
2. Change locale to German, see it set another cookie on `/` path
3. Try to change locale back to English

### Expected behaviour

Expected it to change locale and probably to not set another cookie

amannn:
Fixes #997

### DIFF
--- a/packages/next-intl/package.json
+++ b/packages/next-intl/package.json
@@ -122,11 +122,11 @@
     },
     {
       "path": "dist/production/navigation.react-client.js",
-      "limit": "2.94 KB"
+      "limit": "2.96 KB"
     },
     {
       "path": "dist/production/navigation.react-server.js",
-      "limit": "3.03 KB"
+      "limit": "3.06 KB"
     },
     {
       "path": "dist/production/server.react-client.js",
--- a/packages/next-intl/src/navigation/shared/utils.tsx
+++ b/packages/next-intl/src/navigation/shared/utils.tsx
@@ -180,6 +180,13 @@ export function getRoute<Locales extends AllLocales>({
   return template as keyof Pathnames<Locales>;
 }
 
-export function getBasePath(pathname: string) {
-  return window.location.pathname.replace(pathname, '');
+export function getBasePath(
+  pathname: string,
+  windowPathname = window.location.pathname
+) {
+  if (pathname === '/') {
+    return windowPathname;
+  } else {
+    return windowPathname.replace(pathname, '');
+  }
 }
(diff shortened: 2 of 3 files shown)


130db:
Interesting!
Can not upgrade to 3.11.2 or higher after this fix.

Here is my autogenerated config.
Screenshot 2024-07-18 at 20 06 01

const locales is not casted with "as const" because line 9 will fail in i18n.ts

Screenshot 2024-07-18 at 20 08 16

akhlivniukallane:
We experience the same problem with the same config parameters as above.
cc @amannn any idea? It looks like some basic misconfiguration, but where..

130db:
Yes! I fixed it.

Tried different approach by following Mohammad's tutorial
https://phrase.com/blog/posts/next-js-app-router-localization-next-intl/

Did POC on fresh NextJS install and then tried to replicate on my website.
Guess what, it failed :D with

**TypeError [ERR_INVALID_STATE]: Invalid state: ReadableStream is already closed**

The problem is:
When you use async page and call **await getTranslations** it fails.

Solution
Make page non-async and replace **getTranslations** with **useTranslations**
Move every other await to coresponding component

PR #1000, opened 2024-04-17
fix: When using domain-based routing, use `defaultLocale` of a domain instead of the top-level one in case no other locale matches better on the domain

### LINKED ISSUE #998: defaultLocale of domain ignored if defaultLocale set at top-level
### Description

For domain-based routing, I think `defaultLocale` at the top-level is overriding the defaultLocale at the domain level, which I'm convinced is the wrong way around based on the documentation.

I've made a repo (a fork of next-intl) to demonstrate this using the `example-app-router` example.  By the way thanks for your work on next-intl.  It's brilliant!

### Verifications

- [X] I've verified that the problem I'm experiencing isn't covered in [the docs](https://next-intl-docs.vercel.app/).
- [X] I've searched for similar, existing issues on GitHub and [Stack Overflow](https://stackoverflow.com/search?q=next-intl).
- [X] I've compared my app to [a working example](https://next-intl-docs.vercel.app/examples) to look for differences.

https://github.com/cleverlight/next-intl/blob/bugreport-defaultLocale-undefined/examples/example-app-router/src/middleware.ts

### Reproduction description

Steps to reproduce:
1. Clone reproduction URL (repo), branch `bugreport-defaultLocale-undefined`
2. `curl -IL localhost:3000`
3. Redirect goes to /en, when it should go to /de

These steps can be further extended to demonstrate the correct behaviour if defaultLocale is not set at the top-level:
1. Comment line 5 in `middleware.ts` (`// defaultLocale: 'en',`)
2. `curl -IL localhost:3000`
3. Redirect goes to /de as expected

I suspect the problem is inheritance in the structure generally.  You'll note in my example that I had to define `locales` in each domain.  These too should be inheritted from the top-level, but if they're not explicitly set in each domain, an `undefined` redirect loop occurs:
1. Comment line 13 in `middleware.ts` (`// locales`)
2. `curl -IL localhost:3000`
3. Redirect goes to /undefined/undefined/undefined/undefined/...

### Expected behaviour

The [documentation](https://next-intl-docs.vercel.app/docs/routing/middleware#domain-based-routing) states:
```
2. If the host of the request is configured in domains, the defaultLocale of the domain is used
```

I would expect top-level `defaultLocale` or `locales` to be used if they're not defined in the `domains: []` entry.

amannn:
Fixes #998 

Note that the `defaultLocale` of a domain is used if no other locale matches better. However,  if a domain supports multiple locales, the best-matching one will be selected based on the `accept-language` header. If you want to always use the `defaultLocale` in case no prefix is provided, then you can turn off `localeDetection`.

### DIFF
--- a/packages/next-intl/test/middleware/middleware.test.tsx
+++ b/packages/next-intl/test/middleware/middleware.test.tsx
@@ -41,7 +41,7 @@ function createMockRequest(
   customHeaders?: HeadersInit
 ) {
   const headers = new Headers({
-    'accept-language': `${acceptLanguageLocale};q=0.9,en;q=0.8`,
+    'accept-language': `${acceptLanguageLocale};q=0.9`,
     host: new URL(host).host,
     ...(localeCookieValue && {
       cookie: `${COOKIE_LOCALE_NAME}=${localeCookieValue}`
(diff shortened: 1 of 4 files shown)


packages/next-intl/test/middleware/middleware.test.tsx
```diff
@@ -1765,6 +1765,23 @@ describe('domain-based routing', () => {
...
+            defaultLocale: 'fr',
+            domain: 'ca.example.com'
+          }
+        ]
+      });
+      m(createMockRequest('/', 'de', 'http://ca.example.com'));
+      expect(MockedNextResponse.redirect.mock.calls[0][0].toString()).toBe(
+        'http://ca.example.com/fr'
+      );
+    });
```
amannn, line 1783: Note that the `defaultLocale` of a domain is used if no other locale matches better. However, note that if a domain supports multiple locales, the best-matching one will be selected based on the `accept-language` header.

Issue #653, opened 2023-11-22
Customization of prefixes / multi-tenancy


amannn:
### Is your feature request related to a problem? Please describe.

Currently, `next-intl` supports [prefix-based routing](https://next-intl-docs.vercel.app/docs/routing/middleware#prefix-based-routing) (e.g. `/en`), [domain-based routing](https://next-intl-docs.vercel.app/docs/routing/middleware#domain-based-routing) (e.g. `domain.co.uk`) as well as a combination of both. Additionally, an optional [`basePath`](https://next-intl-docs.vercel.app/docs/routing/middleware#base-path) is automatically considered.

What is required, however, is that if a locale shows up in the URL, it is expected to be the same locale that the app uses internally (e.g. `/en` → `en`, `/en-gb` → `en-gb`).

Users have in various issues expressed the need for further customization where a mapping between a locale and a prefix is necessary:

**Examples:**
1. `example.com/uk` should use `en-gb`
2. `example.com/eu` should use `en-gb`
3. `example.com/eu/en` should use `en-gb`
4. `example.com/ca/fr` should use `fr-ca`
5. `example.com/en` should use `en-gb`
6. `example.com/ar` should use `ar-u-nu-arab` (i.e. Arabic with the arab numbering system)

**Relevant issues:**
- https://github.com/amannn/next-intl/issues/609
- https://github.com/amannn/next-intl/issues/549
- https://github.com/amannn/next-intl/issues/650
- https://github.com/amannn/next-intl/discussions/789
- https://github.com/amannn/next-intl/discussions/860
- https://github.com/amannn/next-intl/issues/938
- https://github.com/amannn/next-intl/issues/1069

This affects the rewrites and redirects in [the middleware](https://next-intl-docs.vercel.app/docs/routing/middleware), as well as [the navigation APIs](https://next-intl-docs.vercel.app/docs/routing/navigation). Note that other features like `useTranslations` aren't affected.

We should at least provide a guide or alternatively built-in support in case this is arguably hard to achieve in userland.

There might still be cases that we don't handle, so we should also add docs for how to add your own middleware and routing APIs while still being able to use all component-based APIs from `next-intl`.

### Describe the solution you'd like

`next-intl` relies on a `[locale]` param, therefore a mapping needs to be created for these cases.

[...]

PR #1017, opened 2024-04-24
feat: Support providing a locale in `i18n.ts` instead of reading it from the pathname

### LINKED ISSUE #960: Support providing a locale in `i18n.ts` instead of reading it from the pathname (single-language workflow and prefix-less routing)

While handling multiple languages is the common case, `next-intl` is also expected to work well for apps that only support a single language. After all, if using internationalization right from the start is easier than hardcoding text and formatting, then this should be a common workflow. The Pages Router integration was already capable of this, but ideally we support this similarly for the App Router as well.

Additionally, some apps read the locale from user settings and don't require it to be available in the URL. This use case is closely related (see also https://github.com/amannn/next-intl/discussions/892 and https://github.com/amannn/next-intl/discussions/778).

Maybe a `locale` could be returned in `i18n.tsx`:

```tsx filename="src/i18n.ts"

export default getRequestConfig(async () => ({
  messages: (await import(`../messages/en.json`)).default,

  // Statically set or read from user settings
  locale: 'en'
}));
```

The tricky part is that we currently attempt to fetch the `locale` before this function is called. Maybe it could be lazily evaluated if `locale` is read from the params passed to the function.

This is already possible by supporting only a single language and [setting `localePrefix` to `never`](https://next-intl-docs.vercel.app/docs/routing/middleware#locale-prefix-never). However, the `[locale]` segment is still required.

For reading a locale from the db, this can be used as a workaround: https://github.com/amannn/next-intl/discussions/892#discussioncomment-8607185

With built-in support, no middleware should be required at all.

amannn:
Resolves #960

**Use cases**
1. Apps where the `locale` should be read from user settings instead of the `[locale]` segment
2. Apps where only a single locale is supported

**Prerelease**

```
npm install next-intl@3.12.0-provided-locale.1
```

**Changes**

The gist is instead of reading the locale from the argument that's passed to `getRequestConfig` in `i18n.ts` you can now return a `locale` from the function:

```ts
// i18n.ts

export default getRequestConfig(async () => {
  // This can either be defined statically if only a single locale
  // is supported, or alternatively read from user settings
  const locale = 'en';

  return {
    locale,
    messages: (await import(`../messages/${locale}.json`)).default
  };
});
```

This simplified setup comes with some benefits:

1. No `[locale]` segment needs to be added
3. No routing APIs need to be used (middleware, navigation APIs)
4. Static rendering works without tradeoffs if you only support a single language

[Example implementation](https://github.com/amannn/next-intl/tree/feat/960-provide-locale/examples/example-app-router-without-i18n-routing)

**Docs changes**

- Provide a separate getting started guide for usage without i18n routing
- Remove overview/index pages from table of contents (expandable areas now have a main click area and the expand icon button)
- Adapt docs where necessary to cater to both use cases (with/without i18n routing)
- Mark external links as such

→ [Proposed docs](https://next-intl-docs-git-feat-960-provide-locale-next-intl.vercel.app/)


belgattitude:
Would be awesome ! Thx for your work

AhmedBaset:
> 4. Static rendering works without tradeoffs

How is Static rendering supposed to work for more than one locale (most cases)? 

If the value is derived from a runtime variable, such as userSession, then Dynamic Rendering will be opted for. If the value comes from predefined routes, such as ["en", "de"], then only one locale will be generated, which contradicts the concept of internationalization.

amannn:
@AhmedBaset Yep, that's of course correct—thanks for the feedback! The statement only applies to using a single language, I've changed it accordingly above.

amannn:
My main concern with this currently is that the `locale` is returned asynchronously.

Looking through all the options we support in `i18n.ts`, I think they can be grouped like this:
1. Definitely not async: `formats`, `defaultTranslationValues`, `onError`
2. Likely not async: `now`
3. Might be async: `locale` (if read from user profile), `timeZone`
4. Likely async: `messages`

So generally, it seems reasonable that we support an async config object.

However, this currently has the implication, that the locale can't be read synchronously during an RSC render. This is mostly not an issue, but the navigation APIs are affected by this. For `Link` it's no problem since we can make the component `async`, but for `redirect` this makes a difference.

I need to investigate if there's a way where we can call `redirect` before we know which locale we're redirecting to …

Note that as long as we advertise this feature as something to be used without routing APIs from `next-intl`, this would be fine. I do however have the hope that this could pave the path to a very customizable solution to https://github.com/amannn/next-intl/issues/653, in case the user could read parts of the pathname deeply in an RSC render (see https://github.com/vercel/next.js/discussions/58862). Let's see.

PR #1021, opened 2024-04-25
docs: Update Clerk integration example to Clerk Core 2

gcascio:
Clerk recently release [Clerk Core 2](https://clerk.com/docs/upgrade-guides/core-2/overview) wich changes the procedure of integrating with next-intl. This PR updates the docs to reflect these changes.

### DIFF
--- a/docs/pages/docs/routing/middleware.mdx
+++ b/docs/pages/docs/routing/middleware.mdx
@@ -593,10 +593,14 @@ Note that if you use a [`localePrefix`](#locale-prefix) other than `always`, you
 
 ### Example: Integrating with Clerk
 
-[`@clerk/nextjs`](https://clerk.com/docs/references/nextjs/overview) provides a middleware that can be integrated with `next-intl` by using the [`beforeAuth` hook](https://clerk.com/docs/references/nextjs/auth-middleware#using-before-auth-to-execute-middleware-before-authentication). By doing this, the middleware from `next-intl` will run first, potentially redirect or rewrite incoming requests, followed by the middleware from `@clerk/next` acting on the response.
+[`@clerk/nextjs`](https://clerk.com/docs/references/nextjs/overview) provides a middleware that can be [combined](https://clerk.com/docs/references/nextjs/clerk-middleware#combine-middleware) with other middlewares like the one provided by `next-intl`. By combining them, the middleware from `@clerk/next` will first ensure protected routes are handled appropriately. Subsequently, the middleware from `next-intl` will run, potentially redirecting or rewriting incoming requests.
 
 ```tsx filename="middleware.ts"
-import {authMiddleware} from '@clerk/nextjs';
+import {
+  clerkMiddleware,
+  createRouteMatcher,
+  redirectToSignIn,
+} from '@clerk/nextjs/server';
 import createMiddleware from 'next-intl/middleware';
 
 const intlMiddleware = createMiddleware({
@@ -604,18 +608,21 @@ const intlMiddleware = createMiddleware({
   defaultLocale: 'en'
 });
 
-export default authMiddleware({
-  beforeAuth(request) {
-    return intlMiddleware(request);
-  },
+const isProtectedRoute = createRouteMatcher([
+  'dashboard/(.*)',
+]);
 
-  // Ensure that locale-specific sign in pages are public
-  publicRoutes: ['/:locale', '/:locale/sign-in']
+export default clerkMiddleware((auth, req) => {
+  if (isProtectedRoute(req)) auth().protect();
+
+  return intlMiddleware(req);
 });
 
 export const config = {
-  // Match only internationalized pathnames
-  matcher: ['/', '/(de|en)/:path*']
+  /**
+   * Match all paths except of internal /_next/ routes, static files and api paths
+   */
+  matcher: ['/((?!.+\\.[\\w]+$|_next).*)', '/', '/(api)(.*)'],
 };
 ```
 
--- a/docs/pages/docs/routing/middleware.mdx
+++ b/docs/pages/docs/routing/middleware.mdx
@@ -599,7 +599,6 @@ Note that if you use a [`localePrefix`](#locale-prefix) other than `always`, you
 import {
   clerkMiddleware,
   createRouteMatcher,
-  redirectToSignIn,
 } from '@clerk/nextjs/server';
 import createMiddleware from 'next-intl/middleware';
 
--- a/docs/pages/docs/routing/middleware.mdx
+++ b/docs/pages/docs/routing/middleware.mdx
@@ -608,7 +608,7 @@ const intlMiddleware = createMiddleware({
 });
 
 const isProtectedRoute = createRouteMatcher([
-  'dashboard/(.*)',
+  '/:locale/dashboard(.*)',
 ]);
 
 export default clerkMiddleware((auth, req) => {
@@ -618,10 +618,8 @@ export default clerkMiddleware((auth, req) => {
 });
 
 export const config = {
-  /**
-   * Match all paths except of internal /_next/ routes, static files and api paths
-   */
-  matcher: ['/((?!.+\\.[\\w]+$|_next).*)', '/', '/(api)(.*)'],
+  // Match only internationalized pathnames
+  matcher: ['/', '/(de|en)/:path*'],
 };
 ```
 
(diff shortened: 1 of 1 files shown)


docs/pages/docs/routing/middleware.mdx
```diff
@@ -593,29 +593,36 @@ Note that if you use a [`localePrefix`](#locale-prefix) other than `always`, you
...
 
-[`@clerk/nextjs`](https://clerk.com/docs/references/nextjs/overview) provides a middleware that can be integrated with `next-intl` by using the [`beforeAuth` hook](https://clerk.com/docs/references/nextjs/auth-middleware#using-before-auth-to-execute-middleware-before-authentication). By doing this, the middleware from `next-intl` will run first, potentially redirect or rewrite incoming requests, followed by the middleware from `@clerk/next` acting on the response.
+[`@clerk/nextjs`](https://clerk.com/docs/references/nextjs/overview) provides a middleware that can be [combined](https://clerk.com/docs/references/nextjs/clerk-middleware#combine-middleware) with other middlewares like the one provided by `next-intl`. By combining them, the middleware from `@clerk/next` will first ensure protected routes are handled appropriately. Subsequently, the middleware from `next-intl` will run, potentially redirecting or rewriting incoming requests.
 
 ```tsx filename="middleware.ts"
-import {authMiddleware} from '@clerk/nextjs';
+import {
+  clerkMiddleware,
+  createRouteMatcher,
+  redirectToSignIn,
```
amannn, line 602: This doesn't seem to be used.

docs/pages/docs/routing/middleware.mdx
```diff
@@ -593,29 +593,36 @@ Note that if you use a [`localePrefix`](#locale-prefix) other than `always`, you
...
+  return intlMiddleware(req);
 });
 
 export const config = {
-  // Match only internationalized pathnames
-  matcher: ['/', '/(de|en)/:path*']
+  /**
+   * Match all paths except of internal /_next/ routes, static files and api paths
+   */
+  matcher: ['/((?!.+\\.[\\w]+$|_next).*)', '/', '/(api)(.*)'],
```
amannn, line 625: Is there a need to update the matcher from the previous value? Generally, [we recommend a matcher](https://next-intl-docs.vercel.app/docs/routing/middleware#matcher-config) based on available locales as this is safer than checking for requests not containing a dot for example.

```suggestion
  // Match only internationalized pathnames
  matcher: ['/', '/(de|en)/:path*']
```

amannn (commented):
Many thanks for proposing this update @gcascio! 🙌

docs/pages/docs/routing/middleware.mdx
gcascio: Yes you are correct, I removed it.

docs/pages/docs/routing/middleware.mdx
gcascio: The idea is that the clerk middleware runs on all page routes ensuring that auth checks are performed. Alternatively one could manually ensure that the matcher covers all protected routes, something like this:

```tsx filename="middleware.ts"
const isProtectedRoute = createRouteMatcher([
  '/(de|en)/dashboard(.*)',
]);

// ...

export const config = {
  // Match only internationalized pathnames
  matcher: ['/', '/(de|en)/:path*'],
};
```

Maybe the above would be good for the current example, while more sophisticated use cases should be handled individually. WDYT?

amannn: Hmm, that's a good point.

Previously we had:

```tsx
  publicRoutes: ['/:locale', '/:locale/sign-in']
```

The updated docs now include:

```tsx
const isProtectedRoute = createRouteMatcher([
  'dashboard/(.*)',
]);
```

At least in the default config of `next-intl`, there should be a locale prefix for every route. That's also what the default matcher I've mentioned above assumes.

Does Clerk still support a pattern like `/:locale`? Or would this be `/[\w-]+/dashboard(.*)` with the new config to assume a locale prefix? If we don't need to hardcode locales that's a good thing, the Next.js `matcher` unfortunately requires a static config.

Do you have an app where you've upgraded to the latest Clerk version and is this working for you?

gcascio: Yes Clerk still does support the mentioned path pattern!

I upgraded a production app to the latest Clerk version without issues so far, unfortunately it is closed source. The biggest change IMHO ist that now by default, `clerkMiddleware` treats all pages as public.

I just created a test repo to play around, you can find it [here](https://github.com/gcascio/next-intl-clerk). There I used the `/:locale` pattern in the clerk middleware which looks like this:

```ts
import createMiddleware from 'next-intl/middleware';
import {
  clerkMiddleware,
  createRouteMatcher,
} from '@clerk/nextjs/server';

const intlMiddleware = createMiddleware({
  locales: ['en', 'de'],
  defaultLocale: 'en'
});

const isProtectedRoute = createRouteMatcher([
  '/:locale/dashboard(.*)',
]);

export default clerkMiddleware((auth, req) => {
  if (isProtectedRoute(req)) auth().protect();

  return intlMiddleware(req);
});

export const config = {
  // Match only internationalized pathnames
  matcher: ['/', '/(de|en)/:path*'],
};
```

I think this is as close as it gets to the previous example, if I am not missing something.

amannn: Ah yes, that looks perfect from my perspective! Would you mind updating the PR again with that snippet? Happy to merge then!

gcascio: Great, I updated the PR!

docs/pages/docs/routing/middleware.mdx
```diff
@@ -593,29 +593,33 @@ Note that if you use a [`localePrefix`](#locale-prefix) other than `always`, you
...
 
-  // Ensure that locale-specific sign in pages are public
-  publicRoutes: ['/:locale', '/:locale/sign-in']
+  return intlMiddleware(req);
 });
 
 export const config = {
   // Match only internationalized pathnames
-  matcher: ['/', '/(de|en)/:path*']
+  matcher: ['/', '/(de|en)/:path*'],
```
amannn, line 622:
```suggestion
  matcher: ['/', '/(de|en)/:path*']
```

docs/pages/docs/routing/middleware.mdx
```diff
@@ -593,29 +593,33 @@ Note that if you use a [`localePrefix`](#locale-prefix) other than `always`, you
...
 
-[`@clerk/nextjs`](https://clerk.com/docs/references/nextjs/overview) provides a middleware that can be integrated with `next-intl` by using the [`beforeAuth` hook](https://clerk.com/docs/references/nextjs/auth-middleware#using-before-auth-to-execute-middleware-before-authentication). By doing this, the middleware from `next-intl` will run first, potentially redirect or rewrite incoming requests, followed by the middleware from `@clerk/next` acting on the response.
+[`@clerk/nextjs`](https://clerk.com/docs/references/nextjs/overview) provides a middleware that can be [combined](https://clerk.com/docs/references/nextjs/clerk-middleware#combine-middleware) with other middlewares like the one provided by `next-intl`. By combining them, the middleware from `@clerk/next` will first ensure protected routes are handled appropriately. Subsequently, the middleware from `next-intl` will run, potentially redirecting or rewriting incoming requests.
 
 ```tsx filename="middleware.ts"
-import {authMiddleware} from '@clerk/nextjs';
+import {
+  clerkMiddleware,
+  createRouteMatcher,
+} from '@clerk/nextjs/server';
```
amannn, line 602:
```suggestion
import {clerkMiddleware, createRouteMatcher} from '@clerk/nextjs/server';
```

docs/pages/docs/routing/middleware.mdx
```diff
@@ -593,29 +593,33 @@ Note that if you use a [`localePrefix`](#locale-prefix) other than `always`, you
...
   locales: ['en', 'de'],
   defaultLocale: 'en'
 });
 
-export default authMiddleware({
-  beforeAuth(request) {
-    return intlMiddleware(request);
-  },
+const isProtectedRoute = createRouteMatcher([
+  '/:locale/dashboard(.*)',
```
amannn, line 611:
```suggestion
  '/:locale/dashboard(.*)'
```

amannn (approved):
Just some formatting nits to be consistent, I'll apply the suggestions from the GitHub UI.

PR #1038, opened 2024-05-02
Update documentation for TypeScript usage

ScreamZ:
Hello,

Just proposed some documentation improvements.

### DIFF
--- a/docs/pages/docs/workflows/typescript.mdx
+++ b/docs/pages/docs/workflows/typescript.mdx
@@ -47,3 +47,47 @@ You can freely define the interface, but if you have your messages available loc
 3. The path of your `import` is correct.
 4. Your type declaration file is included in `tsconfig.json`.
 5. Your editor has loaded the most recent type declarations. When in doubt, you can restart.
+
+## Extract `href` for custom components
+
+This library has its own layer for typed routes, but just like the [documentation for NextJS](https://nextjs.org/docs/app/building-your-application/configuring/typescript#statically-typed-links) shows you might want to wrap your router components for some reasons.
+
+In NextJS typed router you proceed that way:
+
+```tsx
+import type { Route } from 'next'
+import Link from 'next/link'
+
+type Props<T extends string> = { href: Route<T> | URL }
+
+function Card<T extends string>({ href }: Props<T>) {
+  return (
+    <Link href={href}>
+      <div>My Card</div>
+    </Link>
+  )
+}
+```
+Using next-intl you can achieve similar behavior this way:
+
+```tsx
+// For the same of the exemple, this is created here but in a real app, this lives in another file.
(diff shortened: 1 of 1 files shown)


amannn:
Hey, thanks for proposing this change! We do in fact already have docs for how to compose the `Link` with its `href` prop in case of localized pathnames: [`Link`](https://next-intl-docs.vercel.app/docs/routing/navigation#link) (click on "Localized pathnames" and see "How can I compose the link with its href prop?" below).

I think that should already cover this sufficiently, what do you think?

Statically typed links for shared pathnames is tracked here: https://github.com/amannn/next-intl/issues/396

ScreamZ:
> Hey, thanks for proposing this change! We do in fact already have docs for how to compose the `Link` with its `href` prop in case of localized pathnames: [`Link`](https://next-intl-docs.vercel.app/docs/routing/navigation#link) (click on "Localized pathnames" and see "How can I compose the link with its href prop?" below).
> 
> I think that should already cover this sufficiently, what do you think?
> 
> Statically typed links for shared pathnames is tracked here: #396

Oh… That was hidden in a dropdown. Got it. Your implementation is a bit more verbose but makes sense too. Thanks for your kind answer :)

I ended up with 

```tsx
export type LinkHref = ComponentProps<typeof L>["href"] | `http${"s" | ""}://${string}` | string;

type LinkProps = Omit<ComponentProps<typeof L>, "href"> & {
  href: LinkHref;
};

/**
 * A link that works for internal or external links.
 * TODO: This is supposed to work as Link from `next/link` handle internal and external link, otherwise we need some logic.
 */
export const Link: React.FC<LinkProps> = ({ href, ...props }: LinkProps) => (
  <L
    {...props}
    href={href as ComponentProps<typeof L>["href"]}
    className={twMerge("transition-opacity hover:opacity-70", props.className)}
  >
    {props.children}
  </L>
);
```

I don't know if it's better, it covers the topic that's right, but I just found that some topics are —sometime— hidden deep in nested things and hard to found. Maybe they could be part of a FAQ section?

Also not sure about having a dedicated TypeScript section alone, or maybe it should backlink to such content like how to compose links? Or sometimes this might be hard to found. What do you think?

The FAQ section makes sense, and is the right answer I think, for all this dropdown stuff. (And we can still backlink each of these dropdown back to the FAQ section too)

amannn:
There was an FAQ section at some point but it ended up just collecting random stuff that didn't fit anywhere else better 🙂. I found that it's better to provide optional expandable sections while you're researching about a particular feature. E.g., if you're exploring the docs for `Link`, it makes sense that there is a hint on how to compose the `href` prop, instead of listing this on a separate page.

The TypeScript docs are actually somewhat an exception currently, maybe we could even move them into the messages docs at some point 🤔. But generally, all of the docs are TypeScript-first.

Hope this makes sense for you!

amannn:
I'm going to close this for the time being as there are already docs about composing the `href` prop. Thanks for the input though @ScreamZ!

Issue #779, opened 2024-01-16
☂️ Umbrella: List of potential features / changes for a major version


amannn:
The big goal for the next major release is to get everything out of the way that a) makes the setup harder than necessary and b) could slow down feature development for the foreseeable future.

This is a list of ideas for API changes that we could potentially introduce in a major version:
- [x] Maybe we could automatically forward `formats` to Client Components with an option to opt-out (`formats={null}`). Keep in mind: https://github.com/amannn/next-intl/pull/633. See previously proposed docs changes in https://github.com/amannn/next-intl/pull/1046/commits/80332fc2a2daf66ea271617deb5abd2b3b10e46e.
- [x] Should nested `NextIntlClientProvider` inherit from each other? That would be very helpful to configure error handling as long as there's no better alternative (see https://github.com/amannn/next-intl/issues/611)
- [x] Go ESM-only? This has been supported [since Next.js 12](https://nextjs.org/blog/next-12#es-modules-support-and-url-imports). See also [Pure ESM package](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c?permalink_comment_id=3850849). `next.config.js` might be the trickiest one, but Node.js 22 (scheduled to be recommended by Autumn '24) supports `require`ing ESM from CJS too. Alternatively we could consider making an exception for `next-intl/plugin` ([ref](https://github.com/vercel/next.js/pull/71162)). If we do this, we can likely also go far a faster size limit preset.
- [x] Target more modern browsers. Currently we compile e.g. rest parameters, which have been supported by browsers since [a long time](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters#browser_compatibility). Not sure if this is a bug in `@babel/preset-env`, as "safari 12" results in rest params being compiled. In exchange, we can reduce bundle size. Does Next.js apply code transforms for backwards compat anyway to `node_modules`? See also various options in next.config.js like [bundlePagesRouterDependencies](https://nextjs.org/docs/app/building-your-application/upgrading/version-15#bundlepagesrouterdependencies).
- [x] Use modern JSX runtime consistently
  - [x] Babel: `['@babel/preset-react', {runtime: 'automatic'}]`
  - [x] TypeScript: `"jsx": "react-jsx"`
- [x] Remove deprecated [`defaultTranslationValues`](https://github.com/amannn/next-intl/pull/1411)
- [x] Remove support for detecting `src/i18n.ts`
[...]

Issue #1, opened 2020-11-25
Automatic tree-shaking of messages


amannn:
It would be beneficial for performance if we could analyse what messages are used on a given page and only return those from `getInitialProps`.

There's [babel-plugin-inline-i18n-messages](https://github.com/hjylewis/babel-plugin-inline-i18n-messages) but it has some limitations with Next.js currently (https://github.com/hjylewis/babel-plugin-inline-i18n-messages/issues/6).

In order to support also static rendering, what we'd have to do is:
1. Before the Next.js build runs, we have to make a webpack build of every page which should [emit all module names](https://stackoverflow.com/questions/29966345/how-to-get-a-list-of-all-files-or-modules-included-per-chunk-in-webpack) on a page. Also lazy loaded components should be included.
2. We can use these module names to read the files and detect which messages are used. By doing this, we'll have to force the user to use the exported APIs directly and will disallow e.g. the wrapping of `useTranslations`. This is a similar limitation as what `react-intl` has.
3. Run the actual Next.js build and use the map of used messages on a page within `getInitialProps` to slice off unneeded messages.

amannn:
An alternative (or possibly addition) is a concept based on the recently announced server components.

Code like this:

```jsx
function LatestFollower({user}) {
  const t = useTranslations('LatestFollower');

  function onFollowClick() {
    UserAPI.follow(user.id);
  }

  return (
    <>
      <Text>{t('latestFollower', {username: user.name})}</Text>
      <IconButton
        aria-label={t('followBack')}
        icon={<FollowIcon />}
        onClick={onFollowClick}
      />
    </>
  );
}
```

… could be turned into:

```jsx
function LatestFollowerServer({user}) {
  const t = useTranslations('LatestFollower');

  return (
    <LatestFollowerClient
      $1={<Text>{t('latestFollower', {username: user.name})}</Text>}
      $2={t('followBack')}
      $3={<FollowIcon />}
    />
  );
}

function LatestFollowerClient({$1, $2, $3, user}) {
  function onFollowClick() {
    UserAPI.follow(user.id);
  }

  return (
    <>
      {$1}
      <IconButton aria-label={$2} icon={$3} onClick={onFollowClick} />
    </>
  );
}
```

These are the interactive features that React currently has:
- useState
- useReducer
- use(Layout)Effect
- refs (and thus useImperativeHandle)
- event listeners

[...]

PR #1046, opened 2024-05-03
docs: Getting started with all messages on the client side

amannn:
I know that users are frequently bitten by messages not being automatically available on the client side. We do this in the best interest of pushing performance as much as possible, but I think this has gone a bit too far. Especially when just starting out with an i18n setup, this can be frustrating.

Furthermore, I think for a number of apps this doesn't make a huge difference anyway. The number and length of messages in apps can vary significantly, so it's hard to make assumptions, but here are two examples from apps I'm maintaining:

1. An emerging one: 88 messages (1,5K gzip)
2. A more sophisticated one: 317 messages (7,6K gzip)

Apps can of course get much larger and it's important to be able to optimize, but it's questionable how early this is needed.

On top of this, app messages aren't the only place where translations are coming from, but typically they represent a part of the total translatable content (with other parts coming from a CMS, backend APIs, etc.).

We're planning on implementing tree-shaking of messages anyway in https://github.com/amannn/next-intl/issues/1, I think for the time being we could be a bit more relaxed about the suggestions we give to users.

→ [New getting started docs for the App Router](https://next-intl-docs-git-docs-client-components-first-next-intl.vercel.app/docs/getting-started/app-router)

~~As part of this, I'd now also consider automatically inheriting `formats` in Client Components if this is defined in `i18n.ts` for similar reasoning.~~ This is a breaking change that will be part of https://github.com/amannn/next-intl/issues/779

### DIFF
--- a/docs/pages/docs/environments/server-client-components.mdx
+++ b/docs/pages/docs/environments/server-client-components.mdx
@@ -23,7 +23,7 @@ Moving internationalization to the server side unlocks new levels of performance
 
 **Benefits of server-side internationalization:**
 
-1. Your messages never leave the server and don't need to be serialized for the client side
+1. Your messages never leave the server and don't need to be passed to the client side
 2. Library code for internationalization doesn't need to be loaded on the client side
 3. No need to split your messages, e.g. based on routes or components
 4. No runtime cost on the client side
@@ -123,7 +123,9 @@ In regard to performance, async functions and hooks can be used very much interc
 
 ## Using internationalization in Client Components
 
-Depending on your situation, you may need to handle internationalization in Client Components as well. There are several options for using translations or other functionality from `next-intl` in Client Components, listed here in order of recommendation.
+Depending on your situation, you may need to handle internationalization in Client Components as well. While providing all messages to the client side is typically the easiest way to [get started](/docs/getting-started/app-router#layout) and a reasonable approach for emerging apps, you can be more selective about which messages are passed to the client side to reduce the bundle size of your app.
+
+There are several options for using translations from `next-intl` in Client Components, listed here in order of enabling the best performance:
 
 ### Option 1: Passing translations to Client Components
 
@@ -221,15 +223,15 @@ To keep internationalization on the server side, it can be helpful to structure 
 ```tsx filename="app/register/page.tsx"
 import {useTranslations} from 'next-intl';
 
-// A Client Component, so that it can use `useFormState` to
-// potentially display errors received after submission.
+// A Client Component, so that `useFormState` can be used
+// to potentially display errors received after submission.
 import RegisterForm from './RegisterForm';
 
-// A Client Component, so that it can use `useFormStatus`
+// A Client Component, so that `useFormStatus` can be used
 // to disable the input field during submission.
 import FormField from './FormField';
 
-// A Client Component, so that it can use `useFormStatus`
+// A Client Component, so that `useFormStatus` can be used
 // to disable the submit button during submission.
 import FormSubmitButton from './FormSubmitButton';
 
@@ -284,7 +286,7 @@ In particular, page and search params are often a great option because they offe
 
 ### Option 3: Providing individual messages
 
-To reduce bundle size, `next-intl` doesn't automatically provide [messages](/docs/usage/configuration#messages) or [formats](/docs/usage/configuration#formats) to Client Components.
+To reduce bundle size, `next-intl` doesn't automatically provide [messages](/docs/usage/configuration#messages) to Client Components.
 
 If you need to incorporate dynamic state into components that can not be moved to the server side, you can wrap these components with `NextIntlClientProvider` and provide the relevant messages.
 
@@ -310,14 +312,12 @@ export default function Counter() {
 }
 ```
 
-In case you prefer to make all messages available to the client side, you can [configure `NextIntlClientProvider` in the root layout](#option-4-providing-all-messages) instead.
-
 <Details id="messages-client-namespaces">
 <summary>How can I know the messages I need to provide to the client side?</summary>
 
 Currently, the messages you select for being passed to the client side need to be picked based on knowledge about the implementation of the wrapped components.
 
-An automatic, compiler-driven approach is being evaluated in [`next-intl#2`](https://github.com/amannn/next-intl/issues/1).
+An automatic, compiler-driven approach is being evaluated in [`next-intl#1`](https://github.com/amannn/next-intl/issues/1).
 
 </Details>
 
@@ -346,11 +346,6 @@ export default function LocaleLayout({children, params: {locale}}) {
   );
 }
 ```
-
-<Callout type="warning">
-  Note that this is a tradeoff in regard to performance (see the bullet points
-  at the top of this page).
-</Callout>
 
 ## Troubleshooting
 
@@ -384,6 +379,7 @@ export default function MyCustomNextIntlClientProvider({
   locale,
   timeZone,
   now,
+  formats,
   ...rest
 }) {
   return (
@@ -396,6 +392,7 @@ export default function MyCustomNextIntlClientProvider({
       locale={locale}
       timeZone={timeZone}
       now={now}
+      formats={formats}
       {...props}
     />
   );
@@ -404,4 +401,4 @@ export default function MyCustomNextIntlClientProvider({
 
 By doing this, your custom provider will already be part of the client-side bundle and can therefore define and pass functions as props.
 
-**Important:** Be sure to pass explicit `locale`, `timeZone` and `now` props to `NextIntlClientProvider` in this case, since the props aren't automatically inherited from a Server Component when you import `NextIntlClientProvider` from a Client Component.
+**Important:** Be sure to pass explicit `locale`, `formats`, `timeZone` and `now` props to `NextIntlClientProvider` in this case, since the props aren't automatically inherited from a Server Component when you import `NextIntlClientProvider` from a Client Component.
--- a/docs/pages/docs/environments/server-client-components.mdx
+++ b/docs/pages/docs/environments/server-client-components.mdx
@@ -286,7 +286,7 @@ In particular, page and search params are often a great option because they offe
 
 ### Option 3: Providing individual messages
 
-To reduce bundle size, `next-intl` doesn't automatically provide [messages](/docs/usage/configuration#messages) to Client Components.
+To reduce bundle size, `next-intl` doesn't automatically provide [messages](/docs/usage/configuration#messages) or [formats](/docs/usage/configuration#formats) to Client Components.
 
 If you need to incorporate dynamic state into components that can not be moved to the server side, you can wrap these components with `NextIntlClientProvider` and provide the relevant messages.
 
@@ -379,7 +379,6 @@ export default function MyCustomNextIntlClientProvider({
   locale,
   timeZone,
   now,
-  formats,
   ...rest
 }) {
   return (
@@ -392,7 +391,6 @@ export default function MyCustomNextIntlClientProvider({
       locale={locale}
       timeZone={timeZone}
       now={now}
-      formats={formats}
       {...props}
     />
   );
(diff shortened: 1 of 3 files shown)


docs/pages/docs/environments/server-client-components.mdx
```diff
@@ -23,7 +23,7 @@ Moving internationalization to the server side unlocks new levels of performance
 
 **Benefits of server-side internationalization:**
 
-1. Your messages never leave the server and don't need to be serialized for the client side
+1. Your messages never leave the server and don't need to be passed to the client side
```
amannn, line 26: Maybe less scary.

docs/pages/docs/environments/server-client-components.mdx
```diff
@@ -123,7 +123,9 @@ In regard to performance, async functions and hooks can be used very much interc
 
 ## Using internationalization in Client Components
 
-Depending on your situation, you may need to handle internationalization in Client Components as well. There are several options for using translations or other functionality from `next-intl` in Client Components, listed here in order of recommendation.
+Depending on your situation, you may need to handle internationalization in Client Components as well. While providing all messages to the client side is typically the easiest way to [get started](/docs/getting-started/app-router#layout) and a reasonable approach for emerging apps, you can be more selective about which messages are passed to the client side to reduce the bundle size of your app.
```
amannn, line 126: "emerging" was an attempt at a better phrasing for "smaller apps" :). Might be more accurate anyway, also big apps start out as small ones.

docs/pages/docs/environments/server-client-components.mdx
```diff
@@ -221,15 +223,15 @@ To keep internationalization on the server side, it can be helpful to structure
 ```tsx filename="app/register/page.tsx"
 import {useTranslations} from 'next-intl';
 
-// A Client Component, so that it can use `useFormState` to
-// potentially display errors received after submission.
+// A Client Component, so that `useFormState` can be used
+// to potentially display errors received after submission.
```
amannn, line 227: Reads slightly better IMO.

docs/pages/docs/environments/server-client-components.mdx
```diff
@@ -310,14 +312,12 @@ export default function Counter() {
 }
 ```
 
-In case you prefer to make all messages available to the client side, you can [configure `NextIntlClientProvider` in the root layout](#option-4-providing-all-messages) instead.
```
amannn, line 313: This is covered in the very next headline.

docs/pages/docs/environments/server-client-components.mdx
```diff
@@ -347,11 +347,6 @@ export default function LocaleLayout({children, params: {locale}}) {
 }
 ```
 
-<Callout type="warning">
-  Note that this is a tradeoff in regard to performance (see the bullet points
-  at the top of this page).
-</Callout>
```
amannn, line 353: Too dramatic IMO, we already told the user everything relevant at the top of the page :).

docs/pages/docs/getting-started/app-router.mdx
```diff
@@ -148,24 +148,45 @@ export const config = {
 
 ### `app/[locale]/layout.tsx` [#layout]
 
-The `locale` that was matched by the middleware is available via the `locale` param and can be used to configure the document language.
+The `locale` that was matched by the middleware is available via the `locale` param and can be used to configure the document language. Additionally, we can use this place to pass configuration from `i18n.ts` to Client Components via `NextIntlClientProvider`.
```
amannn, line 151: Always somewhat of a balance to include necessary information in the getting started guide, but be rather to the point here. We can save many details for later IMO.

docs/pages/docs/getting-started/app-router.mdx
```diff
@@ -148,24 +148,45 @@ export const config = {
...
+export default async function LocaleLayout({
   children,
   params: {locale}
 }: {
   children: React.ReactNode;
   params: {locale: string};
 }) {
+  // Providing all messages to the client
+  // side is the easiest way to get started
+  const messages = await getMessages();
```
amannn, line 165: This component could also use `useMessages`, but I think we can reasonably suggest implementing `LocaleLayout` as an async Server Component. A minor benefit is that this ensures that `NextIntlClientProvider` is rendered from a Server Component now. This enables automatically inheriting configuration from `i18n.ts`. If the user makes the root layout a Client Component, this wouldn't be possible and more wiring is needed.

docs/pages/docs/getting-started/app-router.mdx
```diff
@@ -148,24 +148,45 @@ export const config = {
...
+
+export default async function LocaleLayout({
   children,
   params: {locale}
 }: {
   children: React.ReactNode;
   params: {locale: string};
 }) {
+  // Providing all messages to the client
+  // side is the easiest way to get started
```
amannn, line 164: Inline comments are often copied along to user apps, so this could be a reminder when revisiting this code at a later stage of the project.

docs/pages/docs/usage/configuration.mdx
```diff
@@ -39,10 +39,11 @@ The configuration object is created once for each request by internally using Re
 `NextIntlClientProvider` can be used to provide configuration for **Client Components**.
 
 ```tsx filename="app/[locale]/layout.tsx" /NextIntlClientProvider/
-import {NextIntlClientProvider, useMessages} from 'next-intl';
+import {NextIntlClientProvider} from 'next-intl';
+import {getMessages} from 'next-intl/server';
 
-export default function LocaleLayout({children, params: {locale}}) {
-  const messages = useMessages();
+export default async function LocaleLayout({children, params: {locale}}) {
```
amannn, line 45: Similar reasoning as above.

docs/pages/docs/usage/configuration.mdx
```diff
@@ -95,29 +90,37 @@ export default getRequestConfig(async ({locale}) => {
...
 
 ```tsx
+// Regular components
 import {useMessages} from 'next-intl';
-
 const messages = useMessages();
+
+// Async Server Components
+import {getMessages} from 'next-intl/server';
+const messages = await getMessages();
```
amannn, line 104: `getMessages` might be more common than `useMessages`. This is in contrast to all other APIs from `next-intl`, like `useTranslations`/`getTranslations` etc.

Due to this, it might be a good idea to briefly mention this API on this page too. To be consistent, I've included them for the other config options below as well.

docs/pages/docs/usage/configuration.mdx
```diff
@@ -95,29 +90,37 @@ export default getRequestConfig(async ({locale}) => {
...
 </Tab>
 <Tab>
 
 ```tsx
-import {NextIntlClientProvider, useMessages} from 'next-intl';
+import {NextIntlClientProvider} from 'next-intl';
+import {getMessages} from 'next-intl/server';
 
-// Read messages configured via `i18n.ts`. Alternatively,
-// messages can be fetched from any other source too.
```
amannn, line 113: Reading from other sources shouldn't be relevant here (see also "How can I load messages from remote sources?" above).

docs/pages/docs/getting-started/app-router.mdx
```diff
@@ -56,8 +56,6 @@ Now, set up the plugin which creates an alias to provide your i18n configuration
 <Tabs items={['next.config.mjs', 'next.config.js']}>
 <Tab>
 
-If you're using ECMAScript modules for your Next.js config, you can use the plugin as follows:
```
amannn, line 59: Not really necessary IMO.

docs/pages/docs/getting-started/app-router.mdx
```diff
@@ -72,8 +70,6 @@ export default withNextIntl(nextConfig);
 </Tab>
 <Tab>
 
-If you're using CommonJS for your Next.js config, you can use the plugin as follows:
```
amannn, line 75: Same as above.

docs/pages/docs/environments/server-client-components.mdx
lmac-1: I know that the aim of this docs rewrite is to not scare or overwhelm people with performance implications, but is there a way to somewhere define what is "emerging" and what is considered a bigger app in terms of number of messages? 

Perhaps in an expandable section here or in [this](https://next-intl-docs-git-docs-client-components-first-next-intl.vercel.app/docs/getting-started/app-router#nextintlclientprovider-messages) expandable section in Getting Started.

docs/pages/docs/getting-started/app-router.mdx
lmac-1: In which part of the docs is there more information on this?

docs/pages/docs/environments/server-client-components.mdx
amannn: I've tried to rephrase this again in https://github.com/amannn/next-intl/pull/1046/commits/30ab6304b4ec54c388b395a303af71e3a8ba32ab and included some more details.

Coming back to this, I think even "emerging" is the wrong term here. For some apps, splitting messages will never be a necessity. 

I think a particular number of messages doesn't really answer this question here, in the latest commit I tried to focus on pointing you towards metrics and tools that can help you to ensure your app meets your performance goals.

Do you find this helpful?

I've now also removed the expandable section from the getting started docs. I think the comment "Providing all messages to the client side is the easiest way to get started" should be sufficient in terms of setting up an app.

docs/pages/docs/getting-started/app-router.mdx
amannn: The [global configuration docs](https://next-intl-docs-git-docs-client-components-first-next-intl.vercel.app/docs/usage/configuration) have all the details about this.

I'd prefer to be quite selective about which links to include in the getting started docs and therefore not link to this page at this point. Links tend to create branches of user flows where a developer is just trying to get something up and running.

In the latest iteration I've now removed all links that would create "branches" and instead only used them if prerequisite knowledge is required (e.g. what a dynamic segment like `[locale]` is in Next.js).

Once the user has finished the setup, we can help the developer continue with the links in the "Next steps" section.

Does that sound reasonable to you?

amannn:
@lmac-1 I'll merge this in the current form, as I believe this is already quite an improvement and so that I can stack proposed docs from https://github.com/amannn/next-intl/pull/1017 on top of this. Happy to follow-up though if you find some places that could use rephrasing or clarification!

docs/pages/docs/getting-started/app-router.mdx
lmac-1: I completely agree. I think that the getting started steps should be as simple as possible to allow users to get things working and then provide links later to allow for more exploration. Otherwise, as you said, users can get a bit distracted reading into things and perhaps confused.

docs/pages/docs/environments/server-client-components.mdx
lmac-1: Just had a look at the latest commit. I think that this is much better. In the previous version with "emerging", I might have read that and been worried that I would need to optimise things from the get-go if my app was on a grander scale. 

I really like the newer explanation, particularly where you remind users to "always measure before you optimize".

docs/pages/docs/environments/server-client-components.mdx
lmac-1: Definitely agree, the wording scared me off a bit when I read this for the first time 😁

lmac-1 (commented):
I know this is already merged, but just commenting to say that I think that the changes are great. I really like that you've simplified the getting started steps. It should make it easier for users to get up and running in a basic way and can then dive into the docs if they're wanting to understand things more deeply. 

I guess my only other comment would be perhaps the App Router example should be updated to match the new getting started steps? Or there could be two examples? 

Not sure but just thinking I might be confused if I follow the getting started steps, then want to see a full working example, and can see that it hasn't followed the steps that I did during the initial steps. What do you think?

amannn:
> I guess my only other comment would be perhaps the App Router example should be updated to match the new getting started steps? 

That's a great point, yes! I've added a follow-up PR to address this here: https://github.com/amannn/next-intl/pull/1060

Thanks a ton for your feedback here! 🙏

PR #1053, opened 2024-05-06
docs: Fix incorrect locale in example

Piorunx22:
Corrected to `en_GB`. `en_BR` would be Brazilian English :D


amannn (approved):
Oh man, you're right—thanks a lot! 😅

Issue #1036, opened 2024-05-01
ESLint plugin


amannn:
### Is your feature request related to a problem? Please describe.

This issue acts to collect ideas for an ESLint rule that `next-intl` could provide to support developers.

**Ensure navigation APIs from `next-intl` are used**

If you're using the [navigation APIs from `next-intl`](https://next-intl-docs.vercel.app/docs/routing/navigation), you might want to ensure your team uses these consistently instead of the ones from Next.js.

Note that this can already be achieved with a custom rule based on `no-restricted-syntax`:

```tsx
  rules: {
    'no-restricted-syntax': [
      'error',
      {
        selector: "ImportDeclaration[source.value='next/link']",
        message: "Please use `import {Link} from '@/navigation'` instead."
      }
    ]
  },
```

We might want to at least provide docs for this.

**Linting for unused translations**

External tools support this, but we could consider supporting this if there's an ESLint plugin.

See also: https://github.com/amannn/next-intl/discussions/503

### Describe the solution you'd like

-

### Describe alternatives you've considered

-

Gawdfrey:
I think a lot of the power in the lint rules here are that they have a customisable message that allows the developer to point others to the correct location, unsure on how doable/ergonomic it is do that with a eslint package 🤔 
But, I am currently using these rules:
```json
  "rules": {
    "no-restricted-imports": [
      "error",
      {
        "name": "next/link",
        "message": "Please use @shared/navigation instead."
      },
      {
        "name": "next/navigation",
        "importNames": [
          "redirect",
          "permanentRedirect",
          "useRouter",
          "usePathname"
        ],
        "message": "Please use @shared/navigation instead."
      },
      {
        "name": "next/router",
        "message": "Please use @shared/navigation instead."
      }
    ]
  }
```
Just banning the usage of the function that are exported from `next-intl` and pointing to where I export them. The last rule is just to avoid usage of the old router.

I definitely think some of these should be documented somehow :)

amannn:
Oh that's a good idea with `no-restricted-imports`, even easier to manage!

Being able to provide a customizable message is certainly possible with ESLint, but you'd have to configure at least the rule yourself—just using a preset wouldn't be enough. I think as a first step, adding a page to [the workflows section](https://next-intl-docs.vercel.app/docs/workflows) with the rules you're using would be a good idea!

gndelia:
Something I've seen developers do when using this libs, is concatenating translated strings, like

```ts
[...]

PR #1082, opened 2024-05-22
docs: add documentation for linting with basic examples

Gawdfrey:
This PR adds basic docs for linting, based on discussions in #1036.

We could maybe add biomejs examples as well seeing as it is rising in popularity?

Let me know if we want any more changes

### DIFF
--- a/docs/pages/docs/routing/navigation.mdx
+++ b/docs/pages/docs/routing/navigation.mdx
@@ -13,6 +13,11 @@ import Details from 'components/Details';
 
 ## Strategies
 
+<Callout>
+  Use linting rules to enforce the usage of the different strategies. See the
+  [Linting](/docs/workflows/linting) guide for more information.
+</Callout>
+
 There are two strategies that you can use based on your needs.
 
 **Shared pathnames:** The simplest case is when your app uses the same pathnames, regardless of the locale.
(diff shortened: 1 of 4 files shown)


amannn (commented):
Thanks a lot for setting up the page! I've added a PR with a few ideas for the content in https://github.com/Gawdfrey/next-intl/pull/1, let me know what you think!

> We could maybe add biomejs examples as well seeing as it is rising in popularity?

Generally yes, but it seems like their [`noRestrictedImports`](https://biomejs.dev/linter/rules/no-restricted-imports/) is currently experimental. Also it seems like they forbid a whole package name, not just specific imports. Maybe worth waiting until this stabilizes a bit!

But I think it's a good idea to call the page "linting", so we could potentially add it in the future!

Gawdfrey:
> Thanks a lot for setting up the page! I've added a PR with a few ideas for the content in [Gawdfrey#1](https://github.com/Gawdfrey/next-intl/pull/1), let me know what you think!
> 
> > We could maybe add biomejs examples as well seeing as it is rising in popularity?
> 
> Generally yes, but it seems like their [`noRestrictedImports`](https://biomejs.dev/linter/rules/no-restricted-imports/) is currently experimental. Also it seems like they forbid a whole package name, not just specific imports. Maybe worth waiting until this stabilizes a bit!
> 
> But I think it's a good idea to call the page "linting", so we could potentially add it in the future!

Yeah I saw the biome bit, I think that is a good idea.

amannn:
Many thanks! 🙌

PR #1084, opened 2024-05-22
fix: Handle external URLs passed to `redirect` and `permanentRedirect`

mhogeveen:
This PR addresses an issue where absolute URL's used as pathnames in the `redirect` and `permanentRedirect` functions returned from `createSharedPathnamesNavigation` are localized.

**Current behaviour:**

```ts
const { redirect, permanentRedirect } =
  createSharedPathnamesNavigation({
    locales: ['en-GB'],
    localePrefix: 'as-needed',
  })

redirect('https://example.com')
// will result in a redirect to '/en-GBhttps://example.com'

permanentRedirect('https://example.com')
// will result in a redirect to '/en-GBhttps://example.com'
```

**Expected behaviour:**

```ts
const { redirect, permanentRedirect } =
  createSharedPathnamesNavigation({
    locales: ['en-GB'],
    localePrefix: 'as-needed',
  })

redirect('https://example.com')
// will result in a redirect to 'https://example.com'

permanentRedirect('https://example.com')
// will result in a redirect to 'https://example.com'
```

**Proposed solution:**

Include a check whether to localize the given pathname based on the path starting with a `/` in:
- `packages/next-intl/src/navigation/shared/baseRedirect.tsx`
- `packages/next-intl/src/navigation/shared/basePermanentRedirect.tsx`

If there are any questions, feedback, and or further improvements required please let me know 😄

### DIFF
--- a/packages/next-intl/package.json
+++ b/packages/next-intl/package.json
@@ -122,7 +122,7 @@
     },
     {
       "path": "dist/production/navigation.react-client.js",
-      "limit": "2.96 KB"
+      "limit": "2.965 KB"
     },
     {
       "path": "dist/production/navigation.react-server.js",
--- a/packages/next-intl/src/navigation/shared/basePermanentRedirect.tsx
+++ b/packages/next-intl/src/navigation/shared/basePermanentRedirect.tsx
@@ -4,7 +4,7 @@ import {
   LocalePrefix,
   ParametersExceptFirst
 } from '../../shared/types';
-import {prefixPathname} from '../../shared/utils';
+import {isLocalHref, prefixPathname} from '../../shared/utils';
 
 export default function basePermanentRedirect(
   params: {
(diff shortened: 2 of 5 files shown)


mhogeveen:
I've just updated the PR to use the same external link detection used in the `ClientLink` component and I've added a couple of tests for the `baseRedirect` and `basePermanentRedirect` functions. That covers point 1 & 2. Point 3 however already seems to be the case:

https://github.com/amannn/next-intl/blob/312956e08dd59535a353c198811e5b60b6e6794d/packages/next-intl/src/navigation/react-client/useBaseRouter.tsx#L44

https://github.com/amannn/next-intl/blob/312956e08dd59535a353c198811e5b60b6e6794d/packages/next-intl/src/shared/utils.tsx#L33-L53

If I'm missing something let me know!

amannn (approved):
Fantastic PR, thanks a lot @mhogeveen! 🙌

I've added another commit with minor cleanup stuff, hope you don't mind. 

Thanks also for checking `useRouter`! You're right, all the necessary handling was already in place there, only the redirect functions were missing.

mhogeveen:
Thanks for the quick response @amannn. It's much appreciated!

Issue #940, opened 2024-03-14
Path rewriting does not work for the home page


ponjs:
### Description

I'm rewriting `/about` as `/` but when I visit the `/` page, the url address bar is `/about`

https://github.com/amannn/next-intl/assets/39897744/d274635e-f168-4525-8b17-b696549a26ac

### Mandatory reproduction URL

https://stackblitz.com/edit/stackblitz-starters-nwjigc?file=src%2Fmiddleware.ts

### Reproduction description

Visit stackblitz's open prview in new tab and visit `/` page

### Expected behaviour

When the `/` page is accessed, the page is a `/about` page, but the url is still the `/` page.

amannn:
For the record, navigating to a locale root works as expected:

![Screenshot 2024-03-15 at 10 42 34](https://github.com/amannn/next-intl/assets/4038316/d659ca08-4b01-432f-88b6-1663982d1363)

So as far as I understand the issue, this is about the location that's picked when there's a locale-specific redirect at the root:

![Screenshot 2024-03-15 at 10 43 12](https://github.com/amannn/next-intl/assets/4038316/9b004773-bab2-480a-bbea-0e0f40f81d01)

As a workaround, you can consider adding your own redirect at the root in the middleware, but I think it would be helpful to investigate wether there's something to address here in `next-intl`.

amannn:
I've refactored the middleware handling somewhat in #1086 and have added a test for your use case. Should work in the future after the PR is merged!

PR #1086, opened 2024-05-23
feat: Add support for custom prefixes for i18n routing

amannn:
Resolves https://github.com/amannn/next-intl/issues/653

### Information

The `localePrefix` option for `createMiddleware` as well as the navigation APIs now not only accepts a string, but an object where custom prefixes can be added per locale:

```tsx
import {LocalePrefix} from 'next-intl/routing';

export const locales = ['en-US', 'de-AT', 'zh'] as const;

export const localePrefix = {
  mode: 'always',
  prefixes: {
    'en-US': '/us',
    'de-AT': '/eu/at'
    // (/zh will be used as-is)
  }
} satisfies LocalePrefix<typeof locales>;
```

In addition to this implementation, I've restructured the routing docs to discuss config options that are shared among the middleware and navigation APIs on a single page. We might expand on this in the future (see below).

See the [proposed docs](https://next-intl-docs-git-feat-custom-prefixes-next-intl.vercel.app/docs/routing#locale-prefix-custom).

**Important:** As mentioned in the docs, you might have to adjust your middleware matcher to handle the custom prefixes.

### Prerelease

```
next-intl@3.15.0-custom-prefixes.0
```

### Changes
- Add support for custom prefixes for i18n routing
- Fix initial render when passing a relative href to `Link` (`<Link href="about" />`)
- Deprecate `import {Pathnames} from 'next-intl/navigation'` in favor of `import {Pathnames} from 'next-intl/routing'`

### Potential future changes

I'm considering adding an API like `const routing = defineRouting({locales: ['en', 'de'], ...})` to gather all routing-related config. This could be passed as `createMiddleware({routing, ...})` (see https://github.com/amannn/next-intl/issues/779).

(also resolves https://github.com/amannn/next-intl/issues/940)


examples/example-app-router/src/config.ts
```diff
@@ -1,22 +1,23 @@
...
 // Use the default: `always`
 export const localePrefix = undefined;
+export const localePrefix: LocalePrefixConfig<typeof locales> = 'always';
 
 export type AppPathnames = keyof typeof pathnames;
+
+export const port = process.env.PORT || 3000;
+export const host = process.env.VERCEL_URL
+  ? `https://${process.env.VERCEL_URL}`
+  : `http://localhost:${port}`;
```
amannn, line 23: API considerations:

```
import {createSharedPathnamesNavigation} from 'next-intl/navigation';

// # Option 0: Don't provide a general API
//
// good:
// - middleware-only config can be set there
// - individual parts can be imported in granular way
// - don't overabstract now and possibly regret it later
//
// bad:
// - autocomplete requires more wiring (but we already do this)
// - need to make sure you're passing everything where it's needed

// # Option 1: Satisfies
//
// good:
// - autocomplete for pathnames without satisfies Pathnames<typeof locales>
// - discoverability
// - opt-in typing (is this good?)
//
// bad:
// - individual parts can't be passed to components (e.g. locales without pathnames)
//   realistically maybe this makes no difference
export const routing = {
  defaultLocale: 'en',
  locales: ['en', 'de'],
  pathnames: {
    '/': '/',
    '/pathnames': {
      en: '/pathnames',
      de: '/pfadnamen'
    }
  },
  localePrefix: {
    mode: 'always',
    prefixes: {
      // en: '/uk',
      de: '/ch/de'
    }
  }
} satisfies RoutingConfig;

// # Option 2: defineRouting
// good:
// - maybe more beginner friendly? (not much though)
// - somehow cool and magical that everything works without opt-in types (look ma, no types)
// - known from cool popular tools like vite, cypress
// - matches i18n
//
// bad:
// - affects bundle size (not much though)
// - return type a bit magical?

  defaultLocale: 'en',
  locales: ['en', 'de'],
  pathnames: {
    '/': '/',
    '/pathnames': {
      en: '/pathnames',
      de: '/pfadnamen'
    }
  },
  localePrefix: {
    mode: 'always',
    prefixes: {
      // en: '/uk',
      de: '/ch/de'
    }
  }
});

createSharedPathnamesNavigation(routing);
// two boolean props … really a need to pass them to middleware?
createMiddleware({...routing, alternateLinks: false, localeDetection: false});
createMiddleware({routing, alternateLinks: false, localeDetection: false}); // could do this in a breaking change
```

PR #1098, opened 2024-05-30
Update metadata-route-handlers.mdx

Unsleeping:
update link to Static Rendering Anchor


Unsleeping:
will be fixed in !1099

PR #1106, opened 2024-06-03
fix: update react19 api

Unsleeping:
found a reference to useFormState which was renamed in react-19-rc to useActionState
see: https://react.dev/reference/react/useActionState


amannn (approved):
Thanks a lot! 🙌

PR #1111, opened 2024-06-06
Removed unused code from 'Wihout i18n routing documentation'

AbhayVAshokan:
The PR removes a line of code from 'Without i18n routing documentation' which is unused in the `src/i18n.ts` file.


amannn (approved):
Good point, thanks!

PR #1140, opened 2024-06-19
fix: Update `@formatjs/intl-localematcher` to latest version

amannn:
Includes updates to the distance algorithm for matching locales


nhducit:
LGTM 🙏

Issue #1136, opened 2024-06-18
Matcher string in the docs is buggy


tkrotoff:
### Description

In the documentation https://next-intl-docs.vercel.app/docs/routing/middleware#matcher-no-prefix

```JS
matcher: [
  // Match all pathnames except for
  // - … if they start with `/api`, `/_next` or `/_vercel`
  // - … the ones containing a dot (e.g. `favicon.ico`)
  '/((?!api|_next|_vercel|.*\\..*).*)',
  // However, match all pathnames within `/users`, optionally with a locale prefix
  '/([\\w-]+)?/users/(.+)'
]
```

The part `.*\\..*` matches any route that contains a dot. It's not just about files like favicon.ico

This is not what you want, the middleware won't be executed with these examples and many others:
- `https://example.com/hello.world`
- `https://example.com/welcome?message=hello.world`

Instead:

```JS
matcher: "/((?!api|_next/static|_next/image|favicon.ico|icon.svg|apple-icon.png|manifest.webmanifest|robots.txt).*)"`
```

- it follows Next.js documentation https://nextjs.org/docs/app/building-your-application/routing/middleware#matcher
- it's explicit

### Verifications

- [X] I've verified that the problem I'm experiencing isn't covered in [the docs](https://next-intl-docs.vercel.app/).
- [X] I've searched for similar, existing issues on GitHub and [Stack Overflow](https://stackoverflow.com/search?q=next-intl).
- [X] I've compared my app to [a working example](https://next-intl-docs.vercel.app/examples) to look for differences.

### Mandatory reproduction URL

https://example.com/hello.world

### Reproduction description

Steps to reproduce:
1. Open reproduction
2. Click on …
3. See error: … 

### Expected behaviour

Middleware matcher should all ignore all URLs containing a dot

amannn:
There are unfortunately tradeoffs with every approach we could suggest.

Your proposal with excluding only specific files has the tradeoff that it will incorrectly match other static assets a developer might add (e.g. `/hero.jpg` might result in a rewrite to `/en/hero.jpg`). The developer needs to be careful to add all assets to the negative lookahead in the matcher.

The issues you're pointing out with the currently suggested matcher are discussed with the `/users` example in the code snippet and is also mentioned in the paragraph right above.

The most frictionless approach from my experience is the matcher that we can suggest for the default `localePrefix` strategy (see [matcher config](https://next-intl-docs.vercel.app/docs/routing/middleware#matcher-config)).

PR #1180, opened 2024-07-09
Middleware matcher showed on docs does not work properly

GustavoMelloGit:
I was having trouble finding out why my intl wasnt working. After a lot of research i have found that i just needed to change the matcher.


amannn:
Thanks for opening a PR! There are unfortunately tradeoffs with every solution, in your case you have to be careful to adapt the matcher once you add assets in the `public` folder. See also https://github.com/amannn/next-intl/issues/1136.

Which issue did you run into precisely? Did you add a locale and haven't updated the matcher? Or when did the issue occur?

amannn:
Going to close this due to inactivity …

PR #1187, opened 2024-07-10
docs: Update mixed routing example with improved compatibility for Server Actions

imCorfitz:
Passing the request headers as response headers messes with server actions. https://github.com/vercel/next.js/issues/50659 

We need to pass the updated request headers to the NextResponse's request object.

### DIFF
--- a/examples/example-app-router-mixed-routing/src/middleware.ts
+++ b/examples/example-app-router-mixed-routing/src/middleware.ts
@@ -14,7 +14,9 @@ export default function middleware(request: NextRequest) {
   if (isAppRoute) {
     // Add a hint that we can read in `i18n.ts`
     request.headers.set('x-app-route', 'true');
-    return NextResponse.next({headers: request.headers});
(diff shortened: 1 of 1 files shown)


examples/example-app-router-mixed-routing/src/middleware.ts
```diff
@@ -14,7 +14,9 @@ export default function middleware(request: NextRequest) {
   if (isAppRoute) {
     // Add a hint that we can read in `i18n.ts`
     request.headers.set('x-app-route', 'true');
-    return NextResponse.next({headers: request.headers});
+    return NextResponse.next({request: {
+      headers: request.headers
+    });
```
amannn, line 19: Satisfy the linter and add a missing bracket …

```suggestion
    return NextResponse.next({request: {headers: request.headers}});
```

amannn (approved):
Looks good, thanks a lot @imCorfitz!

PR #1193, opened 2024-07-11
feat: Cache `Intl.*` constructors

### LINKED ISSUE #215: Cache Intl.* constructors

It seems like on every call to `format.number`, or similar, there's a call to the corresponding `Intl.NumberFormat` constructor. This is potentially expensive: A microbenchmark that I constructed suggests a 97% performance decrease (see https://jsbench.me/09lf9t531d/1). An older blog post suggests a more dramatic performance difference: https://blog.david-reess.de/posts/hBEx9w-on-number-formatting-and-performance.

As someone coming from FormatJS, whose documentation emphasizes the use of `createIntlCache` (see https://formatjs.io/docs/intl), I was a bit surprised to see that this optimization is not present in `next-intl`.

Admittedly, on modern hardware like my M2 MacBook Pro, the absolute performance seems OK for both cases, even if there's a 49x relative difference.

I'd suggest caching the construction of `Intl.*Format` objects between calls, perhaps by the `createFormatter` function, or in the context.

Admittedly, this would add complexity and size to the library, for the caching and memoization functionality, and might not be considered worth it.

Alternatively, if this is a known trade-off, I think it'd be good to mention in the documentation that it's been made. For example, in the FAQ section "[How is this library different from using react-intl?](https://next-intl-docs.vercel.app/docs/faq#how-is-this-library-different-from-using-react-intl)".

amannn:
Fixes #215 

`Intl.*` constructors like `Intl.DateTimeFormat` are now cached across components in order to improve rendering performance (see [benchmark](https://github.com/amannn/next-intl/issues/215#issuecomment-2180169107)).

**Details:**
 - The message format cache that was already available has now been integrated with the cache for `Intl.*` constructors.
 - Cache lifetime:
   - Client: The cache is released when the locale prop of the provider changes. For long-running apps with a persistent provider at the root, this can reduce the memory footprint (e.g. in React Native).
   - Server: Lifetime of the request.
 - The `.format` calls are currently not cached. The reason is that while `Intl.*` constructor caching can work across various component instances and formatting calls, `.format` calls typically only affect a single call site. The overhead of memoization (serialization of args, diffing, bundle size) therefore might not be a good tradeoff here. In this area, avoiding re-renders is much more relevant, with the ideal solution being React Compiler.


packages/use-intl/package.json
```diff
@@ -63,6 +63,7 @@
     "formatting"
   ],
   "dependencies": {
+    "@formatjs/fast-memoize": "^2.2.0",
```
amannn, line 66: Already in use due to `intl-messageformat`

packages/next-intl/package.json
```diff
@@ -120,27 +120,27 @@
   "size-limit": [
     {
       "path": "dist/production/index.react-client.js",
-      "limit": "15.785 KB"
+      "limit": "16.055 KB"
```
amannn, line 123: I've added a note to https://github.com/amannn/next-intl/issues/779 to make sure we target modern browsers, currently we're compiling too much syntax (e.g. rest params).

Issue #1240, opened 2024-08-07
router issue when switching locale: path.startsWith is not a function


chaotic-justice:
### Description

![image](https://github.com/user-attachments/assets/53f786e3-e504-4456-a0ba-34755c2b2020)

![image](https://github.com/user-attachments/assets/c05cccad-bc7c-49a2-a1db-82f8b6708c30)

TypeError: path.startsWith is not a function
encountered when switching locale

### Verifications

- [X] I've verified that the problem I'm experiencing isn't covered in [the docs](https://next-intl-docs.vercel.app/).
- [X] I've searched for similar, existing issues on GitHub and [Stack Overflow](https://stackoverflow.com/search?q=next-intl).
- [X] I've compared my app to [a working example](https://next-intl-docs.vercel.app/examples) to look for differences.

### Mandatory reproduction URL

https://github.com/chaotic-justice/summer-bee

### Reproduction description

Steps to reproduce:
1. Open reproduction
2. Click on `git clone`
3. run `yarn && yarn dev`
4. See error: … 

### Expected behaviour

switching to a diff locale will trigger the routing error

r0skar:
I am seeing the same issue.

```bash
"next": "14.2.5",
"next-intl": "3.17.2",
```

meabed:
I have the same issue

tgdn:
Same issue

nahsooyeon:
```tsx
import { useParams} from "next/navigation";
import { usePathname, useRouter } from "@/i18n";
....

  const router = useRouter();
  const pathname = usePathname();
  const params = useParams();
  const [isPending, startTransition] = useTransition();


  const changeLanguage = (newLocale: string) => {
    const nextLocale = newLocale;
    startTransition(() => {
      router.replace(pathname, {
        // @ts-expect-error -- TypeScript will validate that only known `params`
           locale: nextLocale
         })
    })
  };
  ```

I solved the issue on this way.

NaderTate:
I do that as a workaround:
`<button
                    onClick={() => {
                      startTransition(() => {
                        const newPath = pathname.replace(
                          lang,
                          lang === "en" ? "ar" : "en"
                        );
                        push(newPath, { scroll: false });
                      });
                    }}
                  >
                    {lang === "en" ? "AR" : "EN"}
                   
                  </button>`

tgdn:
Same issue

tgdn:
[...]

PR #1256, opened 2024-08-14
docs: Adjust locale switcher example to provide more type safety

amannn:
Related:
- https://github.com/amannn/next-intl/issues/1240
- https://github.com/amannn/next-intl/pull/853

### DIFF
--- a/examples/example-app-router/src/components/LocaleSwitcherSelect.tsx
+++ b/examples/example-app-router/src/components/LocaleSwitcherSelect.tsx
@@ -4,6 +4,7 @@ import clsx from 'clsx';
 import {useParams} from 'next/navigation';
 import {ChangeEvent, ReactNode, useTransition} from 'react';
 import {useRouter, usePathname} from '@/navigation';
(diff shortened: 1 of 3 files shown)


examples/example-app-router/src/components/LocaleSwitcherSelect.tsx
```diff
@@ -22,13 +23,16 @@ export default function LocaleSwitcherSelect({
...
-        // @ts-expect-error -- TypeScript will validate that only known `params`
-        // are used in combination with a given `pathname`. Since the two will
-        // always match for the current route, we can skip runtime checks.
-        {pathname, params},
+        {
+          pathname,
+          // @ts-expect-error -- TypeScript will validate that only known `params`
+          // are used in combination with a given `pathname`. Since the two will
+          // always match for the current route, we can skip runtime checks.
+          params
```
amannn, line 34: By moving the directive here, we keep the validation that an object can be received by `router.replace`. In case of shared pathnames, this will show an error.

PR #1269, opened 2024-08-18
docs: Fixed case of variable in example doc

dave-meyer:
Fixed case of 'host' variable in example on 'Server Actions, Metadata & Route Handlers' page.
Once the variable was written in uppercase, once in lowercase. I changed the uppercase variant to lowercase, since that is how the 'host' variable in another example is spelled.


amannn (approved):
Thanks a lot @dave-meyer! 🙌

Issue #1268, opened 2024-08-17
Add Support for Setting HttpOnly and Secure Flags on NEXT_LOCALE Cookie


Michael-Grupp:
### Is your feature request related to a problem? Please describe.

Yes, the issue is that the cookie set by `next-intl` does not have the `HttpOnly` flag. This can present a security risk because cookies without this flag are accessible via JavaScript, which could potentially be exploited by cross-site scripting (XSS) attacks. 

Additionally, when running security tests, such as with Mozilla Observatory, the cookie set by `next-intl` is flagged because it does not have the `Secure` flag. While transmission over HTTP is prevented by HSTS, adding the `Secure` flag would ensure that the cookie is only sent over HTTPS, providing an additional layer of security.

### Describe the solution you'd like

I would like the `next-intl` library to support setting the `HttpOnly` and `Secure` flags on the cookies it creates. Specifically:

- The `HttpOnly` flag should be added to prevent client-side scripts from accessing the cookies, reducing the risk of XSS attacks.
- The `Secure` flag should be set on cookies to ensure they are only sent over HTTPS, even when HSTS is enabled.

The implementation could involve adding options to the existing cookie configuration, allowing developers to easily enable these flags as needed.

### Describe alternatives you've considered

One alternative is to manually set these flags on the cookies after they are created by `next-intl`, but this approach is less desirable because:

- It adds additional complexity to the codebase, requiring developers to remember to manually secure cookies.
- It increases the risk of inconsistencies or errors, particularly in larger projects where this might be overlooked.

amannn:
Hey @Michael-Grupp, thank you for your thoughtful report!

`Secure`: That's a good point, I've just added https://github.com/amannn/next-intl/pull/1272 to add this. Do you by chance know if this could be a breaking change for any consumers? As far as I can tell only, in case someone would host a Next.js app without HTTPS, which seems very unlikely to me. Would you agree?

`HttpOnly`: This is currently not set, because `next-intl` does in fact change the cookie on the client side (see https://github.com/amannn/next-intl/pull/790). I understand that ideally we'd not have to do this, but due to the way Next.js works, this is currently necessary as far as I can tell.

> One alternative is to manually set these flags on the cookies after they are created by next-intl, but this approach is less desirable

I agree, ideally `next-intl` would work out-of-the-box without any changes necessary.

Michael-Grupp:
[...]

PR #1272, opened 2024-08-20
feat: Set `secure` flag for `NEXT_LOCALE` cookie (ON HOLD)

amannn:
This is a security enhancement and will disallow users to read the `NEXT_LOCALE` cookie if the website is served over HTTP. The only exception is on `localhost`.

See also: [Block access to your cookies on MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#block_access_to_your_cookies)

If you'd like to unset this flag, you can [compose the middleware](https://next-intl-docs.vercel.app/docs/routing/middleware#composing-other-middlewares) and modify the cookie accordingly before the response is returned.

Fixes https://github.com/amannn/next-intl/issues/1268

### DIFF
--- a/packages/next-intl/src/middleware/syncCookie.tsx
+++ b/packages/next-intl/src/middleware/syncCookie.tsx
@@ -16,7 +16,8 @@ export default function syncCookie(
     response.cookies.set(COOKIE_LOCALE_NAME, locale, {
       path: request.nextUrl.basePath || undefined,
       sameSite: COOKIE_SAME_SITE,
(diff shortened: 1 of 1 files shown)


amannn:
As discussed in https://github.com/amannn/next-intl/issues/1268#issuecomment-2303986513, the `secure` flag might not be a good idea for all users. Instead, we now provide an opt-in way for this via `localeCookie` (available in [v3.22](https://github.com/amannn/next-intl/releases/tag/v3.22.0)).

PR #1305, opened 2024-08-29
fix: Print warning for inconsistent i18n setup where no `locale` is read in `getRequestConfig` and also none is returned

amannn:
/cc @raphaelbadia


raphaelbadia (approved):
damn you're on fire

PR #1316, opened 2024-09-03
feat: `createNavigation`

amannn:
This PR provides a new **`createNavigation`** function that supersedes the previously available APIs:
1. `createSharedPathnamesNavigation`
2. `createLocalizedPathnamesNavigation`

This new function is a reimplementation of the existing navigation functionality that unifies the API for both use cases and also fixes a few quirks in the previous APIs.

**Usage**

```tsx
import {defineRouting} from 'next-intl/routing';
 
export const routing = defineRouting(/* ... */);
 
export const {Link, redirect, usePathname, useRouter} =
  createNavigation(routing);
```

(see the [updated navigation docs](https://next-intl-docs-git-canary-next-intl.vercel.app/docs/routing/navigation))

**Improvements**
1. A single API can be used both for shared as well as localized pathnames. This reduces the API surface and simplifies the corresponding docs.
2. `Link` can now be composed seamlessly into another component with its `href` prop without having to add a generic type argument.
3. `getPathname` is now available for both shared as well as localized pathnames (fixes https://github.com/amannn/next-intl/discussions/785)
4. For improved compatibility, `router.push` and `redirect` now accept search params consistently via the object form (e.g. `router.push({pathname: '/users', query: {sortBy: 'name'})`)—regardless of if you're using shared or localized pathnames. You can still use `router.push('/users?sortBy=name')` if you prefer though.
5. When using `localePrefix: 'as-needed'`, the initial render of `Link` now uses the correct pathname immediately during SSR (fixes [#444](https://github.com/amannn/next-intl/issues/444)). Previously, a prefix for the default locale was added during SSR and removed during hydration. Also `redirect` now gets the final pathname right without having to add a superfluous prefix (fixes [#1335](https://github.com/amannn/next-intl/issues/1335)). The only exception is when you use `localePrefix: 'as-needed'` in combination with `domains` (see [Special case: Using `domains` with `localePrefix: 'as-needed'`](https://next-intl-docs-git-canary-next-intl.vercel.app/docs/routing#domains-localeprefix-asneeded))
6. Slightly smaller in size: `createNavigation` weighs 2.98 kB (`createSharedPathnamesNavigation` was 3.01 kB, `createLocalizedPathnamesNavigation` was 3.27 kB)
7. Prepares `next-intl` for Next.js 15 to run without warnings.

**Migrating to `createNavigation`**

`createNavigation` is generally considered a drop-in replacement, but a few changes might be necessary:
1. `createNavigation` is expected to receive your complete routing configuration. Ideally, you define this via the [`defineRouting`](https://next-intl-docs.vercel.app/docs/routing#define-routing) function and pass the result to `createNavigation`.
2. If you've used `createLocalizedPathnamesNavigation` and have [composed the `Link` with its `href` prop](https://next-intl-docs.vercel.app/docs/routing/navigation#link-composition), you should no longer provide the generic `Pathname` type argument (see [updated docs](https://next-intl-docs-git-canary-next-intl.vercel.app/docs/routing/navigation#link-composition)).
```diff
- ComponentProps<typeof Link<Pathname>>
+ ComponentProps<typeof Link>
```
3. If you've used [`redirect`](https://next-intl-docs.vercel.app/docs/routing/navigation#redirect), you now have to provide an explicit locale (even if it's just [the current locale](https://next-intl-docs.vercel.app/docs/usage/configuration#locale)). The previously passed href (whether it was a string or an object) now needs to be wrapped in an object and assigned to the `href` prop. This change was necessary for an upcoming change in Next.js 15 where `headers()` turns into a promise (see [#1375](https://github.com/amannn/next-intl/issues/1375) for details).
```tsx
// Retrieving the current locale
// ... in regular components:
const locale = useLocale();
// ... in async components:
```
```diff
- redirect('/about')
+ redirect({href: '/about', locale})

- redirect({pathname: '/users/[id]', params: {id: 2}})
+ redirect({href: {pathname: '/users/[id]', params: {id: 2}}, locale})
```
4. If you've used [`getPathname`](https://next-intl-docs.vercel.app/docs/routing/navigation#getpathname) and have previously manually prepended a locale prefix, you should no longer do so—`getPathname` now takes care of this depending on your routing strategy.
```diff
- '/'+ locale + getPathname(/* ... */)
+ getPathname(/* ... */);
```
5. If you're using a combination of `localePrefix: 'as-needed'` and `domains` and you're using `getPathname`, you now need to provide a `domain` argument (see [Special case: Using `domains` with `localePrefix: 'as-needed'`](https://next-intl-docs-git-canary-next-intl.vercel.app/docs/routing#domains-localeprefix-asneeded))

### DIFF
--- a/packages/next-intl/.size-limit.ts
+++ b/packages/next-intl/.size-limit.ts
@@ -15,31 +15,31 @@ const config: SizeLimitConfig = [
     name: 'import {createSharedPathnamesNavigation} from \'next-intl/navigation\' (react-client)',
     path: 'dist/production/navigation.react-client.js',
     import: '{createSharedPathnamesNavigation}',
-    limit: '3.855 KB'
+    limit: '3.885 KB'
   },
   {
     name: 'import {createLocalizedPathnamesNavigation} from \'next-intl/navigation\' (react-client)',
     path: 'dist/production/navigation.react-client.js',
     import: '{createLocalizedPathnamesNavigation}',
-    limit: '3.855 KB'
+    limit: '3.885 KB'
   },
   {
     name: 'import {createNavigation} from \'next-intl/navigation\' (react-client)',
     path: 'dist/production/navigation.react-client.js',
     import: '{createNavigation}',
-    limit: '3.865 KB'
+    limit: '3.885 KB'
   },
   {
     name: 'import {createSharedPathnamesNavigation} from \'next-intl/navigation\' (react-server)',
     path: 'dist/production/navigation.react-server.js',
     import: '{createSharedPathnamesNavigation}',
-    limit: '16.485 KB'
+    limit: '16.515 KB'
   },
   {
     name: 'import {createLocalizedPathnamesNavigation} from \'next-intl/navigation\' (react-server)',
     path: 'dist/production/navigation.react-server.js',
     import: '{createLocalizedPathnamesNavigation}',
-    limit: '16.475 KB'
+    limit: '16.545 KB'
   },
   {
     name: 'import {createNavigation} from \'next-intl/navigation\' (react-server)',


packages/next-intl/.size-limit.ts
```diff
@@ -2,43 +2,79 @@ import type {SizeLimitConfig} from 'size-limit';
...
+    name: 'import {createLocalizedPathnamesNavigation} from \'next-intl/navigation\' (react-client)',
+    path: 'dist/production/navigation.react-client.js',
+    import: '{createLocalizedPathnamesNavigation}',
+    limit: '3.825 KB'
+  },
+  {
+    name: 'import {createNavigation} from \'next-intl/navigation\' (react-client)',
+    path: 'dist/production/navigation.react-client.js',
+    import: '{createNavigation}',
+    limit: '3.825 KB'
```
amannn, line 30: I think tree shaking in size limit has a bug. If only `createNavigation` is exported, the size amounts to 2.95 kB.

For comparison, exporting only `createSharedPathnamesNavigation` accounts for 2.97 kB and `createLocalizedPathnamesNavigation` for 3.24 kB.

docs/pages/docs/routing.mdx
```diff
@@ -84,19 +84,16 @@ export const routing = defineRouting({
...
 #### Never use a locale prefix [#locale-prefix-never]
 
 If you'd like to provide a locale to `next-intl`, e.g. based on user settings, you can consider setting up `next-intl` [without i18n routing](/docs/getting-started/app-router/without-i18n-routing). This way, you don't need to use the routing integration in the first place.
 
 However, you can also configure the middleware to never show a locale prefix in the URL, which can be helpful in the following cases:
 
-1. You're using [domain-based routing](#domains) and you support only a single locale per domain
-2. You're using a cookie to determine the locale but would like to enable static rendering
+1. You want to use [domain-based routing](#domains) and have only one locale per domain
+2. You want to use a cookie to determine locale while enabling static rendering
```
amannn, line 96: Typo

docs/pages/docs/routing.mdx
```diff
@@ -373,3 +363,72 @@ export const routing = defineRouting({
...
+
+<Details id="domains-localeprefix-asneeded">
+<summary>Special case: Using `domains` with `localePrefix: 'as-needed'`</summary>
+
+Since domains can have different default locales, this combination requires some tradeoffs that apply to the [navigation APIs](/docs/routing/navigation) in order for `next-intl` to avoid reading the current host on the server side (which would prevent the usage of static rendering).
+
+1. [`<Link />`](/docs/routing/navigation#link): This component will always render a locale prefix on the server side, even for the default locale of a given domain. However, during hydration on the client side, the prefix is potentially removed, if the default locale of the current domain is used. Note that the temporarily prefixed pathname will always be valid, however the middleware will potentially clean up a superfluous prefix via a redirect if the user clicks on a link before hydration.
+2. [`redirect`](/docs/routing/navigation#redirect): When calling this function, a locale prefix is always added, regardless of the current locale and domain. However, similar to the handling with `<Link />`, the middleware will potentially clean up a superfluous prefix.
```
amannn, line 408: Adapt

packages/next-intl/src/navigation/shared/createSharedNavigationFns.tsx
```diff
@@ -0,0 +1,260 @@
...
+        ref={ref}
+        href={{
+          ...(typeof href === 'object' && href),
+          // @ts-expect-error -- This is ok
+          pathname: finalPathname
+        }}
+        locale={locale}
+        // Provide the minimal relevant information to the client side in order
+        // to potentially remove the prefix in case of the `forcePrefixSsr` case
+        unprefixConfig={
```
amannn, line 141: Check again

packages/next-intl/src/navigation/shared/utils.tsx
```diff
@@ -198,3 +221,100 @@ export function getBasePath(
...
+            console.error(
+              "You're using a routing configuration with `localePrefix: 'as-needed'` in combination with `domains`. In order to compute a correct pathname, you need to provide a `domain` parameter.\n\nSee: https://next-intl-docs.vercel.app/docs/routing#domains-localeprefix-asneeded"
+            );
+          } else {
+            // If a domain was provided, but it wasn't found in the routing
+            // configuration, this can be an indicator that the user is on
+            // localhost. In this case, we can simply use the domain-agnostic
+            // default locale.
+          }
+        }
```
amannn, line 284: Check again

DevOsaWebsite:
>@DevOsaWebsite Thanks for providing feedback here!

> Some questions:

> Besides updating the version, did you upgrade to createNavigation?
> Are you referring to the markup resulting from the server side render, or the final markup after hydration that renders in the browser?
> Are you using domains in combination with localePrefix: 'as-needed'?
> Please refer to #1316, you should find an answer there.


1. Yes. You can see it here -- [](https://stackoverflow.com/questions/79062267/next-intl-the-link-component-renders-a-url-with-defaultlocale) 

2. About SSR (I am primarily a technical SEO + marketing specialist), SSR and avoidance of third-party effects are important for me. Therefore, before writing a ticket, I checked with the eyes of a Google bot: in the Google browser console with the disabled JS, through View Page Source, and through the HTML validator. In all cases, there was a defaultLocale prefix in the DOM, which will create a lot of problems for multi-page sites.
Screenshot 2024-10-07 at 19 33 38

As far as I understand, the prefix is ​​important for the feature related to locale detection for "x-middleware-rewrite". Then I'm 
then I tried to turn it off

`const handleI18nRouting = createMiddleware(routing, {
    localeDetection: false,
  });`
but the problem is still there

then i went into the code and found the function

**ClientLink.js**
`  const finalLocale = locale || defaultLocale;
  const prefix = utils.getLocalePrefix(finalLocale, localePrefix);
`
**utils.js**
`function getLocalePrefix(locale, localePrefix) {
  var _localePrefix$prefixe;
  return localePrefix.mode !== 'never' && ((_localePrefix$prefixe = localePrefix.prefixes) === null || _localePrefix$prefixe === void 0 ? void 0 : _localePrefix$prefixe[locale]) ||
  // We return a prefix even if `mode: 'never'`. It's up to the consumer
  // to decide to use it or not.
  '/' + locale;
}`

locale || defaultLocale - no chance 😀😀😀

3. not use **domains**, but use **as-needed**
`import {DEFAULT_LOCALE, LOCALES} from '@/config/locale';
import {type DomainsConfig} from 'next-intl/routing';

const locales = LOCALES as never as string[];
export const config: {
  locales: string[];
  defaultLocale: string;
  localePrefix: 'always' | 'as-needed' | 'never';
  domains?: DomainsConfig<string[]>;
} = {
  locales,
  defaultLocale: DEFAULT_LOCALE,
  localePrefix: 'as-needed',
};
`

amannn:
@DevOsaWebsite 

Thanks for including a link to your Stack Overflow question!

The reason why a prefix is rendered, is because you're linking to the default locale from a locale switcher (i.e. you're providing the `locale` prop). In this case, it's expected to link to `/en`, because if the user is on a secondary locale (that is prefixed) and would navigate to `/`, the cookie would redirect the user back to the secondary locale.

Instead, in this case the user should navigate to `/en` explicitly, update the cookie value and then go back to `/`.

Here's a test for the relevant implementation:

https://github.com/amannn/next-intl/blob/6b2087378db405091d04a7a79537c20c4646a1ed/packages/next-intl/src/navigation/createNavigation.test.tsx#L643-L651

If you want to avoid this, you could for example implement a locale switcher based on a select and use `useRouter().push` instead (the [App Router with i18n routing example](https://next-intl-docs.vercel.app/examples#app-router) does that).

Does this help? I should probably mention this in the docs …

Issue #1112, opened 2024-06-07
Strict typing of global formats


amannn:
### Is your feature request related to a problem? Please describe.

Similar to how we support [strict typing of messages](https://next-intl-docs.vercel.app/docs/workflows/typescript), we could provide strict typing for formats.

### Describe the solution you'd like

Allow the user to define `IntlFormats` to enable this.

### Describe alternatives you've considered

Not providing strict formats.

dBianchii:
Hey, I think I might be interested in implementing this. I was thinking of setting it close to their type definition for messages, like:

```ts
import en from './messages/en.json'
import formats from '~/i18/request.ts'
 
type Messages = typeof en;
 
declare global {
  // Use type safe message keys with `next-intl`
  interface IntlMessages extends Messages {}
type IntlFormats = typeof formats

}
```

amannn:
If you're up for it, that would be 💯!

As far as I know, it has to be an interface in order to work. Interfaces have this special ability in TypeScript that [declarations can be merged](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) when re-declared.

Based on the implementation for the messages, I think these are the required steps:
1. Add a local type in `use-intl/types` that will be used for type checking within this repo:
https://github.com/amannn/next-intl/blob/aa60ddebea13ddbd9bf8377b28774c4a9839cd7d/packages/use-intl/types/index.d.ts#L1-L3
2. Use the type were relevant:
https://github.com/amannn/next-intl/blob/aa60ddebea13ddbd9bf8377b28774c4a9839cd7d/packages/use-intl/src/react/useTranslations.tsx#L24

Now, when a consumer installs the package, the `IntlMessages` type is not available, therefore defaulting to `any`. As soon as the consumer declares the type on the `global` namespace, this type will be picked up where relevant.

Maybe with a condition we could still ensure that only a string is accepted for a local format in case no `IntlFormats` are provided? That would probably be something to be researched.

Let me know in case I can help with something! 🙌

dBianchii:
> If you're up for it, that would be 💯!
> 
> As far as I know, it has to be an interface in order to work. Interfaces have this special ability in TypeScript that [declarations can be merged](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) when re-declared.
> 
> Based on the implementation for the messages, I think these are the required steps:
> 
> 1. Add a local type in `use-intl/types` that will be used for type checking within this repo:
[...]

PR #1346, opened 2024-09-18
feat: Typesafe Globa Formats

dBianchii:


dBianchii:
Hi @ amann/maintainers

You said [in here](https://github.com/amannn/next-intl/issues/1112#issuecomment-2355575809) that the package supports `full`,  `long`, `medium` and `short`. I believe that this only applies for [dates and times within messages](https://next-intl-docs.vercel.app/docs/usage/dates-times#dates-and-times-within-messages), yes? In that comment, did you suggest to make those the defaults as well for the formatter?

You also said:
> For [number formatting](https://next-intl-docs.vercel.app/docs/usage/numbers) there are defaults for currency and percent.

I haven't found this logic in the code, and also in my tests when I do `format.number(5, "percent")` or `format.number(5, "currency")`, nothing happens

Could you explain a bit more of what you meant?

packages/use-intl/src/core/createFormatter.tsx
```diff
@@ -284,7 +284,7 @@ export default function createFormatter({
   type FormattableListValue = string | ReactElement;
   function list<Value extends FormattableListValue>(
     value: Iterable<Value>,
-    formatOrOptions?: string | Intl.ListFormatOptions
+    formatOrOptions?: IntlFormats['list'] | Intl.ListFormatOptions
```
amannn, line 287: I noticed for `format.list` this doesn't work yet, I guess this is why:

```suggestion
    formatOrOptions?: keyof IntlFormats['list'] | Intl.ListFormatOptions
```

- [ ] Fix type checking for `.list`

packages/use-intl/types/index.d.ts
```diff
@@ -1,7 +1,15 @@
...
-// by the consumer for optional type safety
+// by the consumer for optional type safety of messages
 declare interface IntlMessages extends Record<string, any> {}
 
+// This type is intended to be overridden
+// by the consumer for optional type safety of formats
+declare interface IntlFormats {
+  dateTime: {};
+  number: {};
+  list: {};
```
amannn, line 10: I think we could avoid all the `// @ts-expect-error` declarations in the test code by using a more flexible definition like this:

```suggestion
  dateTime: any;
  number: any;
  list: any;
```

- [ ] Avoid `@ts-expect-error`

examples/example-app-router/src/i18n/request.ts
```diff
@@ -1,6 +1,29 @@
...
+    }
+  },
+  list: {
+    enumeration: {
+      style: 'long',
+      type: 'conjunction'
+    }
+  }
+} as const satisfies Partial<Formats>
+
```
amannn, line 26: Regarding tests, I currently use `example-app-router-playground` for type tests regarding `t`. Having the types directly inside `use-intl` and `next-intl` would definitely be preferable, but I haven't figured out a way to achieve that yet.

You can have a look here for inspiration:

https://github.com/amannn/next-intl/blob/2b363635db6542d726e6a4231bf1ac579fad7322/examples/example-app-router-playground/src/components/AsyncComponent.tsx#L15-L23

I'm also wondering if it'd be possible to test both cases, with an explicit declaration and without one, both in `example-app-router-playground` ([potential solution](https://www.perplexity.ai/search/in-typescript-override-a-globa-4qthXmTpRJengYPNaW9riA#0)).

- [ ] Tests for `useFormatter` and `getFormatter`

examples/example-app-router/src/i18n/request.ts
```diff
@@ -1,6 +1,29 @@
 import {notFound} from 'next/navigation';
 import {getRequestConfig} from 'next-intl/server';
 import {routing} from './routing';
+import { Formats } from 'next-intl';
```
amannn, line 4: Can you execute [the linter](https://github.com/amannn/next-intl/blob/main/CONTRIBUTORS.md#code-formatting-eslint--prettier) after your changes?

- [ ] Fix code formatting

packages/use-intl/src/core/createFormatter.tsx
```diff
@@ -163,7 +163,7 @@ export default function createFormatter({
     value: Date | number,
     /** If a time zone is supplied, the `value` is converted to that time zone.
      * Otherwise the user time zone will be used. */
-    formatOrOptions?: string | DateTimeFormatOptions
+    formatOrOptions?: keyof IntlFormats['dateTime'] | DateTimeFormatOptions
```
amannn, line 166: One important characteristic is that if the user doesn't supply types for global formats, then the overload of `DateTimeFormatOptions` still works. Seems to be the case based on my tests 👌

amannn (commented):
Damn you're quick! 🔥🔥

I somehow left this feature on the backlog for a long time, but you made it happen just over night! 😄 Super cool, thanks a lot!

I've left a few comments inline, let me know what you think!

Btw. regarding the defaults across usage in messages / via `format.*`: You're absolutely right, I've opened https://github.com/amannn/next-intl/issues/1347 as a separate issue to discuss this. This is definitely out of scope for this PR and type-safety is only relevant for `format.*` calls, therefore we should assume no defaults there.

packages/use-intl/src/core/createFormatter.tsx
dBianchii: But if the user doesn't supply the types for global formats, then the type should be as it was before yes? `string | DateTimeFormatOptions`

I'm still going to work on making it fallback to string when types are not provided

examples/example-app-router/src/i18n/request.ts
dBianchii: Yeah, so I am having some problems with formatting the code. I had setup the defaultFormatter/formatOnSave on user `settings.json` based on the CONTRIBUTORS.md file, but still it didn't work. I managed to format it by using the cli command though.

image

Why use eslint for formatting instead of prettier?

packages/use-intl/src/core/createFormatter.tsx
amannn: If that is possible, that would be ideal—yes!

examples/example-app-router/src/i18n/request.ts
amannn: The common issue that people run into is this:

> For ESLint to work correctly, you should open individual packages you work on as the workspace root (e.g. `./packages/next-intl`).

But yes, I know the toolchain is a bit controversial. Just a note that Prettier is still used, however via ESLint.

I'd like to get rid of this though and have opened https://github.com/amannn/next-intl/issues/1349 now. Thanks for the nudge!

dBianchii: Ok, I more or less managed to make it work. Wasn't familiar with this type of workflow before. I am having to use >format Document (forced) for now

packages/use-intl/src/core/createFormatter.tsx
```diff
@@ -284,7 +296,11 @@ export default function createFormatter({
   type FormattableListValue = string | ReactElement;
   function list<Value extends FormattableListValue>(
     value: Iterable<Value>,
-    formatOrOptions?: string | Intl.ListFormatOptions
+    formatOrOptions?:
+      | (keyof IntlFormats['list'] extends string
```
dBianchii, line 300: For code readibility I tried to create a type helper for this like that:

```
type GlobalFormat<T extends keyof Formats> =
    keyof IntlFormats[T] extends string ? keyof IntlFormats[T] : string;
```

So then we could do
```
formatOrOptions?: GlobalFormat<"dateTime"> | Intl.ListFormatOptions
```

But I don't know why, it's not working, so I just caved and placed this verbose thing for now

dBianchii:
There is one problem with this right now:

I see that the global format api has been designed to be able to handle global formats for client and server differently/separately. (We must provide formats to both `getRequestConfig` and `NextIntlClientProvider` component. Each can take in different global formats. ) I don't really understand why it is this way, because in my opinion it would be better if we could just use the same global formats everywhere (client and server). The way I set this up makes it impossible to separate both types, and the user can only provide one type as `IntlFormats`. 

How can we resolve this problem?

examples/example-app-router/src/i18n/request.ts
dBianchii: will do!

amannn:
>  it would be better if we could just use the same global formats everywhere (client and server)

Yep, I agree!

The discussion is related to the one about messages: Messages are defined globally, but it's the user's job to provide the relevant messages where relevant. This is to allow the user to optionally pass only certain messages to the client side to optimize bundle size, which is also the reason why the `messages` need to be passed explicitly to `NextIntlClientProvider`.

The current state of the library has the same mechanism for `formats`: In order to potentially reduce bundle size, formats aren't automatically inherited and need to be passed explicitly to `NextIntlClientProvider`. I've changed my mind in the meantime about this though: Out-of-the-box formats should be inherited automatically, and therefore the user doesn't have to worry. You could still opt-out with `formats={null}`. This change is already implemented on the `v4` branch (see https://github.com/amannn/next-intl/pull/1191), but the release is still a bit further down the road.

That being said, I think it's fine to model `IntlFormats` the same as the global `IntlMessages`, especially with the change in v4 coming up.

Hope that sounds reasonable to you!

dBianchii:
That's awesome to hear. Makes total sense to me. I'll continue with this decision in mind. Thanks!

examples/example-app-router/src/i18n/request.ts
dBianchii: I was not able to adapt `example-app-router-playground` to have just one file with a different `IntlFormat` type. I guess we can do this same test but on a different app/workspace, where we don't have yet declared `IntlFormat` ?

dBianchii: Oh, just to let you know, I can't click on these checkboxes you are adding :upside_down_face:

dBianchii:
Ok, so I added a few tests to `app-router-playground`. Please let me know if I should add more. As I mentioned in the comment, I am only testing an environment that has the global `IntlFormat` type declared. Maybe if you'd like I could do it as well in another workspace.

Please let me know if you think that this implementation is good enough, or if there could be something missing/edge cases. Whenever you think all is good, I'll go ahead and start working on the docs. I have to say I'm very impressed with the documentation on this project. I'll do my best to follow suit and try to explain things in the best manner possible.

amannn:
Looks awesome, thanks a lot! I had another look over the changed files and cleaned up a few minor things along the way and added some more tests for the case when no strict formats are supplied: https://github.com/amannn/next-intl/pull/1346/commits/3fa91e89cdafda17303416660329f6ec528238a3. Hope that's ok for you!

I think the implementation is solid now! 💯

You wanna give the docs a shot? 🙂 Means a lot to me if you like them! 😊

examples/example-app-router/src/i18n/request.ts
amannn: I've added a few tests directly in `use-intl`, there we we have the use case of no strict formats available!

dBianchii:
Ready to go. Please check though: I don't know why but I am getting some problems running `pnpm i` for this branch and also `main` branch.

image

I noticed I accidentally pushed a bunch pnpm lockfile changes, so I tried to use the correct pnpm version to revert it but no luck in running the installer.

Feel free to make any further changes

docs/pages/docs/workflows/typescript.mdx
```diff
@@ -2,6 +2,8 @@ import Callout from 'components/Callout';
 
 # TypeScript integration
 
+## Strict typing of messages
```
amannn, line 5: Maybe a bit shorter?

```suggestion
## Messages
```

(note that this would change the permalink)

docs/pages/docs/workflows/typescript.mdx
```diff
@@ -47,3 +49,76 @@ You can freely define the interface, but if you have your messages available loc
 3. The path of your `import` is correct.
 4. Your type declaration file is included in `tsconfig.json`.
 5. Your editor has loaded the most recent type declarations. When in doubt, you can restart.
+
+## Strict typing of global formats
```
amannn, line 53:
```suggestion
## Formats
```

docs/pages/docs/workflows/typescript.mdx
```diff
@@ -47,3 +49,76 @@ You can freely define the interface, but if you have your messages available loc
...
+
+`next-intl` provides a way to define [global formats](/docs/usage/configuration#formats) for your application. You can provide the type of your formats similarly to how you [define the type of messages](#strict-typing-of-messages).
```
amannn, line 55:
```suggestion
`next-intl` provides a way to define [global formats](/docs/usage/configuration#formats) for your application. You can provide the type of your formats similarly to how you define the type of [messages](#strict-typing-of-messages).
```

Over time I found that reducing the amount of words that are linked increases readability a bit. There are still areas in the docs where this can be improved, this looks like a chance too.

docs/pages/docs/workflows/typescript.mdx
```diff
@@ -2,6 +2,8 @@ import Callout from 'components/Callout';
 
 # TypeScript integration
 
+## Strict typing of messages
+
 `next-intl` integrates with TypeScript out-of-the box without additional setup. You can however provide the shape of your messages to get autocompletion and type safety for your namespaces and message keys.
```
amannn, line 7: I think this sentence could be split into a generic sentence right below the main headline and then one specific to messages here. What do you think?

docs/pages/docs/workflows/typescript.mdx
```diff
@@ -47,3 +49,76 @@ You can freely define the interface, but if you have your messages available loc
...
+export default getRequestConfig(async ({locale}) => {
+  // ...
+ 
+  return {
+    formats,
+  }
+});
+```
+
+Then in your `global.d.ts` at the root of your project, define the `IntlFormats` type:
```
amannn, line 93:
```suggestion
Then, in your `global.d.ts` at the root of your project, define the `IntlFormats` type:
```

docs/pages/docs/workflows/typescript.mdx
```diff
@@ -27,7 +29,7 @@ function About() {
 
 To enable this validation, add a global type definition file in your project root (e.g. `global.d.ts`):
 
-```jsx filename="global.d.ts"
+```ts filename="global.d.ts"
```
amannn, line 32: 👍

docs/pages/docs/workflows/typescript.mdx
```diff
@@ -47,3 +49,76 @@ You can freely define the interface, but if you have your messages available loc
...
+declare global {
+  // Use type safe formats with `next-intl`
+  interface IntlFormats extends Formats {}
+}
+```
+
+Now you will have full autocompletion and type safety for your formats:
+
+```tsx filename="CurrentDate.tsx"
+function CurrentDate() {
```
amannn, line 109:
```suggestion
function Component() {
```

… since it does a few different things :)

examples/example-app-router-playground/src/i18n/request.tsx
```diff
@@ -1,9 +1,31 @@
 import {headers} from 'next/headers';
 import {notFound} from 'next/navigation';
+import type {Formats} from 'next-intl';
```
amannn, line 3: Is there a benefit here from your perspective for using `import type`?

For simplicity, I think I'd leave the `type` part out—that's at least what I usually do.

amannn (commented):
Super cool, the new docs for the feature look great to me! 💯

I only left a few minor comments, let me know what you think!

> I don't know why but I am getting some problems running pnpm i for this branch and also main branch.

Hmm, that's odd. For me it works on `main` with:

```
❯ node -v
v20.11.1

❯ pnpm -v
8.15.0
```

What I did notice is that sometimes when there's been a release on `main` and a branch is slightly outdated, that upon `pnpm i` a version from `npm` gets installed instead of the local one from the workspace is used. In case you run into this, you could consider updating this branch with the latest changes from `main`.

dBianchii:
I was thinking of moving [this section](https://github.com/amannn/next-intl/pull/1346/files#diff-1db885aa3a3e363e26f3a05341ae5af6b71b493a4655807b2436e1565ea3150eR45-R53) to the bottom of the page, so it talks about both cases. But for now I just left it like that because I think that moving it all the way down would require creating another heading which I don't want to do

amannn (approved):
I had another final look and rephrased a few minor things along the way. I agree that the troubleshooting section can now be shared for both types.

I've also adapted the `Formats` type in https://github.com/amannn/next-intl/pull/1367 for easier consumption.

All good from my side now, waiting for CI to turn green! 🚀

amannn:
I did another final test in `example-app-router-playground` and found that interestingly, while type checking was strict and errors were reported where expected, auto-complete of TypeScript was broken.

I've experimented with a few options with this test suite:

```tsx
/* eslint-disable @typescript-eslint/no-unused-vars */

const formats = {
  dateTime: {
    medium: {
      dateStyle: 'medium',
      timeStyle: 'short',
      hour12: false
    }
  }
};

type Formats = typeof formats;
interface IntlFormats extends Formats {}

type T1 = keyof IntlFormats['dateTime'] extends string
  ? keyof IntlFormats['dateTime'] | Intl.DateTimeFormatOptions
  : string | Intl.DateTimeFormatOptions;
type T2 = keyof IntlFormats['dateTime'] | Intl.DateTimeFormatOptions;
type T3 =
  | Extract<keyof IntlFormats['dateTime'], string>
  | Intl.DateTimeFormatOptions;

// Type checking works, but no auto-completion
const t1: T1 = 'medium';

// Auto-completion, no string default
const t2: T2 = 'medium';

// Auto-completion and string default ✅
const t3: T3 = 'medium';
```

ChatGPT gave me the tip of using `Extract`, this seems to cover all cases now: Auto-complete, errors where expected and a string default 💯. I've adapted this accordingly in https://github.com/amannn/next-intl/pull/1346/commits/cca51345c8520af2bdc24796500eaa070965159d.

PR #1365, opened 2024-09-23
fix: typescript support

sahandsn:


docs/pages/docs/workflows/typescript.mdx
```diff
@@ -34,7 +34,7 @@ type Messages = typeof en;
 
 declare global {
   // Use type safe message keys with `next-intl`
-  interface IntlMessages extends Messages {}
+  type IntlMessages = Messages
```
amannn, line 37: Nope, only an interface will work correctly.

Issue #1335, opened 2024-09-12
`redirect` does not respect `as-needed` locale prefix


jorgemoya:
### Description

When using locale prefix `as-needed`, I would expect `redirect` to respect this configuration, instead it will append default locale. This is not how using `Link` behaves, so it is unexpected. For non default locales, I would expect for `redirect` to append locale prefix.

### Verifications

- [X] I've verified that the problem I'm experiencing isn't covered in [the docs](https://next-intl-docs.vercel.app/).
- [X] I've searched for similar, existing issues on GitHub and [Stack Overflow](https://stackoverflow.com/search?q=next-intl).
- [X] I've compared my app to [a working example](https://next-intl-docs.vercel.app/examples) to look for differences.

### Mandatory reproduction URL

https://next-intl-redirect-issue-of1rlux37-jorgemoyas-projects.vercel.app/

### Reproduction description

Steps to reproduce:
1. Open reproduction
2. Click on `Click to redirect`
3. Error: Notice how default locale is appended to url. Navigating with header links works as expected.

### Expected behaviour

The expected behavior would be to respect the `as-needed` locale prefix configuration and for navigation within the default locale to ignore appending locale, much like using the `Link` component. For non default locales it would be fine to append locale prefix.

amannn:
Thanks for the report! I'm currently working on a re-implementation of the navigation APIs in https://github.com/amannn/next-intl/pull/1316 in order to address some long-standing issues. I think the issue you've reported can be solved too as part of that, I've put this on my list!

This is currently my top priority, I'll keep you updated!

amannn:
The feedback raised in this issue has been addressed as part of an upcoming `createNavigation` function (currently available as a prerelease here: https://github.com/amannn/next-intl/pull/1391). In case you decide to give it a spin, let me know if everything works as expected for you!

Also, thanks again for the great feedback here! 🙏❤️

DevOsaWebsite:
HI! 

```
next: 14.2.4
next-intl: 3.21.0-canary.0
```

When use the Link component from routing, for links to the default locale you expect an href like 
```
domain.com
```
but you get 
```
domain.com/defaultLocale
```
DOM render 
```
<a hreflang="en" href="/en/tools"> but expect href="/tools"
```

This does not interfere with user navigation (307 redirect works well). But it is a problem for SEO.
Because when crawling a site, non-existent pages will end up in the Google index. Thus eating up the crawling budget. Also disrupting internal linking, eating up link weight...

[...]

Issue #1055, opened 2024-05-06
Support `localePrefix` setting per domain


markoleavy:
### Is your feature request related to a problem? Please describe.

It would be really useful being able to configure a `localePrefix` preference per domain, in the domain config objects.

### Describe the solution you'd like

Being able to declare a prop inside the domain object and that overrides the default behavior would fix this issue.
```js
export default createMiddleware({
  defaultLocale: 'en-US',
  locales: ['en-US', 'fr-FR', 'en-GB', 'de-DE', 'it-IT']
  domains: [
    {
      domain: 'www.mywebsite.it',
      defaultLocale: 'it-IT',
      localePrefix: 'as-needed',
      locales: ['it-IT']
    },
    {
      domain: 'www.mywebsite.com',
      defaultLocale: 'en-US',
      localePrefix: 'always',
      locales: ['en-US', 'fr-FR', 'en-GB', 'de-DE']
    }
  ]
});
```

### Describe alternatives you've considered

As explained in discussion #1030, at the moment is possible to use an empty string as default locale to get kind of a "always" behavior while "as-needed" is selected in main config.
This is not a documented feature, and might break in the future.

See also: https://github.com/amannn/next-intl/issues/1055#issuecomment-2147830453

amannn:
Prefixes can also be customized with https://github.com/amannn/next-intl/issues/653, this should be taken into account when this is being worked on.

amannn:
As a side note, this can currently be realized by running a separate build for each domain and providing a different config for `localePrefix` e.g. based on an environment parameter.

george-roussos:
Hi, I am trying to use this workaround for the time being, but I am having no luck. Specifically, we have some domains that do not have a locale prefix and a general domain which routes based on prefixes. So I tried something like:

```
export const domains: DomainsConfig<(typeof locales)["en-GB"]> = [
  {
    domain: "test.domain.it",
    defaultLocale: "" as "it-IT",
    locales: ["it-IT"],
  },
  {
    domain: "test.domain.fr",
    defaultLocale: "" as "fr-FR",
    locales: ["fr-FR"],
  },
}

export const navigation: any = {
  locales: locales["en"],
  pathnames,
  localePrefix: "as-needed",
};
```

The `domains` array only includes the special domains. Am I doing something wrong?

amannn:
[...]

PR #1391, opened 2024-10-01
feat: `next-intl@3.22`

amannn:
`next-intl@3.22` is around the corner and already available as a prerelease:

```
npm install next-intl@canary
```

I'd be grateful if some users of `next-intl` could help to test out this version. I've got an extensive test suite, but it'd be great to get feedback on whether it works in all real-world cases. Please leave a note here if you've decided to try out the prerelease and let me know if everything works as expected!

As this is a minor release, there are no adaptions necessary to your app in order to continue working as before. However, two new APIs have been added that will supersede previously available ones that are now deprecated—see the linked PRs below for details. 

**Improvements**
1. https://github.com/amannn/next-intl/pull/1316
2. https://github.com/amannn/next-intl/pull/1383
3. https://github.com/amannn/next-intl/pull/1414
4. https://github.com/amannn/next-intl/pull/1389
5. https://github.com/amannn/next-intl/pull/1411
6. https://github.com/amannn/next-intl/pull/1437

**Next.js 15 compatibility**

Adopting `createNavigation` and `await requestLocale` will get your app in shape for Next.js 15 to work without any warnings (see https://github.com/amannn/next-intl/issues/1375).

**Related issues**

Fixes https://github.com/amannn/next-intl/issues/444
Fixes https://github.com/amannn/next-intl/issues/1335
Resolves https://github.com/amannn/next-intl/issues/454
Resolves https://github.com/amannn/next-intl/issues/1268
Resolves https://github.com/amannn/next-intl/issues/1055
Addresses https://github.com/amannn/next-intl/discussions/785


ScreamZ:
Hey @amannn! As promised here is the first set of feedback, I haven't reviewed your changes in terms of code yet, so I might be missing some implementation details. Once again, sorry for the big block of text!

Here are a few pieces of feedback

First I like the new design and the way you listen to the community :) I'm happy that I can now use various API without being within the middleware!

Just migrated my `not-found.tsx` root

<details>
<summary>➡️ from this</summary>

```tsx
import { validateLocaleOrFallback } from "i18n/validate-locale-or-fallback";
import { getTranslations } from "next-intl/server";

export default async function GlobalNotFound() {
  const locale = validateLocaleOrFallback();
  const t = await getTranslations({ locale });

  return (
    <html lang={locale}>
```

</details>

<details>
<summary>➡️ to  this</summary>

```tsx
import { getLocale, getTranslations } from "next-intl/server";

export default async function GlobalNotFound() {
  const [locale, t] = await Promise.all([getLocale(), getTranslations()]);

  return (
    <html lang={locale}>
```

</details>

And this will also save a lot of other files. I haven't finished migrating things to test the new behavior, I'll keep this updated.

> [!NOTE]
> For the rest of this post, you'll see a lot of calls to my function `validateLocaleOrFallback(locale?: string): AvailableLocales`
>  This function is mainly used to validate the locale or get it from somewhere else than URL. I'm pretty sure this is somehow what you do under the hood, wich leads me to a specific point below (keep reading ahah)
> <details>
> <summary>➡️ Here is the implementation</summary>
> 
> ```tsx
> import { cookies, headers } from "next/headers";
> import { AvailableLocales } from "./locales";
> 
> /**
>  * Ensure locale or fallback if not provided.
>  * A function to use most of time out of next-intl router context, as any function like `getLocale` will trigger a `notFound` invocation and cause unexpected behavior.
>  */
> export function validateLocaleOrFallback(locale?: string): AvailableLocales {
>   if (!locale || !AvailableLocales.includes(locale as AvailableLocales)) {
>     const cookiesList = cookies();
> 
>     locale = cookiesList.get("NEXT_LOCALE")?.value;
> 
>     if (locale) return locale as AvailableLocales;
> 
>     const headersList = headers();
> 
>     // Can be Accept-Language: fr-CH, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5 on safari worst case scenario
>     locale = headersList.get("accept-language")?.split(",")[0]; // fr-CH, fr;q=0.9 => fr-CH
>     locale = locale?.split("-")[0].toLocaleLowerCase(); // fr-CH => fr
> 
>     // Check if exists and valid (also stripe *)
>     if (locale && AvailableLocales.includes(locale as AvailableLocales)) return locale as AvailableLocales;
> 
>     return AvailableLocales[0]; // Final fallback
>   }
> 
>   return locale as AvailableLocales;
> }
> ```
> 
> </details>

## Add precision whether `requestLocale` is not required to be awaited. 

Also, I would rather pass it as a function like `guessLocale`. This way you can lazy load it. If you don't need it, it won't trigger any locale extraction. What do you think? Or are you already guessing values lazily with a promise-compliant object that triggers in `.then`?

## `notFound` in layout

Should I rather do the notFound in metadata or in the page render ?

Below my old implementation still (I'll update this post over time)

```tsx
import { UserNavTracker } from "@modules/navigation/components/UserNavTracker";
import { validateLocaleOrFallback } from "i18n/validate-locale-or-fallback";
import { Metadata } from "next";
import { getTranslations } from "next-intl/server";

type LocaleLayoutProps = Readonly<{
  children: React.ReactNode;
  params: { locale: string };
}>;

export async function generateMetadata({ params: { locale } }: LocaleLayoutProps): Promise<Metadata> {
  locale = validateLocaleOrFallback(locale); // I'm using it here ! hehe
  const t = await getTranslations({ locale, namespace: "SEO" });

  return {
    title: t("common.title"),
    description: t("common.description"),
    keywords: t("common.keywords"),
  };
}

export default async function LocaleLayout({ children }: LocaleLayoutProps) {
  return (
    <>
      <UserNavTracker />
      {children}
    </>
  );
}
```

## Guessing value

It's the topic about my `validateLocaleOrFallback` function. The naming is not best I think I should work on it to something like `guessBestLocale`.

Do you have the equivalent somewhere or can we consider adding this as an helper somewhere.

> Why is that?

Because sometimes, especially with the current authjs beta, you cannot always relies on URL. Or sometimes you want non-i18n webhooks, like for oAuth redirection.

Using this, we can guess the locale, as a fallback on something else.



**BEFORE**
```tsx
import { getRequestConfig } from "next-intl/server";
import { ArrayValues } from "type-fest";
import { AvailableLocales } from "./locales";

export default getRequestConfig(async ({ requestLocale }) => {
  let locale = await requestLocale;

  // Validate that the incoming `locale` parameter is valid
  if (!AvailableLocales.includes(locale as ArrayValues<typeof AvailableLocales>)) {
    // I WANT FALLBACK ON Cookie, then Accept language header, then default to the first locale
    locale = AvailableLocales[0];
  }

  return {
    locale,
    messages: (await import(`./messages/${locale}.json`)).default,
  };
});
```

**AFTER**
```tsx
import { getRequestConfig } from "next-intl/server";
import { validateLocaleOrFallback } from "./validate-locale-or-fallback";

export default getRequestConfig(async ({ requestLocale }) => {
  let locale = await requestLocale;

  // Validate that the incoming `locale` parameter is valid
  locale = validateLocaleOrFallback(locale);

  return {
    locale,
    messages: (await import(`./messages/${locale}.json`)).default,
  };
});
```

Here are my first check, I'll keep posting (this is a big markdown post, I should have split)

amannn:
@ScreamZ Thanks again for the great feedback, and even more so for proposing the idea with the fallback locale in the first place! This came just at the right time since we have to make a switch towards an async API anyway for Next.js 15.

> `Promise.all`

A side note, but regarding this:

```tsx
const [locale, t] = await Promise.all([getLocale(), getTranslations()]);
```

While generally a good idea for API calls, there's no performance difference here. You can absolutely leave it like this to improve readability if you prefer:

```tsx
const t = await getTranslations();
```

> I keep my time casting the locale.

Which cases do you have practically? But yes, I also made a note about this recently: https://github.com/amannn/next-intl/issues/1377.

> Add precision whether `requestLocale` is not required to be awaited.

Hmm, not sure if I follow. It's the same kind of "getter" that e.g. `await params` is in Next.js 15. I.e., if you need to value, then await it, if you don't, then you don't.

There is no "guessing" happening here btw. However the following priorities are taken into account:
1. Overrides like `getTranslations({locale: …})`
2. A static locale provided via `unstable_setRequestLocale`
3. A value that was matched by the middleware (which in turn is based on a list of priorities)

> Should I rather do the notFound in metadata or in the page render ?

Either is fine and will work. While Next.js is expected to call these in parallel, you could try to investigate with some logging if one or the other is reliably called first. In case you have some data fetching, that could help to render the 404 page slightly faster if you abort the render before that kicks in.

> Unable to find next-intl locale because the middleware didn't run on this request.

Is that after the update? Can you give an example of a case where you run into this? I'm also a bit curious since the error message is slightly different now. Since you're always returning a locale from `getRequestConfig` this should honestly not happen _at all_ now. Probably worth digging a bit further into this if you can diagnose a case where this happens.

> Guessing value

I'd recommend not "guessing" the locale. The locale should be a strict contract, with every page and every route that your application provides having a clear assumption of if it's called with a locale or not. For pages, it's the middleware's job to reliably provide this. For Route Handlers, my recommendation is still to provide it explicitly, e.g. via a `locale` search param. If it's still missing somewhere, the new fallback locale can be used, but this should be a conscious decision (e.g. a global language selection page at `/`, or a global 404 page).

Hope this helps, let me know if you have more questions!

ScreamZ:
New use case as I replaced next-auth with custom lucia-auth + artic code.

For OAUth callback, I'm using a non-localized endpoint, this helps not having to add a new locale to authorized locale whitelist in the oAuth provider.

Therefore I would like to access my custom `redirect` to redirect to the proper page in case of oAuth error (localized)

Before 3.22 this wasn't possible as you cannot pass the locale yourself to redirect function. I'll try now with 3.22 canary if this correctly works with new capabilities.

ScreamZ:
@amannn  I can confirm that redirect (from next intl) doesn't care about what I'm returning from getRequestConfig.

```ts
import { getRequestConfig } from "next-intl/server";
import { validateLocaleOrFallback } from "./validate-locale-or-fallback";

export default getRequestConfig(async ({ requestLocale }) => {
  const locale = validateLocaleOrFallback(await requestLocale);

  console.log("getRequestConfig", locale);
  return {
    locale,
    messages: (await import(`./messages/${locale}.json`)).default,
  };
});
```

```ts
export const config = {
  // Match only internationalized pathnames, this cannot be used with code so anytime a locale is used, the locale must be hardcoded here
  matcher: ["/", `/(en|fr|de|es|it)/:path*`, `/slice-simulator`],
};
```

this is voluntarily out of the middleware as i don't want auto-adding /fr or other to route.

```ts
// In my route handler path app/(not-localized)/auth/callback/[provider]/route.ts
    const t = await getTranslations();

      // Report not translated error code
      if (!t.has(`Auth.error.${code}` as any)) {
        await reportLogToDiscord("Authentication", "Not translated error", {
          "error.json": { e, code, provider },
        });
      }

      return redirect({ pathname: "/authentication/sign-in", query: { error: code } });
```

with logs

```
getRequestConfig fr
🪵 [Discord Log] Authentication: Not translated error

Unable to find `next-intl` locale because the middleware didn't run on this request. See https://next-intl-docs.vercel.app/docs/routing/middleware#unable-to-find-locale. The `notFound()` function will be called as a result.

 GET /auth/callback/google?state=1lpy2j_T0p9m9mrO8iXrqFVzs_TI5Rj
```

amannn:
@ScreamZ Cool to hear that things are progressing with your app! Do I understand correctly based on your edit that you got everything working now? Or is there any issue left to resolve?

Regarding ETA: I'm currently working on a follow-up for https://github.com/amannn/next-intl/pull/1414 in https://github.com/amannn/next-intl/pull/1417. Apart from that, there are currently no issues known regarding other features added in this PR. What I am waiting for though is Next.js publishing a new v15 RC, so I can confirm that we're compatible with their latest changes. With Next.js Conf coming up next week, I think that should be soon!

ScreamZ:
> @ScreamZ Cool to hear that things are progressing with your app! Do I understand correctly based on your edit that you got everything working now? Or is there any issue left to resolve?
> 
> Regarding ETA: I'm currently working on a follow-up for #1414 in #1417. Apart from that, there are currently no issues known regarding other features added in this PR. What I am waiting for though is Next.js publishing a new v15 RC, so I can confirm that we're compatible with their latest changes. With Next.js Conf coming up next week, I think that should be soon!

Everything (so far) is working for now.

Yeah I think they will do some announcement on Next conf

Thanks !

nhducit:
I'm testing out the canary version but still getting this error

```
 In route /[locale] a header property was accessed directly with `headers().get('X-NEXT-INTL-LOCALE')`. `headers()` should be awaited before using its value. Learn more: https://nextjs.org/docs/messages/sync-dynamic-apis
 ```

version
```
"next-intl": "3.22.0-canary.0",
"next": "v15.0.0-canary.191",
```

PR #1397, opened 2024-10-02
feat: Add example-app-router-with-next-auth-without-i18n-routing

teplostanski:
Adding Authentication with `next-auth` to the `example-app-router-without-i18n-routing`

This pull request expands the functionality of the existing example by adding user authentication using the `next-auth` library. The application now supports authentication and allows you to protect routes that require authorization.


amannn:
Hey @teplostanski, thank you so much for proposing this example—looks like you spent quite some time on it! ❤️ 

That being said, the existing `example-app-router-without-i18n-routing` is intended to stay without an actual auth service integration to serve more as a routing inspiration. We also have another example `example-app-router-next-auth` that demonstrates a demo setup with Auth.js.

As examples take significant time to maintain, I'd prefer to not add another example that showcases a demonstration of both examples combined—I believe the important patterns should be covered in either example.

Have you considered though to publish your example as a separate repo under your username? I know that quite a few users are looking for a Auth.js v5 example (see https://github.com/amannn/next-intl/issues/596). In this repo I'm currently waiting for the v5 stable version to drop, but in case you're interesting in upgrading your example to v5, you could link to it in the thread at https://github.com/amannn/next-intl/issues/596 for example.

Hope this makes sense, keep up the good work! 🙌

amannn:
Closing this due to inactivity …

PR #1399, opened 2024-10-03
feat: Add `t.has` to check whether a given message exists

### LINKED ISSUE #88: Optional translations
**Is your feature request related to a problem? Please describe.**

It would be good if some translations could be marked as optional, so no error is reported.

**Describe the solution you'd like**

Either an API to ask if a translation is available to conditionally render it, or an API to suppress the error in this case.

Probably `t.has(…)` would be the best option here as it can be used to check for translations before using either `t`, `t.rich`, `t.raw` etc.

**Describe alternatives you've considered**

Adding empty strings in the messages if the label is not relevant for a particular scenario (i.e. `"message": ""`).

If you're using [the VSCode integration for `next-intl` via i18n Ally](https://next-intl-docs.vercel.app/docs/workflows/vscode-integration), you can add empty messages for new keys via `"i18n-ally.keepFulfilled": true`.

I'm currently unsure if `next-intl` needs built-in support for this, or if it can be solved with tooling. I guess the question is: "Is this translation missing currently or is it not intended at all?".

This could also be solved via a combination of [`useMessages`](https://next-intl-docs.vercel.app/docs/usage/configuration#retrieve-global-configuration) and `lodash/get`, e.g. in a custom hook.

tholander:
```js
import {useTranslations} from 'next-intl';

function AboutPage() {
  const t = useTranslations('AboutPage');

  if (!t.has('title')) {
    notFound()
  }

  // ...
}
```

Closes #88


amannn (commented):
@tholander This looks fantastic, thanks a lot for taking the initiative on this! 💯 

During the review I've added a commit to add a few more tests, hope that's ok for you.

I'll do another review tomorrow, but I think this is good to go!

blazingh:
thanks a lot for this.

PR #1089, opened 2024-05-24
docs: Update docs and examples to Next.js 15

amannn:
Fixes https://github.com/amannn/next-intl/issues/1587

Note that https://github.com/amannn/next-intl/issues/1669 is still relevant, therefore you might want to use a [workaround](https://github.com/amannn/next-intl/issues/1669#issuecomment-2605280931) for the time being.

PR #1401, opened 2024-10-04
docs: update with-i18n-routing.mdx with nextjs 15 `params` change

ammubhave:
Update app router with-i18n-routing with a nextjs 15 `params` change. In nextjs 15, the `params` property is a promise. Although direct access is still supported, it is deprecated.

If params is accessed directly, and not awaited then we get the following warning
```
In route /[locale] a param property was accessed directly with `params.locale`. `params` is now a Promise and should be awaited before accessing properties of the underlying params object. In this version of Next.js direct access to param properties is still supported to facilitate migration but in a future version you will be required to await `params`. If this use is inside an async function await it. If this use is inside a synchronous function then convert the function to async or await it from outside this function and pass the result in.
```

For more information see https://github.com/vercel/next.js/pull/68812

### DIFF
--- a/docs/pages/docs/getting-started/app-router/with-i18n-routing.mdx
+++ b/docs/pages/docs/getting-started/app-router/with-i18n-routing.mdx
@@ -186,11 +186,12 @@ import {getMessages} from 'next-intl/server';
 
 export default async function LocaleLayout({
   children,
-  params: {locale}
+  params
 }: {
   children: React.ReactNode;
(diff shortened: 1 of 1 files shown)


amannn:
Thanks for the proposed update! There will be a few more updates necessary to the docs, but let's wait for now until the stable release of Next.js 15 lands. I'll close this as a duplicate of https://github.com/amannn/next-intl/pull/1089.

PR #1413, opened 2024-10-09
feat!: Inherit context between providers

amannn:
If you have nested providers, previously only the configuration of the innermost one would be applied.

With this change, configuration is now passed from one provider to the next, while allowing to override individual props.

**BREAKING CHANGE:** There's a very rare chance that this change affects your app, but in case you've previously relied on providers not inheriting from each other, you now have to reset props manually in case you want to retain the prev. behavior.

### DIFF
--- a/docs/pages/docs/usage/configuration.mdx
+++ b/docs/pages/docs/usage/configuration.mdx
@@ -134,9 +134,6 @@ import IntlErrorHandlingProvider from './IntlErrorHandlingProvider';
 
 export default async function RootLayout({children}) {
   const locale = await getLocale();
-
-  // Providing all messages to the client
-  // side is the easiest way to get started
   const messages = await getMessages();
 
   return (
--- a/docs/pages/docs/usage/configuration.mdx
+++ b/docs/pages/docs/usage/configuration.mdx
@@ -99,41 +99,28 @@ In contrast, these props can be provided as necessary:
 2. `defaultTranslationValues`
 3. `onError` and `getMessageFallback`
 
+Additionally, nested instances of `NextIntlClientProvider` will inherit configuration from their respective ancestors. Note however that individual props are treated as atomic, therefore e.g. `messages` need to be merged manually—if necessary.
+
 <Details id="nextintlclientprovider-non-serializable-props">
 <summary>How can I provide non-serializable props like `onError` to `NextIntlClientProvider`?</summary>
 
 React limits the types of props that can be passed to Client Components to the ones that are [serializable](https://react.dev/reference/rsc/use-client#serializable-types). Since `onError`, `getMessageFallback` and `defaultTranslationValues` can receive functions, these configuration options can't be automatically inherited by the client side.
 
-In order to define these values, you can wrap `NextIntlClientProvider` with another component that is marked with `'use client'` and defines the relevant props:
+In order to define these values on the client side, you can add a provider that defines these props:
 
-```tsx filename="IntlProvider.tsx"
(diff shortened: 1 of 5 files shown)


packages/use-intl/src/react/IntlProvider.tsx
```diff
@@ -47,14 +49,16 @@ export default function IntlProvider({
...
+        locale, // (required by provider)
+        defaultTranslationValues:
+          defaultTranslationValues || prevContext?.defaultTranslationValues,
+        formats: formats || prevContext?.formats,
+        getMessageFallback:
+          getMessageFallback || prevContext?.getMessageFallback,
+        messages: messages || prevContext?.messages,
+        now: now || prevContext?.now,
+        onError: onError || prevContext?.onError,
+        timeZone: timeZone || prevContext?.timeZone
```
amannn, line 61: Doesn't minify terribly well, but can't really think of a better solution …

docs/pages/docs/usage/configuration.mdx
```diff
@@ -99,69 +99,51 @@ In contrast, these props can be provided as necessary:
...
+import {getLocale, getMessages} from 'next-intl/server';
+import IntlErrorHandlingProvider from './IntlErrorHandlingProvider';
 
 export default async function RootLayout({children}) {
   const locale = await getLocale();
-  const now = await getNow();
-  const timeZone = await getTimeZone();
+
+  // Providing all messages to the client
+  // side is the easiest way to get started
```
amannn, line 139:
```suggestion
```

dmitrc:
@amannn, thanks for your great work as always!

I was wondering if it also makes sense to optionally allow the user to merge the existing and current messages as well, eg to support the nested providers where you provide essential client messages in the root layout, and then component-specific messages later down the tree?

amannn:
@dmitrc Thank you for the kind words and happy new year!

Merging such messages is certainly an interesting use case, yes. Currently, you could achieve this manually in userland.

Ideally, with https://github.com/amannn/next-intl/issues/1, relevant messages for Client Components would be picked automatically—not requiring the developer to think in terms of "essential" client messages and "component-specific" ones.

https://github.com/amannn/next-intl/issues/1 and https://github.com/amannn/next-intl/issues/962 are the two main features I'd like to explore next for `next-intl`. If this works out well, I think we might not have to add merging of messages to `next-intl`. Due to this, I think for the time being I'd not add this behavior to `next-intl`.

Does that sound right to you?

jizhi0v0:
@amannn 
I have been using next-intl for a while and it works great!

In the v3 version, I tried to use the server component to pass messages to the provider in different pages under the same layout, but the client did not get the data correctly through useTranslations. Using raw can get the complete data. Does this submission solve this problem? Thanks!

dmitrc:
> Due to this, I think for the time being I'd not add this behavior to `next-intl`.
> 
> Does that sound right to you?

Yup, this makes complete sense, and I am incredibly excited about the improvements that solving #1 would bring!

The only challenge in implementing it in userland is that it requires some opinionated setup:

* We can't obtain the whole `IntlContext` as it is not exposed, so we have to do it piecewise
* We can obtain the messages via `useMessages` to merge them, but the hook will throw if it hasn't been set earlier
* Since wrapping hooks in `try/catch` is a no-no, we have to ensure to set `<NextIntlClientProvider messages={[]}>` at the root
* Subsequently our own `<NestedNextIntlClientProvider ...>` can be used to merge the new messages with prior ones

Not too big of a deal, but the second bullet caught me by surprise a little :)

PR #1420, opened 2024-10-14
fix: support like 'enAU' 'en_AU' locales in IntlMessageFormat

apades:
If locales params like camel case 'enUS', undline 'en_US', it will throw err when i use dynamic values template, i find that `intl-messageformat` only support kebab-case locale.

Error like this
```txt
IntlError: INVALID_MESSAGE: Incorrect locale information provided
    at getFallbackFromErrorAndNotify (webpack-internal:///(rsc)/../../packages/use-intl/dist/development/createFormatter-DbsypjTl.js:170:23)
    at translateBaseFn (webpack-internal:///(rsc)/../../packages/use-intl/dist/development/createFormatter-DbsypjTl.js:236:20)
    at translateFn (webpack-internal:///(rsc)/../../packages/use-intl/dist/development/createFormatter-DbsypjTl.js:258:24)
```

So i try to fixed it, many repositories have these naming schemes, it very distressed when i using dynamic values template

### DIFF
--- a/examples/example-app-router/messages/enAU.json
+++ b/examples/example-app-router/messages/enAU.json
@@ -1,0 +1,46 @@
+{
+  "Error": {
+    "description": "<p>We've unfortunately encountered an error.</p><p>You can try to <retry>reload the page</retry> you were visiting.</p>",
+    "title": "Something went wrong!"
+  },
+  "IndexPage": {
+    "description": "This is a basic example that demonstrates the usage of <code>next-intl</code> with the Next.js App Router. Try changing the locale in the top right corner and see how the content changes.",
+    "title": "next-intl example"
+  },
+  "LocaleLayout": {
+    "title": "next-intl example"
+  },
+  "LocaleSwitcher": {
+    "label": "Change language",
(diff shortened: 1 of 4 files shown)


amannn:
Thanks for considering a PR! However, `next-intl` aligns with the [`Intl.Locale`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale) API which doesn't intend to support these alternative variants.

If your use case is showing these variants on the pathname, you can consider using [custom prefixes](https://next-intl-docs.vercel.app/docs/routing#locale-prefix-custom).

Hope that helps!

apades:
[custom prefixes](https://next-intl-docs.vercel.app/docs/routing#locale-prefix-custom) didn't resolve my problem, but I get your point. Maybe you can alert some `Intl.Locale` error about some alternative variants, it will decrease some confusion about “It's working, but why I can not use dynamic tempalte”, I have seen this related on reddit and stackoverflow😂

amannn:
Oh you mean that we could include a warning if an invalid locale is received? That could be a good idea, yes—I made a note about it in https://github.com/amannn/next-intl/issues/779. Thanks for raising this!

Issue #1441, opened 2024-10-22
Dependency issue after upgrading to Nextjs 15 and Next-Intl 3.22


levipadre:
### Description

Hi,

I get this error after upgrading to nextjs 15:

```
npm error code ERESOLVE
npm error ERESOLVE unable to resolve dependency tree
npm error
npm error While resolving: ciwf-frontend-application@0.1.0
npm error Found: next@15.0.0
npm error node_modules/next
npm error   next@"^15.0.0" from the root project
npm error
npm error Could not resolve dependency:
npm error peer next@"^10.0.0 || ^11.0.0 || ^12.0.0 || ^13.0.0 || ^14.0.0" from next-intl@3.22.0
npm error node_modules/next-intl
npm error   next-intl@"^3.22.0" from the root project
npm error
npm error Fix the upstream dependency conflict, or retry
npm error this command with --force or --legacy-peer-deps
npm error to accept an incorrect (and potentially broken) dependency resolution.
```

### Verifications

- [X] I've verified that the problem I'm experiencing isn't covered in [the docs](https://next-intl-docs.vercel.app/).
- [X] I've searched for similar, existing issues on GitHub and [Stack Overflow](https://stackoverflow.com/search?q=next-intl).
- [X] I've compared my app to [a working example](https://next-intl-docs.vercel.app/examples) to look for differences.

### Other

```
Operating System:
      Platform: darwin
      Arch: x64
      Version: Darwin Kernel Version 23.6.0: Mon Jul 29 21:13:00 PDT 2024; root:xnu-10063.141.2~1/RELEASE_X86_64
    Binaries:
      Node: 18.20.4
      npm: 10.9.0
      Yarn: 1.22.19
      pnpm: 9.7.0
    Relevant Packages:
      next: 13.5.1
      eslint-config-next: N/A
      react: 18.2.0
      react-dom: 18.2.0
      typescript: 5.2.2
    Next.js Config:
      output: N/A
```

### Reproduction description

Steps to reproduce:
1. Upgrade to nextjs 15
2. upgrade to next-intl 3.22
3. run npm run i
4. See error: above in the description

### Expected behaviour

Running without error

amannn:
This will be addressed in https://github.com/amannn/next-intl/pull/1443 (see the PR for details, maybe you want to wait with updating).

levipadre:
Sorry, thanks!

Issue #1075, opened 2024-05-18
Support for Next.js 15 RC and React 19 RC


Aralf:
### Description

I have taken the example code from `example-app-router-next-auth` that has the With i18n Routing setup; and tried to use latest Next canary (18.3.0-canary.70 at the time of writing this) with React 19 RC in order to test both that and the new React compiler.

There seems to be an issue with React 19, as the application fails to run, even without trying the React compiler.

Without turbo it shows the following error.
![error without turbo](https://github.com/amannn/next-intl/assets/7780231/17216932-b918-4e5b-8e17-84fbf94b60c0)

With turbo, instead shows this error.
![error with turbo](https://github.com/amannn/next-intl/assets/7780231/dfed643c-0e62-4a5c-81f5-3fbb12f27101)

I have made another branch where I have follow the documentation for `Without i18n routing`. This removes the `[locale]`, `navigation.ts` and use of next-intl middleware on `middleware.ts`. With these changes the application runs perfectly fine both with/without turbo and the react compiler.

### Verifications

- [X] I've verified that the problem I'm experiencing isn't covered in [the docs](https://next-intl-docs.vercel.app/).
- [X] I've searched for similar, existing issues on GitHub and [Stack Overflow](https://stackoverflow.com/search?q=next-intl).
- [X] I've compared my app to [a working example](https://next-intl-docs.vercel.app/examples) to look for differences.

### Mandatory reproduction URL

https://github.com/Aralf/next-intl-react-19-compiler

### Reproduction description

Steps to reproduce:
1. Run `next dev` with or without turbo
2. It will not work, in both cases throws different errors.
3. There is no need for trying React compiler as it already fails.
4. Change to branch `without-i18n-routing` 
5. React compiler is already enabled on next.config.mjs
6. Run `next dev` with or without turbo
7. It works woks for all cases

### Expected behaviour

It should work for both the With i18n routing and the Without i18n routing setups.

amannn:
Hmm, that's interesting. The Next.js 14 release is scheduled for next week as far as I know, maybe worth waiting until then. It might be that they're still ironing out some details.

heresyrj:
can confirm this issue. 
reproduced the same error on my project.

Gawdfrey:
Able to reproduce it here as well.
As mentioned by @amannn, probably worth waiting until [next 15 is launched this week](https://x.com/feedthejim/status/1791246068971381246) until further investigation.

singuerinc:
[...]

Issue #1442, opened 2024-10-22
Next.js 15 breaks alternate links


amannn:
Recent canary versions of Next.js 15 as well as the stable version that was released yesterday break [alternate links](https://next-intl-docs.vercel.app/docs/routing#alternate-links) if you're using `next/font` or `next/image` with `priority`.

This header is important for apps that use i18n routing and the `alternateLinks` feature (which is turned on by default). Failing to return this header can result in content being marked as duplicate by search engines.

Upstream bug report: https://github.com/vercel/next.js/issues/69000#issuecomment-2416459189

I'm currently waiting for this to be resolved upstream before updating the examples and the bug repro to use Next.js 15.

As a workaround, you can use a [sitemap](https://next-intl-docs.vercel.app/docs/environments/actions-metadata-route-handlers#sitemap) to instruct search engines about multiple variants of your pages.

Wiibleyde:
Hello, I think it's a part of the problem enounced here (if it's not i'm sorryyy), I have this error :
```
Error: Route "/[locale]" used `headers().get('X-NEXT-INTL-LOCALE')`. `headers()` should be awaited before using its value. Learn more: https://nextjs.org/docs/messages/sync-dynamic-apis
    at createHeadersAccessError (turbopack://[project]/node_modules/next/src/server/request/headers.ts:487:9)
    at getMessage (turbopack://[project]/node_modules/next/src/server/create-deduped-by-callsite-server-error-logger.ts:46:20)
    at warnForSyncAccess (turbopack://[project]/node_modules/next/src/server/request/headers.ts:473:2)
    at syncIODev (turbopack://[project]/node_modules/next/src/server/request/headers.ts:376:8)
    at <unknown> (turbopack://[project]/node_modules/next-intl/dist/esm/server/react-server/RequestLocaleLegacy.js:1:278)
    at apply (file:///home/wiibleyde/Alternance/lex-port-cartel/node_modules/next/dist/compiled/next-server/dist/compiled/react/cjs/react.react-server.development.js:888:26)
    at s (turbopack://[project]/node_modules/next-intl/dist/esm/server/react-server/RequestLocaleLegacy.js:1:996)
    at get locale [as locale] (turbopack://[project]/node_modules/next-intl/dist/esm/server/react-server/getConfig.js:1:874)
    at <unknown> (turbopack://[project]/src/i18n/request.ts:5:41)
[...]

PR #1443, opened 2024-10-22
feat: Add Next.js 15 to peer dependencies

amannn:
`next-intl` generally works with Next.js 15.

However, in case you're using i18n routing and the alternate links feature, you might want to wait with upgrading until an upstream bug is resolved in Next.js: https://github.com/amannn/next-intl/issues/1442. Due to this, I'll delay updating the docs and examples in this repo (https://github.com/amannn/next-intl/pull/1089).

Fixes https://github.com/amannn/next-intl/issues/1441
Fixes https://github.com/amannn/next-intl/issues/1075


packages/next-intl/package.json
```diff
@@ -91,7 +91,7 @@
     "use-intl": "workspace:^"
   },
   "peerDependencies": {
-    "next": "^10.0.0 || ^11.0.0 || ^12.0.0 || ^13.0.0 || ^14.0.0",
+    "next": "^10.0.0 || ^11.0.0 || ^12.0.0 || ^13.0.0 || ^14.0.0 || ^15.0.0",
     "react": "^16.8.0 || ^17.0.0 || ^18.0.0"
```
theoludwig, line 95: Thank you for this PR! :smile:

Is it also possible to allow React.js v19 in peerDependencies?

```suggestion
    "react": "^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0 || ^19.0.0-rc"
```

amannn: Next.js 15 [vendors](https://nextjs.org/blog/next-15#react-19) the React 19 RC anyway, you implicitly get React 19 even if React 18 is installed 🙂.

I'll add official support for React 19 once it hits stable and is generally available, hope that makes sense!

PR #1470, opened 2024-10-25
feat!: Make package ESM-only, target modern browsers, use modern JSX transform

amannn:
**Changes**

1. `next-intl` as well as its base library `use-intl` are now ESM-only. With this, the library is now ~7% smaller and provides full support for tree shaking. Most tools support ESM out-of-the-box, but some might require additional configuration (e.g. Jest might benefit from [`next/jest`](https://nextjs.org/docs/app/building-your-application/testing/jest#manual-setup)). The only exception is `next-intl/plugin` which is provided both as ESM as well as CommonJS due to `next.config.js` still being popular. 
2. Syntax is compiled down to the Browserslist [`defaults`](https://github.com/browserslist/browserslist/blob/065c69b0e1e6ad679263d03e22c605830bbc52e5/index.js#L460) query, which is a shortcut for `> 0.5%, last 2 versions, Firefox ESR, not dead`. The `defaults` query was thoroughly designed by the Browserslist community and is considered a reasonable default for modern apps. If you target outdated browsers, you can use [`transpilePackages`](https://nextjs.org/docs/app/api-reference/next-config-js/transpilePackages) in combination with a [`browserslist`](https://nextjs.org/docs/architecture/supported-browsers) config in Next.js to compile `next-intl` for older browsers.
3. The minimum React version is 17 now (which supports the modern JSX transform)

### DIFF
--- a/packages/next-intl/src/navigation/createLocalizedPathnamesNavigation.test.tsx
+++ b/packages/next-intl/src/navigation/createLocalizedPathnamesNavigation.test.tsx
@@ -6,7 +6,6 @@ import {
   usePathname as useNextPathname,
   useParams
 } from 'next/navigation';
-import React from 'react';
 import {renderToString} from 'react-dom/server';
 import {beforeEach, describe, expect, it, vi} from 'vitest';
 import {Pathnames, defineRouting} from '../routing.tsx';
--- a/packages/next-intl/src/navigation/createLocalizedPathnamesNavigation.test.tsx
+++ b/packages/next-intl/src/navigation/createLocalizedPathnamesNavigation.test.tsx
@@ -5,7 +5,7 @@ import {
   redirect as nextRedirect,
   usePathname as useNextPathname,
   useParams
-} from 'next/navigation';
+} from 'next/navigation.js';
 import {renderToString} from 'react-dom/server';
 import {beforeEach, describe, expect, it, vi} from 'vitest';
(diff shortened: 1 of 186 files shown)


packages/next-intl/src/navigation/createLocalizedPathnamesNavigation.test.tsx
```diff
@@ -5,19 +5,18 @@ import {
   redirect as nextRedirect,
   usePathname as useNextPathname,
   useParams
-} from 'next/navigation';
-import React from 'react';
+} from 'next/navigation.js';
```
amannn, line 8: See also: https://github.com/vercel/next.js/issues/46078

packages/next-intl/src/shared/use.tsx
```diff
@@ -0,0 +1,13 @@
...
+  // @ts-expect-error -- Ooof, Next.js doesn't make this easy.
+  // `use` is only available in React 19 canary, but we can
+  // use it in Next.js already as Next.js "vendors" a fixed
+  // version of React. However, if we'd simply put `use` in
+  // ESM code, then the build doesn't work since React does
+  // not export `use` officially. Therefore, we have to use
+  // something that is not statically analyzable. Once React
+  // 19 is out, we can remove this in the next major version.
+  return react['use'.trim()](value);
+}
```
amannn, line 13: React 19 can't come soon enough.

Issue #1452, opened 2024-10-23
Type `useMessages()` hook using `IntlMessages`


astralarya:
### Is your feature request related to a problem? Please describe.

`useMessages()` hook does not get proper typing after declaring `IntlMessages`.

For example:
```typescript
declare global {
  // Use type safe message keys with `next-intl`
  type IntlMessages = {
    foo: {
      bar: string,
      baz: string,
    }
  }
}

function MyComponent() {
  const messages = useMessages();  // expect this to have type IntlMessages
}
```

### Describe the solution you'd like

I would like the `useMessages()` hook to respect the type provided by IntlMessages

### Describe alternatives you've considered

I've considered casting in place every time I invoke the hook, but this is error prone.

amannn:
Sounds good, thanks for the report! This will likely make it into the upcoming v4 release: https://github.com/amannn/next-intl/pull/1489

amannn:
This has been resolved in the [next-intl 4.0 beta](https://next-intl.dev/blog/next-intl-4-0)!

patgpt:
I'm unable to find this on npm 

```bash 
bun add next-intl@v4-beta
[1.12ms] ".env"
bun add v1.1.42 (50eec002)
error: No version matching "v4-beta" found for specifier "next-intl" (but package exists)
error: next-intl@v4-beta failed to resolve
```

amannn:
`v4-beta` is a tag. Not sure if there's anything peculiar with bun, you can also get the exact version from [npm versions](https://www.npmjs.com/package/next-intl?activeTab=versions).

PR #1489, opened 2024-10-30
feat: Return type-safe `IntlMessages` from `useMessages` & `getMessages`

amannn:
If consumers use type-safe `IntlMessages`, these will now be used for `useMessages` and `getMessages`.

Fixes https://github.com/amannn/next-intl/issues/1452

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   build:
     name: Build, lint, and test
-    runs-on: macos-latest
+    runs-on: macos-15
     env:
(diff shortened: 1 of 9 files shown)


.github/workflows/main.yml
```diff
@@ -7,7 +7,7 @@ on:
 jobs:
   build:
     name: Build, lint, and test
-    runs-on: ubuntu-latest
+    runs-on: macos-15
```
amannn, line 10: Potential alternative if this doesn't work well: https://playwright.dev/docs/ci#workers

Issue #410, opened 2023-07-20
Type-safe arguments


amannn:
### Is your feature request related to a problem? Please describe.

next-intl provides opt-in [strict typing for messages](https://next-intl-docs.vercel.app/docs/workflows/typescript). With some additional TypeScript voodoo, it might be possible to support type checking for arguments that need to be passed to `t` calls.

Resources:
- https://twitter.com/trashh_dev/status/1558209179071877120/photo/1
- https://github.com/schummar/schummar-translate/blob/master/src/extractICU.ts
- https://github.com/QuiiBz/next-international/tree/main/packages/international-types#type-safe-params
  - Limitations: Plurals can't start with a non-expression, plurals can't end with a non-expression, select is not supported, rich text not supported (could be augmented though)

### Describe the solution you'd like

```js
{
  "hello": "Hello {name}"
}
```

```tsx
t('hello') // Should be an error
```

We have to evaluate though what it takes to support everything from ICU and also rich text.

Also we should be sure about the performance implications of this (see also https://github.com/amannn/next-intl/issues/792).

**Open question:** Since we support `defaultTranslationValues`, type-safety would only cover types (ie. accept only strings, numbers, etc), but not if an argument is provided or not (it could come from the defaults). This begs the question if `defaultTranslationValues` are really a good idea or if we should get rid of this. https://github.com/amannn/next-intl/issues/611 discusses a separate issue but also comes to the conclusion that `defaultTranslationValues` might be helpful to remove—an alternative is suggested there too.

### Describe alternatives you've considered

Not introducing types for this

DemianParkhomenko:
It would be an extremely useful feature that speeds up the development)

TommySorensen:
@amannn I guess this will also help with having different objects of translations for each locale? Currently i load my translations async, but one locale might have more and different translations then the rest.

matannahmani:
any updates on this one?

loczek:
You can add a `global.d.ts` file in the root of your project and it should work

```ts
type Messages = typeof import("./messages/en.json");
declare interface IntlMessages extends Messages {}
```

DemianParkhomenko:
> You can add a `global.d.ts` file in the root of your project and it should work
> 
> ```ts
> type Messages = typeof import("./messages/en.json");
> declare interface IntlMessages extends Messages {}
> ```

But it does not check arguments, isn't it? i.e. `{name}` in the example 

```json
{
  "hello": "Hello {name}"
}
```

[...]

PR #1499, opened 2024-11-01
feat: Type-safe ICU arguments

amannn:
```json
{
  "UserProfile": {
    "title": "Hello {firstName}"
  }
}
```

```tsx
function UserProfile({user}) {
  const t = useTranslations('UserProfile');
 
  // ✖️ Missing argument
  t('title');
 
  // ✅ Argument is provided
  t('title', {firstName: user.firstName});
}
```

**Changes**
- ICU arguments can now be validated with TypeScript (however, currently this is [opt-in](https://next-intl-docs-git-feat-type-safe-params-next-intl.vercel.app/docs/workflows/typescript#messages-arguments))
- `undefined` and `null` are no longer accepted as values for calls to `t`—please only provide valid values.

Fixes https://github.com/amannn/next-intl/issues/410


packages/use-intl/src/core/ICUArgs.tsx
```diff
@@ -0,0 +1,181 @@
+// From https://github.com/schummar/schummar-translate
```
amannn, line 1: See https://github.com/schummar/schummar-translate/issues/28

PR #1509, opened 2024-11-04
feat: Add type exports to enable `declaration: true` in `tsconfig.json`

osaton:
Re: <https://github.com/amannn/next-intl/discussions/1014>

Fix the type issues encountered when re-exporting types in a client application with `compilerOptions.declaration` set to `true`. This change addresses errors such as:

> The inferred type of 'Link' cannot be named without a reference to '../node_modules/next-intl/dist/types/src/shared/types'. This is likely not portable. A type annotation is necessary.ts(2742)


packages/next-intl/src/navigation/react-client/index.tsx
```diff
@@ -2,6 +2,13 @@ export {default as createSharedPathnamesNavigation} from './createSharedPathname
 export {default as createLocalizedPathnamesNavigation} from './createLocalizedPathnamesNavigation';
 export {default as createNavigation} from './createNavigation';
 
+export type {StrictParams} from '../shared/StrictParams';
+export type {
+  HrefOrHrefWithParams,
+  HrefOrUrlObjectWithParams,
+  QueryParams
+} from '../shared/utils';
```
amannn, line 10: Is there a way to get a list of which types need to be exported? Or was this rather trial & error?

Some of these types should have their names reconsidered in case we have to export them. Ideally, we wouldn't have to. In the past, I think I was able to work around exporting some types via `ReturnType<typeof …>` annotations at some places where I saw declaration errors.

pnpm-workspace.yaml
```diff
@@ -1,4 +1,5 @@
 packages:
-  - "packages/*"
-  - "examples/*"
-  - "docs"
+  - 'packages/*'
+  - 'type-export-test'
```
amannn, line 3: `example-app-router-playground` is in fact already a good example that stress-tests all APIs of `next-intl`. Can we add a `tsconfig.declaration.json` there and add a lint task that calls `tsc` with that config?

That way we don't have to add a completely new app (each example adds some maintenance burden).

amannn (commented):
Super cool, thanks a lot for looking into this!

packages/next-intl/src/navigation/react-client/index.tsx
osaton: Unfortunately, it's just trial & error. At least I don't know of a better way to check them. That said, this approach led to some unnecessary exports which I’ve removed in my latest commit.

amannn: Okay, thanks! I'll also verify the imports again once we can use `example-app-router-playground` to verify that `declaration: true` works.

pnpm-workspace.yaml
osaton: They are used, but not everything is exported. For example `useTimeZone` hook is only used inside components and as they only export `JSX.Element`, TypeScript does not have to try and export the return type of that function. But if you add something like:
```ts
export function useExport() {
  return useTimeZone();
}
```
You'll get a following error:
```

amannn: Awesome, thanks a lot! Just had another look from my side and added a commit with some minor cleanup.

All good from my side, let's see if CI is happy too …

amannn (approved):
Thanks a bunch for your help here @osaton! 👏

PR #1523, opened 2024-11-07
fix: Allow to merge inline formats with global formats in `t`

amannn:
See https://github.com/amannn/next-intl/pull/1523/files#r1832900948

### DIFF
--- a/packages/next-intl/.size-limit.ts
+++ b/packages/next-intl/.size-limit.ts
@@ -15,7 +15,7 @@ const config: SizeLimitConfig = [
   {
     name: "import * from 'next-intl' (react-server, production)",
     path: 'dist/esm/production/index.react-server.js',
-    limit: '14.075 KB'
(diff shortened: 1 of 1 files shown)


packages/use-intl/src/core/createTranslator.test.tsx
```diff
@@ -541,6 +541,29 @@ describe('type safety', () => {
...
+  it('can merge an inline format with global formats', () => {
+    const t = createTranslator({
+      locale: 'en',
+      messages: {label: '{count, number, precise} {count, number, integer}'},
+      formats: {number: {precise: {minimumFractionDigits: 5}}}
+    });
+    expect(
+      t('label', {count: 1.5}, {number: {integer: {minimumFractionDigits: 0}}})
+    ).toBe('1.50000 2');
+  });
```
amannn, line 564: Relevant test for the fix

PR #1531, opened 2024-11-11
experiment: PPR & friends

amannn:
An exploration of how `next-intl` can adapt better for PPR & friends.

**Findings:**
- If params could be read deeply, we can use `dynamicParams = false` to return a 404 for unknown locales (see https://github.com/amannn/next-intl/commit/d1a0308e72d7f5111a7dcc6f34efd22d8ad0e576). This is not a good idea to introduce currently, because a large number of users might not use `generateStaticParams`.

### DIFF
--- a/examples/example-app-router/src/app/[locale]/layout.tsx
+++ b/examples/example-app-router/src/app/[locale]/layout.tsx
@@ -1,5 +1,4 @@
-import {notFound} from 'next/navigation';
-import {Locale, hasLocale} from 'next-intl';
+import {Locale} from 'next-intl';
 import {getTranslations, setRequestLocale} from 'next-intl/server';
 import {ReactNode} from 'react';
 import BaseLayout from '@/components/BaseLayout';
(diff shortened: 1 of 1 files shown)


examples/example-app-router/src/components/BaseLayout.tsx
```diff
@@ -20,10 +20,12 @@ export default async function BaseLayout({children, locale}: Props) {
   return (
     <html className="h-full" lang={locale}>
       <body className={clsx(inter.className, 'flex h-full flex-col')}>
-        <NextIntlClientProvider messages={messages}>
-          <Navigation />
-          {children}
-        </NextIntlClientProvider>
+        <Suspense>
```
amannn, line 23: Should ideally work without this.

examples/example-app-router/src/i18n/request.ts
```diff
@@ -1,16 +1,17 @@
...
-    ? requested
+export default getRequestConfig(async () => {
+  const params = await rootParams();
+  const locale = hasLocale(routing.locales, params.locale)
+    ? params.locale
     : routing.defaultLocale;
 
   return {
     locale,
-    messages: (await import(`../../messages/${locale}.json`)).default
```
amannn, line 14: TODO: Revert

PR #1537, opened 2024-11-12
fix: Correctly handle search params in redirects when using `trailingSlash: true`

deini:
`normalizeTrailingSlash` expects a `pathname` but it's getting pathname + search. 

This causes redirects to urls like `/users/?sort=asc/`


amannn (approved):
Awesome, thanks a bunch for the great PR @deini! 🙌

PR #1538, opened 2024-11-12
docs: clarify TypeScript version requirement and add troubleshooting for TypeScript error

boutterudy:
### 🔍 Overview
This PR enhances the **TypeScript Integration** and **Routing** sections of the `next-intl` documentation by:
- Specifying **TypeScript version 5** or later as a requirement for compatibility with `next-intl`.
- Adding a troubleshooting tip under the **Define routing** section to help developers resolve a common TypeScript error.

---

### 📄 Changes
1. **TypeScript Version Requirement**: 
   - Updated the TypeScript Integration section to recommend **TypeScript version 5 or later** to prevent issues such as `TS2554: Expected 0 arguments, but got 1`.
   - This provides users with upfront guidance to avoid compatibility issues.
2. **New Troubleshooting Detail**:
   - Added a `Details` section in **Define routing** explaining how to resolve the `TS2554` TypeScript error by upgrading to TypeScript 5 or later.
   - This gives clear instructions on how to upgrade TypeScript if users encounter this error during setup.

---

### 🤔 Why This Update?
While setting up `next-intl` with the code in the documentation, I encountered an error _(also found it [on Stackoverflow](https://stackoverflow.com/questions/79168391/next-intl-routing-ts-giving-ts2554-expected-0-arguments-but-got-1))_ due to using **TypeScript 4**.

Upgrading to **TypeScript 5** resolved the issue.

To help others avoid this confusion, I’ve added a clear version requirement and troubleshooting section to the documentation.

---

### 📌 Notes
As this is my first Pull Request for `next-intl`, please let me know if I need to correct anything 🙏

### DIFF
--- a/docs/src/pages/docs/routing.mdx
+++ b/docs/src/pages/docs/routing.mdx
@@ -46,6 +46,23 @@ Still, in case you're defining other routing config, make sure to keep them in s
 
 </Details>
 
+<Details id="typescript-version-error">
+  <summary>What if I encounter a TypeScript error like `TS2554: Expected 0 arguments, but got 1`?</summary>
+
+  If you're seeing TypeScript errors such as `TS2554: Expected 0 arguments, but got 1` when using `defineRouting` or other routing functions, this may be due to an incompatible TypeScript version. To fix this, ensure that your project is using **TypeScript version 5 or later**.
+
+  You can upgrade your TypeScript version by running:
+
+  ```sh
+  npm install typescript@latest
+  # or with yarn
+  yarn add typescript@latest
+  ```
+
+  After upgrading, restart your editor or TypeScript server to ensure the latest types are loaded.
+
+</Details>
+
 ### Locale prefix
 
 By default, the pathnames of your app will be available under a prefix that matches your directory structure (e.g. `/en/about` → `app/[locale]/about/page.tsx`). You can however adapt the routing to optionally remove the prefix or customize it per locale by configuring the `localePrefix` setting.
--- a/docs/src/pages/docs/routing.mdx
+++ b/docs/src/pages/docs/routing.mdx
@@ -46,23 +46,6 @@ Still, in case you're defining other routing config, make sure to keep them in s
 
 </Details>
 
-<Details id="typescript-version-error">
-  <summary>What if I encounter a TypeScript error like `TS2554: Expected 0 arguments, but got 1`?</summary>
-
-  If you're seeing TypeScript errors such as `TS2554: Expected 0 arguments, but got 1` when using `defineRouting` or other routing functions, this may be due to an incompatible TypeScript version. To fix this, ensure that your project is using **TypeScript version 5 or later**.
-
-  You can upgrade your TypeScript version by running:
-
-  ```sh
-  npm install typescript@latest
-  # or with yarn
-  yarn add typescript@latest
-  ```
-
-  After upgrading, restart your editor or TypeScript server to ensure the latest types are loaded.
-
-</Details>
-
 ### Locale prefix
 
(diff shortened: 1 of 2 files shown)


docs/src/pages/docs/routing.mdx
```diff
@@ -46,6 +46,23 @@ Still, in case you're defining other routing config, make sure to keep them in s
...
+
+  ```sh
+  npm install typescript@latest
+  # or with yarn
+  yarn add typescript@latest
+  ```
+
+  After upgrading, restart your editor or TypeScript server to ensure the latest types are loaded.
+
+</Details>
```
amannn, line 64: I think due to `next-intl@4` being on the horizon with the peer dependency check for TypeScript 5, I think we might be able to leave this section out for now, as it shouldn't be required anymore once that version is out. 

As a side note, I just checked and TypeScript 5 was released on March 16, 2023. I'd hope that most people who use App Router features from `next-intl` use somewhat up-to-date dependencies, as Next.js really pushes for that.

From my experience, developers often also search in issues & discussions for terms like "TS2554: Expected 0 arguments, but got 1", so a mention of this error in GitHub comments often also goes a long way without having to include it in the docs (see [docs guidelines](https://github.com/amannn/next-intl/blob/main/CONTRIBUTORS.md#documentation-guidelines)).

Does that sound reasonable to you?

docs/src/pages/docs/workflows/typescript.mdx
```diff
@@ -115,7 +115,7 @@ declare global {
 If you're encountering problems, double check that:
 
 1. Your interface uses the correct name.
-2. You're using TypeScript version 4 or later.
+2. You're using TypeScript version 5 or later.
```
amannn, line 118: Sounds fair to me. Typed messages theoretically only requires TypeScript 4, but if we mention a TypeScript version, it's probably a good idea to mention one that supports all features from `next-intl` 👍.

amannn (commented):
Many many thanks for the thoughtful PR and for sharing your findings to help others avoid this issue! ❤️

I made a note of this recently too. Currently, the TypeScript requirement is a bit in a weird state. While most features work with TypeScript 4, some newer features like `defineRouting` require TypeScript 5—as you've noticed.

I'd like to make TypeScript 5 the baseline requirement to avoid confusion, but this requires a major version bump since for some users TypeScript 4 might still work fine.

Due to this, I've now included a clear requirement for the TypeScript version for the upcoming v4 of `next-intl`: https://github.com/amannn/next-intl/pull/1481. Going forward, TypeScript will be validated using the package manager, either causing an error or printing a warning if you're trying to use `next-intl` with an older version of TypeScript. I hope that this should help users to avoid confusion without having to mention this in the docs.

docs/src/pages/docs/routing.mdx
boutterudy: Thank you very much for your explanations 🙏
I agree with everything you say, it seems reasonable to me.

Regarding the search for issues and discussions about the error, for my part I haven't found anything on GitHub (or even on the internet in general, apart from the StackOverflow discussion mentioned in my PR) 🤔

I admit I'm not very familiar with the `Discussions` part of GitHub. Are you suggesting that I open a discussion to share the solution with the community? 👀 

Anyway, no problem not including that in the documentation! ✨

amannn: > PS: This part of the documentation has been removed via [018e3a2](https://github.com/amannn/next-intl/pull/1538/commits/018e3a2e4a6a3a63db9c029f895ed90ea92f1e53)



Great, thanks! 

> I finally found [this discussion](https://github.com/amannn/next-intl/discussions/1436) 🎉

Ah yep, that seems to be the same one! I've marked the latest comment as the answer there now, hopefully that will help in case other people run into this.

amannn (approved):

PR #1541, opened 2024-11-13
feat!: Don't read a default for `useLocale` from `useParams.locale` on the client side, but rely on `NextIntlClientProvider` being used (preparation for `dynamicIO`)

amannn:
Previously, `useParams.locale` was consulted when reading from `useLocale()` on the client side, allowing to use this API even when no `NextIntlClientProvider` is used.

This behavior has now been removed because:

1. Reading from `useParams().locale` doesn't apply if you're using an [App Router setup](https://next-intl-docs.vercel.app/docs/getting-started/app-router) without i18n routing.
2. Reading from `useParams()` might require additional work from the developer in the future to work with the upcoming [`dynamicIO`](https://nextjs.org/docs/canary/app/api-reference/config/next-config-js/dynamicIO) rendering mode like adding `'use cache'` or a `<Suspense />` boundary.

Therefore, if you use any features from `next-intl` on the client side, you should now add a `NextIntlClientProvider` in the root layout and wrap all relevant components:

```tsx
import {NextIntlClientProvider} from 'next-intl';
 
export default async function LocaleLayout(/* ... */) {
 // ...
 
  return (
    <html lang={locale}>
      <body>
        <NextIntlClientProvider>
          {children}
        </NextIntlClientProvider>
      </body>
    </html>
  );
}
```

Note that also navigation APIs like `Link` rely on `useLocale` internally.


docs/src/pages/docs/usage/configuration.mdx
```diff
@@ -218,41 +228,13 @@ import {getLocale} from 'next-intl/server';
...
 
-The returned value is resolved based on these priorities:
+Depending on how a component renders, the returned locale corresponds to:
 
-1. **Server Components**: If you're using [i18n routing](/docs/getting-started/app-router), the returned locale is the one that you've either provided via [`setRequestLocale`](/docs/getting-started/app-router/with-i18n-routing#static-rendering) or alternatively the one in the `[locale]` segment that was matched by the middleware. If you're not using i18n routing, the returned locale is the one that you've provided via `getRequestConfig`.
-2. **Client Components**: In this case, the locale is received from `NextIntlClientProvider` or alternatively `useParams().locale`. Note that `NextIntlClientProvider` automatically inherits the locale if the component is rendered by a Server Component.
+1. **Server Components**: The locale represents the value returned in [`i18n/request.ts`](#i18n-request).
+2. **Client Components**: In this case, the locale is received from [`NextIntlClientProvider`](#nextintlclientprovider). Note that `NextIntlClientProvider` automatically inherits the locale if it is rendered by a Server Component.
```
amannn, line 237:
```suggestion
2. **Client Components**: The locale is received from [`NextIntlClientProvider`](#nextintlclientprovider). Note that `NextIntlClientProvider` automatically inherits the locale if it is rendered by a Server Component, so you don't need to provide this prop explicitly in this case.
```

lezan:
Hey @amannn not sure if it the best place to ask but I will try.
I am upgrading from NextJS 14 to NextJS 15 and so also next-intl to v4. I was still with the `3.20.0` and if I remember right `<NextIntlClientProvider>` was not necessary if we avoid `next-intl` in Client-Component. Correctly if I am wrong about it.
In my project I was gathering all translations in Server-Component and still now (I used Option 1 and Option 2 states [here](https://next-intl.dev/docs/environments/server-client-components#using-internationalization-in-client-components)), after upgrading to NextJS 15 I am still doing the same. But now I need to wrap all with `<NextIntlClientProvider>` in even if next-intl is used only in Server-component?
Thanks!

lezan:
Hello @amannn thanks for the quickly reply, so much appreciated.

I am coming from the release note quoted by you. So in my case I can just use `messages={null}` and `formats={null}` because I am only using next-intl in Server-Component.
I get a "performance boost", right? I mean, using only in Server-Component and putting at null `messages` and `formats`. Because to use next-intl only in Server-Component I add a lot wrapper around Client-Component to gather translations.

Thanks again!

amannn:
Yep, that sounds right! `messages={null}` and `formats={null}` will effectively bring back the old behavior of `NextIntlClientProvider` in regard to inheriting these props.

I've also add https://github.com/amannn/next-intl/commit/ade8eecdc8d5caff36109cc3d7c4a0441c5494ec now to hopefully clarify this a bit!

lezan:
Hello @amannn thanks again so much for the clarification. So kind!

if I may take advantage of your availability once again, do you suggest to switch to the new behavior or how I can stick with the old approach?

Thanks!

amannn:
Sure, no worries! The change was really about making it easier to get started as mentioned in the [relevant section of the release notes](https://next-intl.dev/blog/next-intl-4-0#nextintlclientprovider-inheritance). Handling translations in Server Components is still helpful to improve the performance of your app, the [Internationalization of Server & Client Components](https://next-intl.dev/docs/environments/server-client-components) page in the docs has all the relevant information on this.

Is there anything from your perspective that the docs leave unanswered?

Issue #1568, opened 2024-11-21
Inconsistent behavior between dev and prod for link href


damianfrizzi:
### Description

The `href` generated by the next-intl Link component is not consistent between dev and prod and produces wrong href results.

My next.config contains `trailingSlash: true` which might causes some issues.

### Verifications

- [X] I've verified that the problem I'm experiencing isn't covered in [the docs](https://next-intl-docs.vercel.app/).
- [X] I've searched for similar, existing issues on [GitHub](https://github.com/amannn/next-intl/issues) and [Stack Overflow](https://stackoverflow.com/search?q=next-intl).
- [X] I've compared my app to [a working example](https://next-intl-docs.vercel.app/examples) to look for differences.

### Mandatory reproduction URL

https://next-revalidate-bug.vercel.app/

### Reproduction description

[Source code](https://github.com/damianfrizzi/next-revalidate-bug)

### To reproduce issue 1
- Go to demo and check the `href` of the "FR" link under "Client Component"
- It shows `/fr/de-ch/` instead of `/fr/`
- Click "Go to some page" and then "Go to homepage"
- The "FR" link under "Client Component" now shows "/fr/"
- This issue happens only when running `bun run build && bun run start` and when deployed to Vercel

### To reproduce issue 2
- Go to demo and check the `href` of the "DE" link under "Client Component" or "Server Component"
- It shows `/de/` even though this is the default locale and shouldn't be shown with the locale prefix mode `as-needed`

### Expected behaviour

- The default locale with mode `as-needed` should never be part of the href
- No incorrect href is created (e.g. /fr/de-ch/) 
- The behaviour between dev and production should be consistent

amannn:
Hey, thanks a lot for the careful reproduction @damianfrizzi!

## Issue 1

Ough, that was a tricky one to diagnose. I found that there's an upstream bug in Next.js that's causing this: https://github.com/vercel/next.js/issues/73085.

I think I can add a workaround in `next-intl` though to address this: https://github.com/amannn/next-intl/pull/1571. Not sure when/if this will be addressed on the Next.js side, so that should hopefully help as an immediate fix.

I'll see if CI is happy and will have another look at this early next week to make sure the workaround doesn't have any unforeseen consequences. If this issue is urgent for you, you can use `href="/"` in the locale switcher for the time being.

## Issue 2

That's probably because you're using a wrong env param here:

```tsx
[...]

PR #1571, opened 2024-11-22
fix: Handle inconsistency in Next.js when using `usePathname` with custom prefixes, `localePrefix: 'as-needed'` and static rendering

amannn:
Fixes https://github.com/amannn/next-intl/issues/1568

### DIFF
--- a/packages/next-intl/src/navigation/react-client/useBasePathname.tsx
+++ b/packages/next-intl/src/navigation/react-client/useBasePathname.tsx
@@ -7,6 +7,7 @@ import {
   Locales
 } from '../../routing/types';
 import {
+  getLocaleAsPrefix,
   getLocalePrefix,
   hasPathnamePrefixed,
   unprefixPathname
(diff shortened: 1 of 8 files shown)


packages/next-intl/src/navigation/react-client/useBasePathname.tsx
```diff
@@ -34,11 +35,20 @@ export default function useBasePathname<
...
+
+    if (isPathnamePrefixed) {
+      unlocalizedPathname = unprefixPathname(pathname, prefix);
+    } else {
+      // https://github.com/vercel/next.js/issues/73085
+      const localeAsPrefix = getLocaleAsPrefix(locale);
+      if (hasPathnamePrefixed(localeAsPrefix, pathname)) {
+        unlocalizedPathname = unprefixPathname(pathname, localeAsPrefix);
+      }
+    }
```
amannn, line 51: One rare case where this behavior might not be wanted is this:

```tsx
const routing = defineRouting({
  locales: ['en'],
  defaultLocale: 'en',
  localePrefix: {
    mode: 'as-need', // Anything other than "always"
    prefixes: {
      en: '/uk'
    }
  }
});
```

… and then having a path like `/[locale]/en/page.tsx`.

In this case, when Next.js prerenders with a pathname set to `/en/en`, we end up with `/en` as the unprefixed pathname—that's fine. However, in production, the Next.js pathname will be `/en` after the rewrite and we'll unprefix this to `/`.

A potential workaround could be based on `useParams()`, `useSelectedLayoutSegments()` etc., but I believe a static path like `/en` can still not be detected due to only the `locale` segment being returned from these hooks. So what would be necessary is a hook that returns the actual, internal pathname, but Next.js currently doesn't support that.

So I think for the time being this is a tradeoff that we have to live with.

PR #1633, opened 2024-12-21
fix: type safe useMessages return type

theoludwig:
Currently, the `useMessages` hook return `AbstractIntlMessages` which is inconvenient for type-safety.
Instead the return type should be `IntlMessages`.

The workaround currently is to use `as`:
```tsx
const messages = useMessages() as IntlMessages
```


amannn:
Thanks for the PR @theoludwig!

This change is in fact already done for the upcoming `next-intl@4`, see https://github.com/amannn/next-intl/pull/1489. I'm aiming to have a first beta release available today, stay tuned! 😊

PR #1664, opened 2025-01-16
docs: Fix horizontal scroll on the landing page

alt-xafiro:
Hi! I'd just like to make a small fix to the site's landing page.

`width: 100vw` causes appearing of a horizontal scrollbar:

image

You can read more about the reasons for this behavior, for example, [here](https://dev.to/tepythai/why-100vw-causes-horizontal-scrollbar-4nlm), but in brief, on Windows, the vertical scrollbar occupies part of the viewport but `100vw` doesn't include the vertical scrollbar's width into its calculation, so there is not enough space for a block with `width: 100vw`. The fix is very simple: use `100%` instead of `100vw`.

In the next-intl case, `width: 100vw` is set on the `<svg>` element at the bottom of the page:

image

I tested the fix, and it doesn't break the layout at any screen breakpoints:

image

image

### DIFF
--- a/docs/src/components/GetStartedBackground.tsx
+++ b/docs/src/components/GetStartedBackground.tsx
@@ -4,7 +4,7 @@ export default function GetStartedBackground() {
   const size = 530;
   const radius = 2;
   const className =
-    'absolute top-1/2 left-1/2 transform -translate-x-1/2 -translate-y-1/2 w-screen';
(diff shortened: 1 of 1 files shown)


amannn (approved):
You're absolutely right, thank you so much for fixing this! ❤️

PR #1719, opened 2025-02-17
fix: Add workaround for OpenTelemetry/Zone.js

### LINKED ISSUE #1711: Server Translation not working if client use Grafana Faro
### Description

When using Grafana Faro, which is a client side observability tool for react application, next-intl stop working only for server side components
and print on server side console the following error
`Error: MISSING_MESSAGE: No messages were configured on the provider.`

if you remove the faro component from the reproduction URL example, it works.

### Verifications

https://github.com/bozzelliandrea/next-faro-intl

### Reproduction description

- add faro configuration (if using example, you can skip the configuration and ignore connectin error client and server side)
- create a next client side component with faro initialization (already in the example)
- add faro component to layout.tsx (already in the example) 
- run install and run next application
- switch locale from the default to another one (ex. switch from /en to /it)
- check server side console log (translation not working but middleware find locale and json mapping)
- removing FaroComponent and then the i18n works as expected with multiple locales

### Expected behaviour

Localization working with Grafana Faro when switching locale.

amannn:
Fixes #1711

Available as `next-intl@0.0.0-canary-84fe6d0`


packages/next-intl/src/shared/utils.tsx
```diff
@@ -233,3 +233,10 @@ function comparePathnamePairs(a: string, b: string): number {
 export function getSortedPathnames(pathnames: Array<string>) {
   return pathnames.sort(comparePathnamePairs);
 }
+
+export function isPromise<Value>(
+  value: Value | Promise<Value>
+): value is Promise<Value> {
+  // https://github.com/amannn/next-intl/issues/1711
+  return typeof (value as any).then === 'function';
```
SalmenBejaoui, line 241: What do you think about adding an additional check of the well-known symbol `[Symbol.toStringTag]` prototype?
Zone.js returns 'Promise', the same as the native Promise. https://github.com/angular/zone.js/blob/master/lib/common/promise.ts#L364

```
Promise.prototype[Symbol.toStringTag] === 'Promise'
zonePromise[Symbol.toStringTag] === 'Promise'
```

bozzelliandrea: Hi @amannn, first of all thanks for the quick support 😄, I tested locally with version `0.0.0-canary-84fe6d0` and it works as expected!

as @SalmenBejaoui suggest, probably `Promise.prototype[Symbol.toStringTag] === 'Promise'` could be useful, at least it should also handle the native promise override case.

about the `zonePromise[Symbol.toStringTag] === 'Promise'` could be a bit unsafe? it seems only zonejs related which isn't used on this library.

amannn: Sure, that's great to hear!

What I'm trying to understand is how the additional usage of `Symbol.toStringTag` is benefical? Is there a case where the current check for `then === 'function'` is not sufficient?

Please also see:
- https://github.com/then/is-promise/issues/38#issuecomment-619929999
- [Promises/A+ spec](https://promisesaplus.com/#terminology)



My main motivation is to have something that minifies as much as possible. Additionally, for the way `isPromise` is used in this codebase we can take some shortcuts in comparison to the `is-promise` package npm.

bozzelliandrea: Oh ok, now with the `Promise spec` I got it, so yes, I think `then === 'function'` is sufficient to cover the promise case (I don't think that someone besides overwriting the global promise object, also edits the signature of the functions, hopefully...).
looks good like this to me.

bozzelliandrea (approved):

packages/next-intl/src/shared/utils.tsx
amannn: Got it, sounds good! Many thanks for your research on this and for providing feedback, I really appreciate it and I'm sure other Next.js users will benefit from this too!

PR #1750, opened 2025-02-28
docs: Fix typo

cabbodge:
there seems to be a minor error in the getting started page, in the provided page.tsx file for `src\app\[locale]` it tries to import:

```ts
import { Link } from "@/i18n/routing";
```

which does not exist, instead it exists in the navigation file"

```ts
import { Link } from "@/i18n/navigation";`
```


amannn (approved):
Right, thank you so much for fixing it! 🙌

PR #1751, opened 2025-02-28
docs: missing vscode setting

alexpuertasr:
I was following the documentation to integrate the extension, but it wasn’t working. After checking the example here, I found that a setting was missing. Once I added it, everything started working correctly.

https://github.com/lokalise/i18n-ally/blob/main/examples/by-frameworks/next-intl/.vscode/settings.json


amannn:
Hmm, that's odd. [Examples in this repo](https://github.com/amannn/next-intl/tree/main/examples/example-app-router) don't use `.vscode/settings.json` at all and work fine for me.

Is there anything special about your setup?

alexpuertasr:
> Hmm, that's odd. [Examples in this repo](https://github.com/amannn/next-intl/tree/main/examples/example-app-router) don't use `.vscode/settings.json` at all and work fine for me.
> 
> Is there anything special about your setup?

I feel was due being a monorepo with turborepo

Issue #1458, opened 2024-10-24
[Docs]: Update Clerk example


amannn:
Clerk has released a new major version, we should update the docs accordingly.

https://clerk.com/docs/references/nextjs/clerk-middleware#combine-middleware

PR #1754, opened 2025-02-28
docs: Update Clerk example in middleware docs

willx12123:
1. change `protect` usage to latest version
2. add locale prefix matcher and non locale prefix matcher at the same time to avoid infinite redirect

Fixes https://github.com/amannn/next-intl/issues/1458

### DIFF
--- a/docs/src/pages/docs/routing/middleware.mdx
+++ b/docs/src/pages/docs/routing/middleware.mdx
@@ -263,10 +263,10 @@ import {routing} from './i18n/routing';
 
 const handleI18nRouting = createMiddleware(routing);
 
-const isProtectedRoute = createRouteMatcher(['/:locale/dashboard(.*)']);
+const isProtectedRoute = createRouteMatcher(['/dashboard(.*)', '/:locale/dashboard(.*)']);
 
-export default clerkMiddleware((auth, req) => {
-  if (isProtectedRoute(req)) auth().protect();
+export default clerkMiddleware(async (auth, req) => {
+  if (isProtectedRoute(req)) await auth.protect();
 
   return handleI18nRouting(req);
 });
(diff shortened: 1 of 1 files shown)


docs/src/pages/docs/routing/middleware.mdx
```diff
@@ -263,10 +263,10 @@ import {routing} from './i18n/routing';
 
 const handleI18nRouting = createMiddleware(routing);
 
-const isProtectedRoute = createRouteMatcher(['/:locale/dashboard(.*)']);
+const isProtectedRoute = createRouteMatcher(['/dashboard(.*)', '/:locale/dashboard(.*)']);
```
amannn, line 266: Can you explain why this is necessary? The example currently is based on `localePrefix: 'always'` since it's the default. Why is there a necessity to match `/dashboard/…`?

docs/src/pages/docs/routing/middleware.mdx
```diff
@@ -263,10 +263,10 @@ import {routing} from './i18n/routing';
...
 
-export default clerkMiddleware((auth, req) => {
-  if (isProtectedRoute(req)) auth().protect();
+export default clerkMiddleware(async (auth, req) => {
+  if (isProtectedRoute(req)) await auth.protect();
```
amannn, line 269: Can you upgrade the version number below the snippet?

amannn (commented):
Awesome thanks!

docs/src/pages/docs/routing/middleware.mdx
willx12123: I`ve made a mistake. The current example configures all routes as public and protects them through an allowlist. However, when we instead set all routes as protected by default and selectively expose public routes using `createRouteMatcher`, both route matchers with `/:locale` and not with `/:locale` become necessary to maintain proper access control.

amannn (approved):
Thanks for updating this! 🙌

PR #1800, opened 2025-03-19
docs: Remove unused createNavigation import in routing.ts setup example

kr4chinin:
Remove unused...

```ts
```

...import in `routing.ts` setup example on _app-router -> with-18n-routing_ page.


amannn (approved):
Ah yes, thanks a lot! 🙌

PR #1814, opened 2025-03-27
docs: Suggest to use `global.ts` instead of `global.d.ts` for better compatibility with `skipLibCheck`

remcohaszing:
Declaration files (`.d.ts`) are for build output. Source files (`*.ts`) are for source code. Declaration files are not type checked when `skipLibCheck` is enabled. There are legimiate use cases, but typically they should not exist outside of `node_modules`.

See https://github.com/vitejs/vite/issues/19693


amannn:
Hmm, `example-app-router` in this repo uses `skipLibCheck` and using `global.d.ts` works just fine in this app:

https://github.com/amannn/next-intl/blob/1e74950280453dedd2fda62734a00ae50353d47f/examples/example-app-router/tsconfig.json#L8

Do you have some more background or a reproduction where `global.d.ts` is ignored due to usage of `skipLibCheck`?

remcohaszing:
I decided to do a little write-up / playground to link to: https://github.com/remcohaszing/skiplibcheck

---

I just started using `next-intl` by the way and I’m loving it! :heart:

amannn:
> I decided to do a little write-up / playground to link to: https://github.com/remcohaszing/skiplibcheck

That's interesting, thanks for writing this down!

So if I understand correctly, with `skipLibCheck`, there's no type checking within `global.d.ts`, but the types from this file will be used fine across an application. I have to admit, I ran into errors myself in this file, e.g. providing an invalid path for messages and the error being swallowed silently with me being left confused as to why it's not working.

That's also the reason why in the [troubleshooting section for type augmentation](https://next-intl.dev/docs/workflows/typescript#troubleshooting), there's an entry for:

> You’re using correct paths for all modules you’re importing into your global declaration file.

I think if we get rid of the `.d` in the filename, we could get rid of that.

I mostly went with `global.d.ts` since it seems to be a known pattern across the ecosystem to use a declaration file for type-only files (like `vite-env.d.ts`, `next-env.d.ts`). That being said, I think I agree with your reasoning here.

I need to leave for today but will have another look at this tomorrow, I'll be in touch!

> I just started using `next-intl` by the way and I’m loving it! ❤️

That's awesome to hear! Definitely let me know if you have some feedback, I'm always really happy to hear from curious developers! 🤓

remcohaszing:
> > You’re using correct paths for all modules you’re importing into your global declaration file.
> 
> I think if we get rid of the `.d` in the filename, we could get rid of that.

I agree! What should we call it? The types file? Augmentation file?

> I mostly went with `global.d.ts` since it seems to be a known pattern across the ecosystem to use a declaration file for type-only files (like `vite-env.d.ts`, `next-env.d.ts`). That being said, I think I agree with your reasoning here.

I understand. I don’t know who started this pattern, but it’s wide-spread. I’m trying to unnormalize it. It’s a pure coincidence I saw this while implementing `next-intl`.

> > I just started using `next-intl` by the way and I’m loving it! ❤️
> 
> That's awesome to hear! Definitely let me know if you have some feedback, I'm always really happy to hear from curious developers! 🤓

I love that you mention a solution for markdown/MDX! Also the way you use type augmentations. One thing I do miss coming from `react-intl` is a way to synchronize message keys between the code and translations, especially to purge unused messages.

amannn (commented):
I just had another look, and I think this is a good idea—many thanks for proposing it! 🙌



---

> I agree! What should we call it? The types file? Augmentation file?

I just had a look through the docs and I think we don't have to give it a particular name—at least as far as the docs are concerned.

amannn:
> One thing I do miss coming from `react-intl` is a way to synchronize message keys between the code and translations, especially to purge unused messages.

Are you referring to [message extraction](https://formatjs.github.io/docs/getting-started/message-extraction/) here?

amannn (approved):
Awesome, thanks a lot @remcohaszing! 🙌

PR #1818, opened 2025-03-28
chore: add MDX files to tsconfig include pattern

remcohaszing:
This makes MDX files part of the TypeScript program in the editor. As a result, MDX IntelliSense uses the same configuration as the rest of the program. This also means TypeScript only needs to provide IntelliSense for one program instead of using a default fallback for MDX. This improves performance.

Before:

An import is unresolved on hover

After:

The former import is resolved on hover

### DIFF
--- a/docs/tsconfig.json
+++ b/docs/tsconfig.json
@@ -24,6 +24,12 @@
       "@/*": ["./src/*"]
     }
   },
-  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", ".next/types/**/*.ts"],
+  "include": [
+    "next-env.d.ts",
+    "**/*.mdx",
+    "**/*.ts",
+    "**/*.tsx",
+    ".next/types/**/*.ts"
+  ],
   "exclude": ["node_modules"]
 }


amannn:
That's interesting, thanks for the suggestion!

Somehow in my IDE (Cursor, based on VSCode) I don't see any hints when hovering over an import:

Screenshot 2025-03-28 at 11 54 40

This is regardless of your change here. Do you have a particular setting in your editor that enables this?

remcohaszing:
That’s interesting. I’m the developer of the [MDX extension for VSCode](https://marketplace.visualstudio.com/items?itemName=unifiedjs.vscode-mdx). I see you have proper syntax highlighting, which makes it likely to me you have it installed. It **should** just work. Sadly TypeScript plugins can conflict. You may have conflicting TypeScript plugins. See https://github.com/mdx-js/mdx-analyzer/issues/413#issuecomment-2003501727

amannn:
Oh, I just noticed the MDX extension I was using got deprecated. I've switched to your MDX extension now and I finally have working autocomplete on React components in MDX! Much better, thanks a lot! ❤️

amannn (approved):

Issue #1838, opened 2025-04-10
Turbopack and next-intl/plugin -  The config property `experimental.turbo` is deprecated.


theoludwig:
### Description

With the release of [Next.js v15.3.0](https://nextjs.org/blog/next-15-3), we can try `next build --turbopack`.

By doing so with the following `next.config.js`:
```js
import createNextIntlPlugin from "next-intl/plugin"

const withNextIntl = createNextIntlPlugin()
export default withNextIntl()
```

It yields this warning:

```js
⚠ The config property `experimental.turbo` is deprecated. Move this setting to `config.turbopack` as Turbopack is now stable.
```

This is because of those lines:
https://github.com/amannn/next-intl/blob/3192ca16d442b7bfd7b6edd579025eaadc2b9578/packages/next-intl/src/plugin/getNextConfig.tsx#L75

Is there a way to detect the Next.js version used, and if above than v15.3, then we don't use `experimental.turbo`, but directly in the `turbo` property, otherwise we keep current behavior for older versions to avoid breaking changes?

Or maybe, we can't do it automatically, but at least we could provide a param to `withNextIntl`, to tell that we should put turbo config under `turbo` property, and not with `experimental.turbo`.

### Verifications

- [x] I've verified that the problem I'm experiencing isn't covered in [the docs](https://next-intl.dev).
- [x] I've searched for similar, existing issues on [GitHub](https://github.com/amannn/next-intl/issues) and [Stack Overflow](https://stackoverflow.com/search?q=next-intl).
- [x] I've compared my app to [a working example](https://next-intl.dev/examples) to look for differences.

### Mandatory reproduction URL

https://github.com/amannn/next-intl/tree/main/examples/example-app-router

### Reproduction description

Steps to reproduce:
1. Install next-intl to a project with Next.js v15.3.0
2. Run `next build --turbopack`
3. See the warning

### Expected behaviour

No warnings.

luiscaceresd:
For the love of god do not move your turbopack configuration out of experimental settings. I started getting an error saying my next-intl configuration couldn't be found because of this.
Literally upgraded to nextintl 4.0 and re-did my whole configuration like 3 times. The issue was that I moved my turbopack configuration on my next.config file because of this warning.

Rikard-Johansson97:
@luiscaceresd I got the same problem...

henningsieh:
Hey @luiscaceresd and @Rikard-Johansson97 

[...]

PR #1840, opened 2025-04-10
update: move next turbo config from experimental to top level

sakib412:


amannn (commented):
This would unfortunately be a breaking change for users on Next.js <15.3.

Please see the previous discussion here:
- https://github.com/amannn/next-intl/issues/1838
- https://github.com/vercel/next.js/issues/78009

sakib412:
> This would unfortunately be a breaking change for users on Next.js <15.3.
> 
> Please see the previous discussion here:
> 
> * [Turbopack and next-intl/plugin -  The config property `experimental.turbo` is deprecated. #1838](https://github.com/amannn/next-intl/issues/1838)
> * [Deprecation warning for `experimental.turbo` with MDX plugin (also affects other downstream plugins) vercel/next.js#78009](https://github.com/vercel/next.js/issues/78009)

Got you! Either we have to wait for to get the nextjs version in next config or we can pass the nextjs version in 

`createNextIntlPlugin('path', options = {nextVersion: '15.3.0'})`

 function arg as options. (we can parse the version from `package.json` or node_modules)
and also this is just a warning we can ignore for now, anyway nextjs will release major version for breaking change.

amannn:
> we can parse the version from `package.json` or node_modules

I thought about that too, but I'm afraid there could be edge cases. E.g., can we make an assumption about where a `package.json` file should be? E.g. Next.js supports the [`[directory]`](https://nextjs.org/docs/pages/api-reference/cli/next#next-dev-options) argument for `next dev`, therefore `process.cwd()` could be somewhere else.

The closest I could think of would be to use `require.resolve('next')` to get a path to the Next.js dependency folder and from there traverse upwards to find the closest `package.json` and read the version from there.

It sounds a bit hairy though and maybe there are edge cases. So ideally, Next.js would help a bit here, let's see if we can get support from them in https://github.com/vercel/next.js/issues/78009.

amannn:
This was resolved in [v4.0.3](https://github.com/amannn/next-intl/releases/tag/v4.0.3)

PR #1850, opened 2025-04-17
fix: Support stable Turbopack config in Next.js 15.3

### LINKED ISSUE #1838: Turbopack and next-intl/plugin -  The config property `experimental.turbo` is deprecated.

amannn:
Fixes #1838

In Next.js 15.3, [Turbopack config has become stable](https://nextjs.org/blog/next-15-3#turbopack-configuration-in-nextconfigts-stable). With this fix, the new option is used in order to avoid a deprecation warning.


DKeken:
+1
need


@repo/client:dev:  ⚠ The config property `experimental.turbo` is deprecated. Move this setting to `config.turbopack` as Turbopack is now stable.

PR #1855, opened 2025-04-21
Docs: update way we import './messages/en.json'

Git-I985:
Replaced top level import with typeof import, to avoid problems with top level imports in d.ts.file https://www.typescriptlang.org/docs/handbook/2/modules.html?utm_source=chatgpt.com#:~:text=file%20without%20any%20top%2Dlevel%20import%20or%20export%20declarations

`import en from './messages/en.json';` will almost turn off this `global.d.ts` file, better to use `typeof import()` as an example

Please update this post also https://next-intl.dev/blog/next-intl-4-0#revamped-augmented-types

### DIFF
--- a/docs/pages/docs/workflows/typescript.mdx
+++ b/docs/pages/docs/workflows/typescript.mdx
@@ -28,9 +28,8 @@ function About() {
 To enable this validation, add a global type definition file in your project root (e.g. `global.d.ts`):
 
 ```jsx filename="global.d.ts"
-import en from './messages/en.json';
 
-type Messages = typeof en;
+type Messages = typeof import("./messages/en.json");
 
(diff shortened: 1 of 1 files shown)


amannn:
Hey, it seems like your branch is based on an outdated state of `main`:

image

 With https://github.com/amannn/next-intl/pull/1814, @remcohaszing switched `global.d.ts` to `global.ts`.

Does that avoid the issue you want to address here?

Git-I985:
>Does that avoid the issue you want to address here?
not exactly
>Hey, it seems like your branch is based on an outdated state of main:
hm, so weird, sorry, i will check

amannn:
Closing due to inactivity …

Issue #1845, opened 2025-04-16
`redirect` ignores locale when `localePrefix` is set to "as-needed"


EhabY:
#### Overview

When using `next-intl`'s `redirect` function with the default locale and `localePrefix: "as-needed"`, redirecting to the default locale appears to fail. Instead of navigating to the correct default locale path, it redirects to the previously visited locale. This behavior makes `redirect` unreliable for default locale navigation.

To work around this issue, I had to use the following logic in my code:

```ts
if (locale === "en") {
  redirect({ href: `/${locale}${path}`, locale });
} else {
  redirect({ href: path, locale });
}
```

This workaround defeats the purpose of using `redirect` and is not ideal.

### Verifications

- [x] I've verified that the problem I'm experiencing isn't covered in [the docs](https://next-intl.dev).
- [x] I've searched for similar, existing issues on [GitHub](https://github.com/amannn/next-intl/issues) and [Stack Overflow](https://stackoverflow.com/search?q=next-intl).
- [x] I've compared my app to [a working example](https://next-intl.dev/examples) to look for differences.

### Mandatory reproduction URL

https://github.com/EhabY/next-intl-redirect-bug

### Reproduction description

1. Open the reproduction app.
2. Navigate to `/es`.
3. Navigate to `/redirect?locale=en`.
4. You’ll notice it redirects to `/es` instead of the default English path.
5. Try the reverse: go to `/redirect?locale=es` — this works as expected.

> Note: Disabling `localeDetection` resolves the issue, but ideally, `redirect` should behave consistently regardless of `localeDetection` settings or stored locale preferences (e.g., in cookies or headers).

### Expected behaviour

1. Navigate to `/es`.
2. Navigate to `/redirect?locale=en`.
3. Expected result: be redirected to `/` (default English locale), not the previously visited `/es`.

EhabY:
My use case might be a bit unique. I'm using [[Payload CMS](https://github.com/payloadcms/payload)](https://github.com/payloadcms/payload), which has its own `locale` handling in the admin dashboard that is separate from the one used by `next-intl`. In my setup, I have an admin dashboard with a `preview` button that allows me to view a draft of the page I'm editing.

The issue I'm encountering is with how locale is handled when previewing drafts. If I first preview a draft in Spanish, then switch to English and preview the same page again, I'm still redirected to the Spanish version.

[...]

PR #1865, opened 2025-04-23
feat: Add `forcePrefix` option for `redirect` and `getPathname`

amannn:
Resolves https://github.com/amannn/next-intl/issues/1845


packages/next-intl/src/navigation/shared/createSharedNavigationFns.tsx
```diff
@@ -101,14 +101,12 @@ export default function createSharedNavigationFns<
...
-            // @ts-expect-error -- This is ok
-            href: pathnames == null ? pathname : {pathname, params}
-          },
-          locale != null || undefined
-        )
+      ? getPathname({
+          locale: locale || curLocale,
+          // @ts-expect-error -- This is ok
+          href: pathnames == null ? pathname : {pathname, params},
+          forcePrefix: locale != null || undefined
```
amannn, line 108:
```suggestion
          // Always include a prefix when changing locales
          forcePrefix: locale != null || undefined
```

Issue #1056, opened 2024-05-06
Support Intl.RelativeTimeFormat for `relativeTime` formatting


kvnxiao:
### Is your feature request related to a problem? Please describe.

Currently, the formatter for [`Formatting date and time ranges`](https://next-intl-docs.vercel.app/docs/usage/dates-times#date-time-ranges) does not support additional configuration other than specifying the exact unit to display the output in.

The relative time formatting should support all the options from [`Intl.RelativeTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat) and not just `RelativeTimeFormatUnit`

### Describe the solution you'd like

Instead of:

```typescript
type RelativeTimeFormatOptions = {
    now?: number | Date;
    unit?: Intl.RelativeTimeFormatUnit;
};
export default RelativeTimeFormatOptions;
```

The `RelativeTimeFormatOptions` type should also accept the options from the [`Intl.RelativeTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat/RelativeTimeFormat#options) constructor:
* numberingSystem
* style
* numeric

### Describe alternatives you've considered

The only alternative would be to use the native `Intl` API directly, but this should be integrated into the formatter in `next-intl` IMO

amannn:
Hey, thanks for the report! I agree, we could definitely support `numberingSystem` and `style`. As for `numeric` I'm a bit hesitant though, as it can lead to issues when values are rounded: https://github.com/amannn/next-intl/pull/765.

Do you need `numeric` for your use case? If so, can you include further details?

I've added #1057 to look into this.

kvnxiao:
At the moment, I only have a need for the `style` to specify `long`, `short`, or `narrow`.

I think numberingSystem and style is sufficient for now!

amannn:
Sounds good, thanks for the feedback!

amannn:
Interesting. Can you share what your exact use case is?

From my personal observation, "yesterday" is a nice thing to read, but it's easy to run into a wrong output due to rounding (see https://github.com/amannn/next-intl/issues/1056#issuecomment-2097876494). "Yesterday" might not actually be the day before today, but could be the day before yesterday.

kvnxiao:
[...]

PR #1872, opened 2025-05-02
feat: Use `numeric: 'auto'` for relative times that don't need to be rounded for a given unit

amannn:
This allows to use idiomatic phrasing like "yesterday" in case a date is provided that doesn't have a fractional part when converted to the target unit.

See the [updated docs](https://next-intl-docs-git-feat-1056-relativetimeauton-a219f0-next-intl.vercel.app/docs/usage/dates-times#relative-times-unit).

([ref](https://github.com/amannn/next-intl/issues/1056#issuecomment-2843281594))

### DIFF
--- a/docs/src/pages/docs/usage/dates-times.mdx
+++ b/docs/src/pages/docs/usage/dates-times.mdx
@@ -168,9 +168,9 @@ function Component() {
 }
 ```
 
-### Customizing the unit [#relative-times-unit]
+### `unit` [#relative-times-unit]
 
-By default, `relativeTime` will pick a unit based on the difference between the passed date and `now` like "3 seconds" or "5 days".
+By default, `relativeTime` will pick an appropriate unit based on the difference between the passed date and `now` like "3 seconds", "5 days" or "1 year".
 
 If you want to use a specific unit, you can provide options via the second argument:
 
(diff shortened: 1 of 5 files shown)


docs/src/pages/docs/usage/dates-times.mdx
```diff
@@ -187,6 +187,33 @@ function Component() {
...
+  // The "start of a day" depends on a time zone
+  return startOfDay(date, {in: tz(timeZone)});
+}
+
+// Renders "yesterday" instead of "1 day ago"
+const result = format.relativeTime(normalize(dateTime), {
+  now: normalize(now),
+  unit: 'day'
+});
+```
```
amannn, line 215: Been scratching my head a bit over this code snippet. I think it should now be correct, but it's a bit verbose.

kvnxiao: This is technically what I had to do anyways with my current approach (with the exception of using Luxon instead of date-fns) to ensure that `numeric: auto` gave the proper localized text for "yesterday", "today", etc.

kvnxiao: Not sure if there's a way to simplify this, unless we allow passing in an optional config param like "shouldDormalizeDate: boolean` to the `format.relativeTime()` options - but that gets into the issue of requiring `next-intl` to supply its own "start of day" function calculation if it wants to be date-library agnostic.

packages/use-intl/src/core/createFormatter.tsx
```diff
@@ -309,24 +300,40 @@ export default function createFormatter(props: Props) {
...
+      // However, it should not be used with rounded values, as the given
+      // dates might cross a threshold were the output isn't correct anymore.
+      // Example: 2024-01-08T23:00:00.000Z and 2024-01-08T01:00:00.000Z would
+      // produce "yesterday", which is not the case. By using `always` in this
+      // case, we can ensure correct output.
+      //
+      // Note that due to approximations being used for months and years, it's
+      // practically impossible to trigger the cases "last month" or "last year".
+      if (unitValue === rounded) {
+        opts.numeric = 'auto';
```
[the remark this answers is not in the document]
amannn: Observation: Even GitHub seems to get this wrong.

Screenshot 2025-06-27 at 14 13 53

(see the date in the upper right corner)

Issue #1385, opened 2024-09-30
[Docs]: Given passthrough Root layout doesn't compile


alexmge:
### Link to page

https://next-intl-docs.vercel.app/docs/environments/error-files#catching-non-localized-requests

### Describe the problem

This documentation says that if you want to catch non localized requests with a `not-found.tsx` file, you can still add `app/not-found.tsx` before the `app/[lang]` route and add a passthrough root layout in `app/layout.tsx`.

The given passthrough root layout example is
``` ts
// Since we have a root `not-found.tsx` page, a layout file
// is required, even if it's just passing children through.
export default function RootLayout({children}) {
  return children;
}
```

However, as described in the NextJS documentation [here](https://nextjs.org/docs/messages/missing-root-layout-tags), the `html` and `body` tags are required in the root layout, leading to a compilation error if missing (at least that's what I experienced).

amannn:
The Next.js docs are unfortunately a bit inconsistent, e.g. [here](https://nextjs.org/docs/app/building-your-application/routing/internationalization#static-generation) the `html` and `body` tag is suggested at `app/[lang]/layout.tsx`.

daveycodez:
<img width="561" alt="Image" src="https://github.com/user-attachments/assets/af3656f3-9e71-42c9-8369-763902c4fd47" />

@amannn Are we supposed to just ignore this error for root not-found.tsx?

daveycodez:
I figured out we can just put suppressHydrationWarning on both html tags on both layouts, that should do the fix

PR #1927, opened 2025-06-19
docs: clarify root layout requirements and `suppressHydrationWarning` usage with App Router

boutterudy:
# ❓ Context
While working with `next-intl` and the App Router, I ran into hydration warnings due to `<html>` and `<body>` tags being duplicated in `app/layout.tsx` and `app/[locale]/layout.tsx`: 
```
Error: Missing <html> and <body> tags in the root layout. 
```

This issue was hard to track down until I found [issue #1385](https://github.com/amannn/next-intl/issues/1385), which explains the cause but doesn't currently link to a fix in the documentation.

# ✅ What This PR Does
This PR updates the documentation to:
- Clarify that `<html>` and `<body>` must exist in both layouts
- Show how to prevent hydration warnings by adding `suppressHydrationWarning`
- Provide full examples for both `app/layout.tsx` and `app/[locale]/layout.tsx`

# ✨ Wonderful Result
image

### DIFF
--- a/docs/src/pages/docs/getting-started/app-router/with-i18n-routing.mdx
+++ b/docs/src/pages/docs/getting-started/app-router/with-i18n-routing.mdx
@@ -215,6 +215,28 @@ const withNextIntl = createNextIntlPlugin(
 
 </Details>
 
+### `src/app/layout.tsx` [#rootlayout]
+
+<Callout>
+<strong>Note on nested `<html>` and `<body>` tags:</strong>
+
+Next.js <strong>requires</strong> `<html>` and `<body>` in `app/layout.tsx`.
+If your `[locale]/layout.tsx` also contains them (it usually does so it can set `lang`), the tags are rendered twice and you’ll see hydration warnings in development.
+
+Add `suppressHydrationWarning` to <em>both</em> tags in <strong>both</strong> layouts
+
+</Callout>
+
+```tsx filename="app/layout.tsx"
+export default function RootLayout({children}: {children: React.ReactNode}) {
+  return (
+    <html suppressHydrationWarning>
+      <body suppressHydrationWarning>{children}</body>
+    </html>
+  );
+}
+```
+
 ### `src/app/[locale]/layout.tsx` [#layout]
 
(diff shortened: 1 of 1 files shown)


amannn:
If you want to add a layout at the root, it shouldn't include tags like `html`, but simply return children (see [the docs](https://next-intl.dev/docs/environments/error-files#catching-non-localized-requests)).

Based on my latest tests, this layout is not strictly required, but for a `not-found.js` file at the root, it currently is.

Note that this will change in the near future though: [`global-not-found.js`](https://x.com/huozhi/status/1921690791694446919)

boutterudy:
> If you want to add a layout at the root, it shouldn't include tags like `html`, but simply return children (see [the docs](https://next-intl.dev/docs/environments/error-files#catching-non-localized-requests)).
> 
> Based on my latest tests, this layout is not strictly required, but for a `not-found.js` file at the root, it currently is.
> 
> Note that this will change in the near future though: [`global-not-found.js`](https://x.com/huozhi/status/1921690791694446919)

Reading the documentation of `next-intl`, that's actually what I thought, until I encountered the error 😄 

Without the root layout, I encountered an error, and that should be the case for everyone (at least in the latest version of Next.js), as described in the Next.js documentation [here](https://nextjs.org/docs/app/getting-started/layouts-and-pages#:~:text=The%20root%20layout%20is%20required%20and%20must%20contain) : 

> _The root layout is required and must contain `html` and `body` tags._

So, with the latest version of Next.js, I personally think that the example I've included in the documentation will help a lot of people that will use `next-intl` with the latest version of Next.js ✨ 

@amannn What do you think? 👀

amannn:
Rendering `html` and `body` tags is unfortunately not the solution, as it will result in invalid markup (which is where your need for `suppressHydrationWarning` comes from).

The upcoming `global-not-found.js` is really the proper solution here and we can get rid of the pass-through root layout.

The examples in this repo btw. work without errors and can be used for inspiration when in doubt.

amannn:
Closing due to inactivity …

PR #1945, opened 2025-06-30
fix: Add friendly error message when failing to load current Next.js version

cupofjoakim:
As discussed in https://github.com/amannn/next-intl/discussions/1943, this adds a flightly friendlier error message for when `createRequire` is accidentally ran in an unexpected environment.

### DIFF
--- a/packages/next-intl/src/plugin/hasStableTurboConfig.tsx
+++ b/packages/next-intl/src/plugin/hasStableTurboConfig.tsx
@@ -7,7 +7,7 @@ function getCurrentVersion() {
     return pkg.version;
   } catch (error) {
     throw new Error(
-      'Failed to get current Next.js version. This can happen if next-intl/plugin is imported into your app code outside of your next.config.js.,
+      'Failed to get current Next.js version. This can happen if next-intl/plugin is imported into your app code outside of your next.config.js.',
       {cause: error}
     );
   }
--- a/packages/next-intl/src/plugin/hasStableTurboConfig.tsx
+++ b/packages/next-intl/src/plugin/hasStableTurboConfig.tsx
@@ -7,7 +7,7 @@ function getCurrentVersion() {
     return pkg.version;
   } catch (error) {
     throw new Error(
-      'Failed to get current Next.js version. This can happen if next-intl/plugin is imported outside of your next.config.js, like in your test setup',
+      'Failed to get current Next.js version. This can happen if next-intl/plugin is imported into your app code outside of your next.config.js.,
       {cause: error}
     );
   }
--- a/packages/next-intl/src/plugin/hasStableTurboConfig.tsx
+++ b/packages/next-intl/src/plugin/hasStableTurboConfig.tsx
@@ -1,8 +1,17 @@
-// eslint-disable-next-line import/order
 import {createRequire} from 'module';
 
-const require = createRequire(import.meta.url);
-const pkg = require('next/package.json');
+function getCurrentVersion() {
+  try {
+    const require = createRequire(import.meta.url);
+    const pkg = require('next/package.json');
+    return pkg.version;
+  } catch (error) {
+    throw new Error(
+      'Failed to get current Next.js version. This can happen if next-intl/plugin is imported outside of your next.config.js, like in your test setup',
+      {cause: error}
+    );
+  }
+}
 
 function compareVersions(version1: string, version2: string) {
   const v1Parts = version1.split('.').map(Number);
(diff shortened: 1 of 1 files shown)


packages/next-intl/src/plugin/hasStableTurboConfig.tsx
```diff
@@ -1,8 +1,17 @@
...
-const require = createRequire(import.meta.url);
-const pkg = require('next/package.json');
+function getCurrentVersion() {
+  try {
+    const require = createRequire(import.meta.url);
+    const pkg = require('next/package.json');
+    return pkg.version;
+  } catch (error) {
+    throw new Error(
+      'Failed to get current Next.js version. This can happen if next-intl/plugin is imported outside of your next.config.js, like in your test setup',
```
amannn: It's true that the error might become visible in a test or e.g. in the Edge runtime (due to missing Node.js APIs). However, I'd say it's less the issue that you're importing this into a test setup, but it's already not ideal if you import this into any of your app code since we're eagerly doing unnecessary work (reading `package.json`).

So I think I'd focus more on that in the error message:

```suggestion
      'Failed to get current Next.js version. This can happen if next-intl/plugin is imported into your app code outside of your next.config.js.,
```

What do you think?

packages/next-intl/src/plugin/hasStableTurboConfig.tsx
```diff
@@ -17,5 +26,6 @@ function compareVersions(version1: string, version2: string) {
   return 0;
 }
 
-const hasStableTurboConfig = compareVersions(pkg.version, '15.3.0') >= 0;
+const hasStableTurboConfig =
+  compareVersions(getCurrentVersion(), '15.3.0') >= 0;
```
amannn, line 30: I'm wondering if it could help to turn this variable into a function instead:

```tsx
function hasStableTurboConfig() {
  return compareVersions(getCurrentVersion(), '15.3.0') >= 0;
}
```

The benefit is that the app will keep working due to the lazy execution of `createRequire`.

A downside however is that in the Edge runtime you'll get a cryptic warning:

```
 ⚠ ./packages/next-intl/dist/esm/development/plugin/createMessagesDeclaration.js:40:3
Ecmascript file had an error
  38 |     compileDeclaration(messagesPath, true);
  39 |   });
> 40 |   process.on('exit', () => {
     |   ^^^^^^^^^^
  41 |     watcher.close();
  42 |   });
  43 | }

A Node.js API is used (process.on at line: 40) which is not supported in the Edge Runtime.
Learn more: https://nextjs.org/docs/api-reference/edge-runtime
```

The full stack trace is also not present, so this might almost look like a bug in `next-intl`. So maybe it's in fact better to keep the execution eagerly to provide a better error message.

amannn (commented):
Tested it and it works! 🙌 

Screenshot 2025-07-01 at 09 18 55

I've left two minor comments. I think I'd reword the warning slightly, but ship it as-is otherwise. Would be curious about your thoughts though!

packages/next-intl/src/plugin/hasStableTurboConfig.tsx
cupofjoakim: I agree, the language is clearer and I much prefer it. I've added it.

packages/next-intl/src/plugin/hasStableTurboConfig.tsx
cupofjoakim: Personally I'd take clarity. Not sure I should have any say on that though! I'm a non-edge consumer of the library that just got burnt by the lack of clarity, ofc clarity will be top of mind for me at the moment :) 

Ideally we'd get to a point where clarity exists in the edge runtime case too, but perhaps that should be left for future improvements.

packages/next-intl/src/plugin/hasStableTurboConfig.tsx
```diff
@@ -1,8 +1,17 @@
...
-const require = createRequire(import.meta.url);
-const pkg = require('next/package.json');
+function getCurrentVersion() {
+  try {
+    const require = createRequire(import.meta.url);
+    const pkg = require('next/package.json');
+    return pkg.version;
+  } catch (error) {
+    throw new Error(
+      'Failed to get current Next.js version. This can happen if next-intl/plugin is imported into your app code outside of your next.config.js.,
```
amannn:
```suggestion
      'Failed to get current Next.js version. This can happen if next-intl/plugin is imported into your app code outside of your next.config.js.',
```

packages/next-intl/src/plugin/hasStableTurboConfig.tsx
amannn: Sounds good, I agree!

PR #1966, opened 2025-07-26
docs: Update vscode-integration.mdx

ifeanyi-ugwu:


docs/src/pages/docs/workflows/vscode-integration.mdx
```diff
@@ -26,6 +26,7 @@ These extensions are known to support `next-intl`:
 
 ```json filename=".vscode/settings.json"
 "i18n-ally.localesPaths": ["./path/to/your/messages"], // E.g. "./messages"
+"i18n-ally.enabledFrameworks": ["next-intl"],
```
amannn, line 29: I don't think this is required, i18n Ally should auto-detect `next-intl`. Did you encounter a problem with this?

ifeanyi-ugwu: Yes, it couldn't auto detect my framework until I added that. Maybe because I was using a monorepo?

amannn: That could be the case, yes! If there's an issue with this, it might be worth reporting upstream in the i18n Ally repo.

ifeanyi-ugwu: Sure. I assumed it was a configuration thing, since adding it seemed to fix it. I'll add this to the bunch of open issues on i18n ally's repo.

amannn: Ok thanks! I'll close this PR then …

PR #1993, opened 2025-08-21
docs: Revamp getting started

amannn:


PMudra:
It appears that this pull request has removed the 'How can I reuse messages?' section from the documentation. I found that section really valuable. It was my go-to reference for structuring translation messages.

I would like to know whether the information in this paragraph is still valid. Or is there a new best practice described in the video?

Is the general plan to replace text documentation with videos?

Thank you very much for your work.

amannn:
Thanks for the question!

Here's the previous section for reference:

<details>

<summary>How can I reuse messages?</summary>

As your app grows, you'll want to reuse messages among your components. If you use component names as namespaces to structure your messages, you'll automatically benefit from reusable messages by reusing your components.

**Examples:**

- You're displaying products in your app and often need to resolve a category identifier to a human readable label (e.g. `new` → "Just in"). To ensure consistency, you can add a `ProductCategory` component that turns the `category` into a styled badge.
- You commonly need a dialog that displays a "confirm" and "cancel" button. In this case, consider adding a `ConfirmDialog` component to reuse the messages along with the functionality.

There might be cases where you want to use the same message in different components, but there's no reasonable opportunity for sharing the message via a component. This might be symptom that these components should use separate messages, even if they happen to be the same for the time being. By using separate labels, you gain the flexibility to use more specific labels (e.g. "not now" instead of "cancel").

For edge cases where the reuse of messages among different components is hard to achieve with shared components, you can either use messages from different namespaces via `useTranslations()`, or you can consider adding a hook that shares the translation behavior:

```tsx filename="useLocaleLabel.tsx"
export default function useLocaleLabel() {
  const t = useTranslations('useLocaleLabel');

  function getLocaleLabel(locale: 'en' | 'de') {
    return t('label', {locale});
  }

  return getLocaleLabel;
}
```

```json
{
  "useLocaleLabel": {
    "label": "{locale, select, en {English} de {German} other {Unknown}}"
  }
}
```

</details>

---

I've removed the section because it misses important context on when it's a good idea to reuse messages and when not. The now-linked video explains this in more detail with a practical example.

But for labels that you surely want to reuse, the approach with the hook is still useful and is also demonstrated as a pattern in the linked video.

Another aspect here is that this goes quite a bit into the area of personal preferences. I know some users simply reuse messages by sharing a key—which is also fine, but has different tradeoffs. Or the recently introduced `useExtracted` makes it also pretty easy to share messages without having to find strategies with namespaces and keys, I know that many developers also follow this approach.

> Is the general plan to replace text documentation with videos?

The docs for `next-intl` are surely here to stay and will remain just as free as the library itself.

For this particular topic there's just less a canonical solution and a lot more of "it depends", I chose to move this content to [learn.next-intl.dev](https://learn.next-intl.dev/) where there's more breathing room for these kinds of discussions. The docs in contrast try to focus more on the functional topics.

Last but not least, there's of course also the side of ensuring a sustainable future for `next-intl`, so for the category of best practices and architectural guidance I chose to make this paid content, which in turn helps `next-intl` to survive as a free and open-source library.

Hope that makes sense!

PR #1996, opened 2025-08-24
feat: Add `withFallback` method for per-message fallback control

seelepeeku:
This is an extension to an already closed issue #88

As far as I'm aware (as a newbie...), currently the only ways of "falling back" are passing a `getMessageFallback` to the provider, or getting messages from another locale. 
The former is too restrictive as it works on a global scale (unless I'm tripping), while the latter additionally may not always work (`MISSING_MESSAGE` for example occurs from the entire messages objects whether you're using a fallback locale or not, so merging messages doesn't do that much here if the message isn't available in other locales).
I thought it would be neat if we had a shorthand way of providing fallback on an individual message request basis, i.e.:
```js
t.withFallback("key", "some fallback")
```

For example, the code may fail to retrieve a message translation if the key does not exist (`MISSING_MESSAGE`). For my usage in the app I'm working on, I'm forced to call `t.has("key")` and conditionally render `t("key")` or a fallback if I don't want the code to just error out:
```js
someName = tItems.has(`${someId}.name`)
      ? t(`${someId}.name`)
      : `Hello ${someId}`;
```
If I want to provide specialized fallback for specific messages, this gets cumbersome with scale. So I had to use a custom hook:
```js
export function useAppTranslations(namespace?: string) {

  const t = useTranslations(namespace);

  const safeTranslate = (
    key: string,
    fallback?: unknown
  ): string => {
    if (!fallback) {
      fallback = "some fallback";
    }
    try {
      return t.has(key) ? t(key) : fallback;
    } catch {
      return fallback;
    }
  };

  return {
    safeTranslate
  };
}
```
Then I just get `safeTranslate` and use it everywhere I need. 
```js
someName = safeTranslate(`${someId}.name`, `Hello ${someId}`);
```
But this was just a hack and came with its own issues, such as breaking integration with VSCode intl extensions.

This PR adds a `withFallback` method that utilizes the `has` method to provide optional fallback in case an error occurs with receiving the message. It supports two arguments; 
- `fallback`: a `string` or a function that resolves to a string.
- `translateFallback`: a `boolean` that defaults to true, which controls whether the translator should attempt to use the `fallback` argument provided as a translation key then output it as fallback.

P.S. I kind of would have really liked for `fallback` to have been a standard optional argument of the base `t` call, but I felt like that would require me to dive deeper into the code to do.

### DIFF
--- a/packages/use-intl/src/core/createBaseTranslator.tsx
+++ b/packages/use-intl/src/core/createBaseTranslator.tsx
@@ -435,5 +435,41 @@ function createBaseTranslatorImpl<
     }
   };
 
+  translateFn.withFallback = (
+    key: string,
+    fallback: string | (() => string),
+    /** Whether to attempt to resolve fallback as a translation key. Defaults to true. */
+    translateFallback = true,
+    /** Key value pairs for values to interpolate into the message. */
+    values?: TranslationValues,
+    /** Provide custom formats for numbers, dates and times. */
+    formats?: Formats
+  ): string => {
+    if (translateFn.has(key)) {
+      const result = translateBaseFn(key, values, formats);
+      if (typeof result === 'string') {
+        return result;
+      }
+    }
+
+    const fallbackValue =
+      typeof fallback === 'function' ? fallback() : fallback;
+
+    if (translateFallback) {
+      if (!translateFn.has(fallbackValue)) {
+        if (process.env.NODE_ENV !== 'production') {
+          throw new Error(`Fallback key '${fallbackValue}' does not exist`);
+        }
+      }
+
+      const fallbackResult = translateBaseFn(fallbackValue);
+      if (typeof fallbackResult === 'string') {
+        return fallbackResult;
+      }
+    }
+
+    return fallbackValue;
+  };
+
   return translateFn;
 }
--- a/packages/use-intl/src/core/createTranslator.test.tsx
+++ b/packages/use-intl/src/core/createTranslator.test.tsx
@@ -882,3 +882,36 @@ describe('t.raw', () => {
     expect(t.raw('rich')).toBe('<b>Hello <i>{name}</i>!</b>');
   });
 });
+
+describe('t.withFallback', () => {
+  it('handles missing messages with fallbacks', () => {
+    const t = createTranslator({
+      locale: 'en',
+      messages: {
+        Home: {fallback: 'Default message'}
+      } as any
+    });
+
+    // Valid translation key fallback
+    expect(t.withFallback('Home.missing', 'Home.fallback')).toBe(
+      'Default message'
+    );
+
+    // Literal string fallback
+    expect(t.withFallback('Home.missing', 'Literal text', false)).toBe(
+      'Literal text'
+    );
+
+    // Invalid fallback key
+    expect(() => t.withFallback('Home.missing', 'Invalid.key')).toThrow();
+
+    // Dynamic function that resolves to a literal string
+    expect(t.withFallback('Home.missing', () => 'Dynamic', false)).toBe(
+      'Dynamic'
+    );
+    // Dynamic function that resolves to a valid translation key
+    expect(t.withFallback('Home.missing', () => 'Home.fallback')).toBe(
+      'Default message'
+    );
+  });
+});
--- a/packages/use-intl/src/core/createTranslator.tsx
+++ b/packages/use-intl/src/core/createTranslator.tsx
@@ -163,6 +163,19 @@ export default function createTranslator<
   has<TargetKey extends NamespacedMessageKeys<TranslatorMessages, Namespace>>(
     key: TargetKey
   ): boolean;
+
+  // `withFallback`
+  withFallback<
+    TargetKey extends NamespacedMessageKeys<TranslatorMessages, Namespace>
+  >(
+    key: TargetKey,
+    fallback: string | (() => string),
+    translateFallback?: boolean,
+    ...args: TranslateArgs<
+      NamespacedValue<TranslatorMessages, Namespace, TargetKey>,
+      RichTagsFunction
+    >
+  ): string;
 } {
   // We have to wrap the actual function so the type inference for the optional
   // namespace works correctly. See https://stackoverflow.com/a/71529575/343045
--- a/packages/use-intl/src/react/useTranslations.test.tsx
+++ b/packages/use-intl/src/react/useTranslations.test.tsx
@@ -605,6 +605,36 @@ describe('t.has', () => {
   });
 });
 
+describe('t.withFallback', () => {
+  function Component() {
+    const t = useTranslations();
+    return (
+      <>
+        <div data-testid="fallback-key">
+          {t.withFallback('missing', 'Home.fallback')}
+        </div>
+        <div data-testid="fallback-literal">
+          {t.withFallback('missing', 'Literal text', false)}
+        </div>
+        <div data-testid="fallback-function">
+          {t.withFallback('missing', () => 'Dynamic', false)}
+        </div>
+      </>
+    );
+  }
+
+  it('returns the fallback value when the message is missing', () => {
+    render(
+      <IntlProvider locale="en" messages={{Home: {fallback: 'Default message'}}}>
+        <Component />
+      </IntlProvider>
+    );
+    screen.getByText('Default message');
+    screen.getByText('Literal text');
+    screen.getByText('Dynamic');
+  });
+});
+
 describe('error handling', () => {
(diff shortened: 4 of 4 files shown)


amannn:
Hey, thanks for sharing your thoughts here!

`getMessageFallback` is in fact only intended to customize the error case for apps, where missing messages should be treated as errors.

If you want to use messages e.g. from the default locale as a fallback, you can consider merging them when [providing messages](https://next-intl.dev/docs/usage/configuration#messages):

```tsx

export default getRequestConfig(async () => {
  const locale = 'fr-CA';

  // Use fallback from `en-US` for missing messages
  const localeMessages = (await import(`../../messages/${locale}.json`))
    .default;
  const defaultMessages = (await import(`../../messages/en-US.json`)).default;
  const messages = {...defaultMessages, ...localeMessages};

  // ...
});
```

seelepeeku:
> Hey, thanks for sharing your thoughts here!
> 
> `getMessageFallback` is in fact only intended to customize the error case for apps, where missing messages should be treated as errors.
> 
> If you want to use messages e.g. from the default locale as a fallback, you can consider merging them when [providing messages](https://next-intl.dev/docs/usage/configuration#messages):
> 
> ```tsx
> import {getRequestConfig} from 'next-intl/server';
> 
> export default getRequestConfig(async () => {
>   const locale = 'fr-CA';
> 
>   // Use fallback from `en-US` for missing messages
>   const localeMessages = (await import(`../../messages/${locale}.json`))
>     .default;
>   const defaultMessages = (await import(`../../messages/en-US.json`)).default;
>   const messages = {...defaultMessages, ...localeMessages};
> 
>   // ...
> });
> ```

Hello and thanks for the reply!

Thanks for the clarification regarding `getMessageFallback`. I've already explained this above, but I'm already aware of merging messages in order to use messages from a different locale as default/fallback. I'm already doing that in my application and it's serving a different purpose.

> As far as I'm aware (as a newbie...), currently the only ways of "falling back" are passing a getMessageFallback to the provider, **_or getting messages from another locale_**.
> The former is too restrictive as it works on a global scale (unless I'm tripping), **_while the latter additionally may not always work (MISSING_MESSAGE for example occurs from the entire messages objects whether you're using a fallback locale or not, so merging messages doesn't do that much here if the message isn't available in other locales)_**.
> 
The main idea here is to be able to have full control over the fallback on an individual message-level. I illustrated this too with an example; if my translations include a key that maps item IDs to names, and the name mapping for a certain ID was not available _in any locale_, merging messages serves no purpose here at all. 

In such a case I may still want to show some placeholder value for the user, which could be "Item {id}". We might also have data elsewhere that at least describes the type of the item if not its name, in which case we could display something more descriptive too like "Item {id} of type {type}". We might even have some sort of formula or dynamic expression that should be used to determine the fallback message before displaying it (such as referring to a completely different message key).

You could certainly hack around this in the way described in the PR with `has`. You could also create a `Fallback` message key in each locale with value interpolation if needed too, and conditionally render it. But as explained that's seems quite finicky if you want to adapt this logic across an entire codebase, with different fallbacks for different places of the code.

I think merging messages is for a different purpose; you'd do that for messages that you might write manually and are more general/critical to the website, such as titles, headers, button text etc. You know for sure you wrote these messages for a given locale, and it makes sense for these types of things to default to a locale if they're not available for another. 
The reason this doesn't apply as well for my purposes is that I have a large amount of automatically processed data with accompanying generated textmaps. Some of it has translations only for a given locale, while other parts do not have any translations at all but still have enough data to warrant showing **something** to the user. You could argue that I can generate the placeholder messages during data processing, but that feels excessive for this use case, especially if I want to easily adjust the placeholder text any time without having to re-generate all the data again, or actually generate a placeholder dynamically based on other site data or user preferences during runtime.

I hope this shows the idea; easily accessible, adjustable, granular control over fallback for messages on an individual-level.

P.S. The entire issue this method resolves essentially just boils down to avoiding the need to write this repeatedly:
```
someName = t.has(`${someId}.name`)
      ? t(`${someId}.name`)
      : `Hello ${someId}`;
```
And just cleanly write
```
t.withFallback("${someId}.name", "Hello ${someId}", false)
```

amannn:
Got it, I understand!

> I think merging messages is for a different purpose; you'd do that for messages that you might write manually and are more general/critical to the website, such as titles, headers, button text etc.

Yeah, I'd say this is generally the type of label that you'd expect to internationalize with `next-intl`.

> You could argue that I can generate the placeholder messages during data processing

Reading your comment, that's also the question that I'd ask here.

For many apps, labels that require internationalization often come from two categories:
1. "Static" app labels for the overall UI
2. Labels for dynamic data, typically from a backend or CMS

As far as I understand your use case, it seems to closely relate to the second category here. I don't know about all the implications of your setup, but generally I'd try to push these kinds of translations in this direction (with for example a backend endpoint having all the flexibility to generate fallbacks etc.).

But apart from that, if your use case really asks for handling these labels with `next-intl`, maybe your workaround is indeed the way to go.

Note that also if this would be supported in core, there would be the same issue around the VSCode integration not picking up `t.withFallback`. 

Sorry for pushing back here a bit, I hope you can understand!

seelepeeku:
> Reading your comment, that's also the question that I'd ask here.
> 
> For many apps, labels that require internationalization often come from two categories:
> 
> 1. "Static" app labels for the overall UI
> 2. Labels for dynamic data, typically from a backend or CMS
> 
> As far as I understand your use case, it seems to closely relate to the second category here. I don't know about all the implications of your setup, but generally I'd try to push these kinds of translations in this direction (with for example a backend endpoint having all the flexibility to generate fallbacks etc.).

My setup doesn't really pull data from a CMS per se. My data is just stored in JSON files locally for now, but I think the same logic would apply even with a dedicated database on SQL or something. The idea of pushing something like generating fallback translations onto the backend doesn't really appeal to me, since I want the backend to focus mostly on more logic and core stuff. It's just unnecessary scattering of responsibilities. 

Going that route is in my opinion also not much different than just implementing a "Fallback" object in the messages and adding keys with value interpolation for every individual message/type of message. It also means that you'd either have two ways of getting translations in your app that you use in separate places and differently (which I don't like -- I want usage to be consistent throughout), or you'd shift entirely to using endpoints to get translations. I really like just getting `t` from `useTranslations` and using it anywhere I need with all the methods, and I want that usage to remain consistent.

And as mentioned, pre-generating fallback stuff during data processing isn't ideal if we want something truly dynamic such as based off user preferences, and easily adjustable without altering/regenerating the data.

> Note that also if this would be supported in core, there would be the same issue around the VSCode integration not picking up `t.withFallback`.

Right. I honestly have no clue how the VSCode integrations work or link into the core, so I just assumed that if it's a method of the `t` object, integrations will just work. My bad! Though, breaking integration with VSCode extensions was honestly just a random thought I wrote on the spot. Even if this makes it in I don't think you'd use this method with a key that extensions will actually be able to pick up even if they work. (Unless this is what you meant?)

> Sorry for pushing back here a bit, I hope you can understand!

No worries! In the end it's just an idea that came to me while I was solving a problem for my app, It might be arguably niche, but I thought others may make use of it as well. Wasn't that hard so thought I'd do it just in case. It's your final call if you don't think this should make it in, and you're 10000% the more experienced developer here compared to me.

If you have anything else to clarify/ask let me know, but otherwise feel free to close this if you'd like. Thanks for the discussion and consideration!

amannn:
Thanks for the additional details!

> you're 10000% the more experienced developer here compared to me

I've learned a ton from the community about i18n while working on this project :).

But I did have many conversations about i18n strategies for projects, and a pattern I've noticed is that if users are looking for dynamic invocations of `t`, often in combination with lists of items, the most robust solutions in this area that I've seen where to usage of some kind of backend internationalization.

It's secondary though if this is an actual backend endpoint, or a service layer in your app, but the overall idea is that translations are maintained along with the data ([example from Sanity](https://www.sanity.io/docs/studio/localization#cb9b506ebe53)).

> Even if this makes it in I don't think you'd use this method with a key that extensions will actually be able to pick up even if they work. (Unless this is what you meant?)

One thing I know from working on OSS is that _if_ an API can theoretically be used in a certain way, then it _will_ eventually be used in that way :). There are also other aspects here, like how would `withFallback` work in combination with `.rich` or `.markup`, as well as other aspects.

I'm generally trying to push more for statically analyzable calls to `t`, as it unlocks further possibilites for users like [validating messages](https://next-intl.dev/docs/workflows/messages) and in the future hopefully #1.

So tldr;, if you really want to use this pattern, I think it's better reserved to be put in your app code currently. Hope that makes somewhat sense!

seelepeeku:
Gotcha. Thanksies and have a good day!

PR #2012, opened 2025-09-05
fix: Fix `usePathname` hydration error with custom prefixes, `localePrefix: 'always'` and static rendering

### LINKED ISSUE #2011: `usePathname` returning wrong value when using custom prefixes and static rendering
### Description

When using the prefixes feature + static rendering the `usePathname` returns the wrong pathname in the initial render.

Related issue https://github.com/vercel/next.js/issues/73085

### Verifications

https://github.com/hugotiger/next-intl-usepathname-hydration-error

### Reproduction description

Steps to reproduce:
1. Generate a production build and run the server (`next build` + `next start`)
2. Check the console and there will be hydration errors

Checking the network tab the initial response seems to contain the "wrong" pathname, aka without the rewrite taken into account.

### Expected behaviour

No hydration errors

hugotiger:
Fixes #2011

This PR introduces a workaround for [vercel/next.js#73085](https://github.com/vercel/next.js/issues/73085).

In Next.js, using `usePathname` in combination with rewrites and middleware can trigger hydration errors. These only occur in production builds, and while they are documented, the information is somewhat hidden:

CleanShot 2025-09-08 at 19 48 12@2x

This issue was partially addressed in [#1571](https://github.com/amannn/next-intl/pull/1571). However, I noticed that when `localePrefix.mode` is set to `'always'`, the workaround was not applied. Instead of adding an extra condition for `'always'`, I flipped the check to `localePrefix.mode !== 'never'`. As far as I can tell this should work fine, but let me know if you see any risks/issues with this.


hugotiger:
@amannn I believe this PR should resolve the issue! Would love your review when you have a chance 😄

packages/next-intl/src/navigation/react-client/useBasePathname.tsx
```diff
@@ -46,7 +46,7 @@ export default function useBasePathname<
     if (isPathnamePrefixed) {
       unlocalizedPathname = unprefixPathname(pathname, prefix);
     } else if (
-      config.localePrefix.mode === 'as-needed' &&
+      config.localePrefix.mode !== 'never' &&
       config.localePrefix.prefixes
     ) {
```
amannn, line 51: There's a chance for a minor simplification here now:

```suggestion
    } else if ('prefixes' in config.localePrefix) {
```

Just to minify a tiny bit better :)

hugotiger: Oh yes that sounds reasonable! But I suppose we still want to check if `prefixes` is truthy? With the `in` operator it'll evaluate to true even when someone passes `undefined` as the value.

So this for example would still run the branch of logic:
```ts
localePrefix: {
  mode: 'always',
  prefixes: undefined,
}
```

So should we still check `config.locale.prefixes` or do you think just checking for the key is fine?

amannn: Ah yes, I think you're right there. It could happen for example if someone conditionally assigns `prefixes` based on an environment parameter.

Checking `config.locale.prefixes` is something TypeScript isn't too happy about though, since we have type-safety for disallowing `prefixes` when `mode: 'never'` is used.

I'd say we leave it like you proposed it originally! 👍

hugotiger: Yeah that's what I was thinking!

Oh yes that's true! I'll leave it as is then :)

amannn:
Hey @hugotiger, many thanks again for your PR here!

I wasn't quite sure where to reach out easiest, so I'll just post it here 😄 

I’m planning to add a few company logos to the next-intl website to highlight adoption, and I was wondering if it would be okay to include Qasa's logo. Of course, I can remove it at any time if you’d prefer not to be listed. Totally fine either way—just let me know!

It would mean a lot and really helps to show the impact next-intl is having in the developer community.

Many thanks in any case!

(I'm also available at jan@amann.work btw.)

hugotiger:
Hey @amannn, sorry the slow response. Had to double check internally that it's fine which it of course is :)

I'll email you our logo assets etc!

Also big thanks from us at Qasa! Both the library itself, as well as your helpful and educational answers in discussions and issues, have been really valuable to us!

PR #2018, opened 2025-09-11
feat: Added support for default formatters

valeryq:
## Add support for default formatters

This PR introduces automatic default formatter handling for ICU message format placeholders and function-based formatters that don't specify a format.

### Changes
**Automatic default format injection:** Messages with placeholders like `{value, number}` are automatically transformed to `{value, number, default}` before parsing.
**Function formatter support:** The `useFormatter` hook and `createFormatter` function now apply default formats when no specific format is provided.
**Generic type support:** Works with any ICU message format or function-based fromat type (number, date, time, plural, etc.)
**Unified configuration:** Both ICU types and function formatters can define default formats in the IntlProvider's formats prop

```ts
// Before: Required explicit format specification
<IntlProvider
  formats={{
    number: {
      default: { useGrouping: false },        // Applied to both ICU and functions
      currency: { style: 'currency', currency: 'USD' }
    },
    dateTime: {
      default: { year: 'numeric' },          // Applied to both ICU and functions
      full: { dateStyle: 'full' }
    }
  }}
>
  {/* ICU message formatting with defaults */}
  {t('price: {value, number}', {value: 10000})}  {/* → "price: 10000" */}
  
  {/* Function formatting with defaults */}
  {format.number(10000)}                         {/* → "10000" */}
  {format.dateTime(new Date())}                  {/* → "2024" */}
</IntlProvider>
```

### Implementation
The transformation is handled by a regex pattern `/\{([^,}]+),\s*(\w+)\s*}/g` that:
- Matches ICU placeholders with exactly 2 parts (variable, type)
- Transforms them to 3 parts (variable, type, default)
- Preserves existing 3-part placeholders unchanged
- Function formatter defaults: Both `useFormatter` and `createFormatter` check for default format configurations when no explicit format is provided

### DIFF
--- a/packages/use-intl/src/core/createBaseTranslator.tsx
+++ b/packages/use-intl/src/core/createBaseTranslator.tsx
@@ -77,6 +77,16 @@ function resolvePath(
   return message;
 }
 
+// Prepare message by converting `{ var, type }` to `{ var, type, default }`
+// to work around absence of default in ICU message syntax
+function prepareMessage(message: string) {
+  return message
+    .replace(
+      /\{\s*(\w+),\s*(\w+)\s*}/g,
+      '{$1, $2, default}'
+    );
+}
+
 function prepareTranslationValues(values: RichTranslationValues) {
   // Workaround for https://github.com/formatjs/formatjs/issues/1467
   const transformedValues: RichTranslationValues = {};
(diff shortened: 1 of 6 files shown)


valeryq:
The implementation doesn't fit well with the roadmap of `next-intl`.

PR #2026, opened 2025-09-16
docs: Loccy IDE extension

dipuzyrev:
Hi there,

I'm excited to announce that Loccy, a modern localization tool with a focus on AI, now supports React and has added ⁠`next-intl` as one of its first supported React i18n frameworks.

Loccy is similar to ⁠i18n-ally but is actively maintained and built with AI use-cases in mind from the ground up. I've taken the liberty of placing it at the top of the list to reflect its active development status.

Thank you for your consideration!


amannn:
Hey @dipuzyrev,

congrats on the launch of Loccy, also glad to hear that it supports `next-intl`!

The features from Loccy are cool, I've tested the extension in Cursor and it mostly worked well for me.

Some points I found:
- When changing a message inline, I noticed the translation in the popup isn't updated.
- After renaming a key, the namespace (from `useTranslations(…)`) is copied into the key.

Screenshot 2025-09-18 at 16 04 42

> I've taken the liberty of placing it at the top of the list to reflect its active development status.

I understand that you have interest in promoting the extension.

As of today, i18n Ally is by far the most popular extension that supports `next-intl` and is valued by many developers (including myself). I know it requires more maintenance, but I've heard that there's interest in revamping its development. And while there's some overlap, it currently also has features that Loccy doesn't have.

Another point is that i18n Ally is completely free and open source (like `next-intl`).

I have to say, this overlay made me reconsider if I really want to proceed:

Screenshot 2025-09-18 at 16 00 14

Not sure if it was a consideration of you to make it open source? It was honestly the first time I've seen a non-open source extension for VSCode. As editor extensions can read secrets from `.env` files and more, at least from my perspective, being open source certainly helps with trust.

So for the time being, I'm fine with including Loccy in the docs, but I'll place it at as the second option on the page. I hope you can understand. I'd also prefer to keep a single video on the page for now as the videos use different editor styles etc.—also the usage of multiple calls to `useTranslations` is not something I advocate for.

So I've updated the page a bit to reflect this, but other than that, we're good to go!

amannn:
Somehow GitHub automatically closed the PR. I've opened a new one here: https://github.com/amannn/next-intl/pull/2031

dimchy:
Hi @amannn,

Thanks for integrating Loccy and for providing such great feedback!

Regarding your points:

> When changing a message inline, I noticed the translation in the popup isn't updated.

Unfortunately, this is a limitation of the VSCode API. To see the changes, the hover message needs to be reopened after an inline edit.

> After renaming a key, the namespace (from `useTranslations(…)`) is copied into the key.

You're right, that shouldn't be happening. I'll push a fix for this soon.

On the `Do you trust the publisher "loccy"?` prompt: Loccy is already verified in the Visual Studio Marketplace. I'm currently working on extending that verification to Cursor and other editors.

I appreciate your perspective on open-sourcing to foster trust. Totally valid point, I will definitely think it over.

PR #2033, opened 2025-09-18
docs: Update static rendering instructions for locale-based routing

ceolinwill:
The docs suggest using `generateStaticParams` for all locales but this can slow down builds, especially if you have lots of locales and routes.

I'm updating that section to mention it's possible to return an empty array or use `export const dynamic = 'force-static'` and the page will be generated on the first visit without becoming dynamic.

### DIFF
--- a/docs/src/pages/docs/routing/setup.mdx
+++ b/docs/src/pages/docs/routing/setup.mdx
@@ -175,7 +175,7 @@ When using locale-based routing, `next-intl` will currently opt into dynamic ren
 
 ### Add `generateStaticParams`
 
-Since we are using a dynamic route segment for the `[locale]` param, we need to pass all possible values to Next.js via [`generateStaticParams`](https://nextjs.org/docs/app/api-reference/functions/generate-static-params) so that the routes can be rendered at build time.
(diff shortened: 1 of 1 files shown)


amannn:
Hey, thanks for proposing this docs update!

From my experience, the bulk of users support rather a few locales and even for a larger number of locales, I can imagine that many people will prefer slightly longer builds instead of slower production performance.

So I'd make sure the docs focus on the majority here without presenting too many options upfront. But I agree there's surely a tradeoff here.

What we could do is instead of making this choice this prominent, we could add an expandable section to [step 1](https://next-intl.dev/docs/routing/setup#add-generatestaticparams):

```tsx
<Details id="generatestaticparams-locales">
<summary>Do I need to return all locales from `generateStaticParams`?</summary>

If you prefer to only render certain locales at build time, or none at all, you can be selective about the ones you return from `generateStaticParams` (see [Static Rendering](https://nextjs.org/docs/app/api-reference/functions/generate-static-params#static-rendering) in the Next.js docs).

</Details>
```

I think that gets the point across, and we don't have repeat too much of the Next.js docs here. Rendering only your most popular locales at build time could also be interesting use case I guess.

Does that sound right to you?

ceolinwill:
Makes sense.

I’ve updated the PR to include your suggestion. I also adjusted the wording from “we need to pass all possible values” to “we need to use generateStaticParams” since it’s not always required to pass all values, and that phrasing could be misleading.

Feel free to make any other changes you see fit.

PR #2043, opened 2025-10-06
docs: fix Vercel URL, add 2nd way to get user's time zone

karlhorky:
Hi @amannn, hope everything's good with you! 👋

Quick PR to:

1. Fix the Vercel URL
2. Add a brief mention of a secondary method to retrieve the user's time zone information

### DIFF
--- a/docs/src/pages/blog/date-formatting-nextjs.mdx
+++ b/docs/src/pages/blog/date-formatting-nextjs.mdx
@@ -223,10 +223,10 @@ export default function BlogPostPublishedDate({published, timeZone}: Props) {
 
 Sticking to a single time zone for your app is definitely the easiest solution here. However, in case you'd like to format dates in the user's time zone, a reasonable approach might require having the time zone available on the server side so that it can be used in server-only code.
 
-As browsers don't include the time zone of the user in an HTTP request, two ways to retrieve the user's time zone:
+As browsers don't include the time zone of the user in an HTTP request, there are two primary ways to retrieve it on the server side:
 
-1. Pass the user's preferences from the client to the server using a cookie or [the client hints approach](https://github.com/rphlmr/nextjs-client-hints)
-2. Use approximate geographical information from the user's IP address, such as the [`x-vercel-ip-timezone`](https://vercel.com/docs/headers/request-headers#x-vercel-ip-timezone) request header if you're running your app on Vercel. However, it's important to note that this is only an approximation, so letting the user choose their time zone explicitly might still be sensible.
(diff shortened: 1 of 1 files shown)


docs/src/pages/blog/date-formatting-nextjs.mdx
```diff
@@ -223,7 +223,10 @@ export default function BlogPostPublishedDate({published, timeZone}: Props) {
...
+As browsers don't include the time zone of the user in an HTTP request, two ways to retrieve the user's time zone:
+
+1. Pass the user's preferences from the client to the server using a cookie or [the client hints approach](https://github.com/rphlmr/nextjs-client-hints)
+2. Use approximate geographical information from the user's IP address, such as the [`x-vercel-ip-timezone`](https://vercel.com/docs/headers/request-headers#x-vercel-ip-timezone) request header if you're running your app on Vercel. However, it's important to note that this is only an approximation, so letting the user choose their time zone explicitly might still be sensible.
```
amannn: Hey @karlhorky, thanks a lot for proposing this update!

I think it's a good point to include reading from cookies here. I think what's missing a bit in the linked GitHub repo is a discussion around what the perf impact of this client hints pattern is. I'm not sure if it can really be seen as a general recommendation, so I think I'd leave it more as a general mention of cookies:

```suggestion
As browsers don't include the time zone of the user in an HTTP request, there are two primary ways to retrieve it on the server side:

1. Retrieve it from [`cookies()`](https://nextjs.org/docs/app/api-reference/functions/cookies), if it was set previously
2. Use approximate geographical information from the user's IP address (e.g. [`x-vercel-ip-timezone`](https://vercel.com/docs/headers/request-headers#x-vercel-ip-timezone)). However, it's important to note that this is only an approximation.
```

Does that look good to you?

Thanks also for fixing the Vercel link! 🙏

karlhorky: Ok, how about like this? A bit more detail about how that cookie would possibly contain the time zone:

```suggestion
As browsers don't include the time zone of the user in an HTTP request, there are two primary ways to retrieve it on the server side:

1. Set a cookie with the browser time zone on the client and retrieve it on the server side using [`cookies()`](https://nextjs.org/docs/app/api-reference/functions/cookies)
2. Use approximate geographical information from the user's IP address (e.g. [`x-vercel-ip-timezone`](https://vercel.com/docs/headers/request-headers#x-vercel-ip-timezone)). However, it's important to note that this is only an approximation.
```

amannn: Yep, that's a good idea—thanks!

karlhorky:
Thanks for the review and merge!

PR #2051, opened 2025-10-07
fix: Only call `NextResponse.rewrite` if really necessary

amannn:
Previously, we always called `.rewrite` just in case since it did no harm, but it seems like this messes with Vercel observability tools. This cleans it up to only call this function when really necessary.

### DIFF
--- a/packages/next-intl/.size-limit.ts
+++ b/packages/next-intl/.size-limit.ts
@@ -42,7 +42,7 @@ const config: SizeLimitConfig = [
   {
     name: "import * from 'next-intl/middleware'",
     path: 'dist/esm/production/middleware.js',
-    limit: '9.695 KB'
+    limit: '9.7 KB'
   },
   {
     name: "import * from 'next-intl/routing'",
--- a/packages/next-intl/src/middleware/middleware.test.tsx
+++ b/packages/next-intl/src/middleware/middleware.test.tsx
@@ -257,42 +257,30 @@ describe('prefix-based routing', () => {
 
     it('serves requests for other locales when prefixed', () => {
       middleware(createMockRequest('/de'));
-      expect(MockedNextResponse.next).not.toHaveBeenCalled();
+      expect(MockedNextResponse.next).toHaveBeenCalled();
       expect(MockedNextResponse.redirect).not.toHaveBeenCalled();
-      expect(MockedNextResponse.rewrite).toHaveBeenCalled();
-      expect(MockedNextResponse.rewrite.mock.calls[0][0].toString()).toBe(
-        'http://localhost:3000/de'
-      );
+      expect(MockedNextResponse.rewrite).not.toHaveBeenCalled();
     });
 
     it('serves requests for other locales when prefixed with a trailing slash', () => {
       middleware(createMockRequest('/de/'));
-      expect(MockedNextResponse.next).not.toHaveBeenCalled();
       expect(MockedNextResponse.redirect).not.toHaveBeenCalled();
-      expect(MockedNextResponse.rewrite).toHaveBeenCalled();
-      expect(MockedNextResponse.rewrite.mock.calls[0][0].toString()).toBe(
-        'http://localhost:3000/de'
-      );
+      expect(MockedNextResponse.rewrite).not.toHaveBeenCalled();
+      expect(MockedNextResponse.next).toHaveBeenCalled();
     });
 
     it('serves requests for other locales with query params at the root', () => {
       middleware(createMockRequest('/de?sort=asc'));
-      expect(MockedNextResponse.next).not.toHaveBeenCalled();
+      expect(MockedNextResponse.next).toHaveBeenCalled();
       expect(MockedNextResponse.redirect).not.toHaveBeenCalled();
-      expect(MockedNextResponse.rewrite).toHaveBeenCalled();
-      expect(MockedNextResponse.rewrite.mock.calls[0][0].toString()).toBe(
-        'http://localhost:3000/de?sort=asc'
-      );
+      expect(MockedNextResponse.rewrite).not.toHaveBeenCalled();
     });
 
     it('serves requests for other locales with query params at a nested path', () => {
       middleware(createMockRequest('/de/list?sort=asc'));
-      expect(MockedNextResponse.next).not.toHaveBeenCalled();
+      expect(MockedNextResponse.next).toHaveBeenCalled();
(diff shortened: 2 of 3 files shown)


roelvan:
@amannn I am getting 404's after this feature's been added ("next-intl": "4.3.11"). Here is my middleware:
```js
export default async function middleware(req) {
  // console.log('Middleware processing:', req.nextUrl.pathname);
  const hostname = req.headers.get('host');
  const countryCode = getCountryCodeByDomainName(hostname);
  const [, locale, ...remainingPath] = req.nextUrl.pathname.split('/');
  let pathnameToRewrite = remainingPath.join('/');
  const context = {
    locale,
    hostname,
    pathname: remainingPath.join('/'),
    searchParams: req.nextUrl.searchParams.toString(),
  };
  let result;

  // redirect shortlinks (no locale needed)
  result = shortLinkRedirectMiddleware(req, { hostname });
  if (result) return result();

  // we need a valid locale before running these middleware functions
  if (localeRegex.test(locale)) {
    result = batchRunMiddleware(
      [
        localeRedirectMiddleware,
        countryCodeRedirectMiddleware,
        legacyB2bCitySlugRedirectMiddleware,
        locationHelpPageRedirectMiddleware,
        legacyTeamLobbyRedirectMiddleware,
      ],
      req,
      context,
    );
    if (result) return result();

    // rewrite country code
    // setting the nextUrl.pathname will pass it to the handleI18nRouting() rewriter
    req.nextUrl.pathname = `/${locale}/${countryCode}/${pathnameToRewrite}`;
  }

  const handleI18nRouting = createIntlMiddleware(middlewareRouting);
  return handleI18nRouting(req);
}
```
Is there a way to trigger the rewrite anyway? Or am I doing something the wrong way?

amannn:
Hey, sorry this caused a bug in your app! And yes, you can adapt your logic to call `NextResponse.rewrite(…)` yourself after the `next-intl` middleware has run. The [additional rewrites example](https://next-intl.dev/docs/routing/middleware#example-additional-rewrites) in the docs might help for inspiration.

roelvan:
> Hey, sorry this caused a bug in your app! And yes, you can adapt your logic to call `NextResponse.rewrite(…)` yourself after the `next-intl` middleware has run. The [additional rewrites example](https://next-intl.dev/docs/routing/middleware#example-additional-rewrites) in the docs might help for inspiration.

Awesome! Didn't catch that updated example. In case other people would ever need this, this is my updated working code:
```js
const handleI18nRouting = createIntlMiddleware(routing);

// test redirects and rewrites with curl:
// curl -s -I -L http://localhost:3000/abcd
export default async function middleware(req) {
  let response = handleI18nRouting(req);

  if (!response.ok) return response; // first let next-intl do its locale redirects

  const [, locale, ...rest] = new URL(response.headers.get('x-middleware-rewrite') || req.url).pathname.split('/');
  const hostname = req.headers.get('host');
  const pathname = `/${rest.join('/')}`;
  const searchParams = req.nextUrl.searchParams.toString();
  const countryCode = getCountryCodeByDomainName(hostname);
  const context = { req, locale, pathname, searchParams, countryCode, hostname };

  if (IS_DEV_ENV) console.log('🚏 Middleware:', { locale, pathname, searchParams, countryCode, hostname });

  // the order of middleware is important
  // as soon as one returns a function, the other ones will not execute anymore
  const middlewareResponse = batchRunMiddleware(
    [
      shortLinkRedirectMiddleware,
      localeRedirectMiddleware,
      countryCodeRedirectMiddleware,
      legacyB2bCitySlugRedirectMiddleware,
      locationHelpPageRedirectMiddleware,
      legacyTeamLobbyRedirectMiddleware,
    ],
    context,
  );
  if (middlewareResponse) return middlewareResponse();

  // add oure countryCode rewrite
  response = NextResponse.rewrite(
    new URL(`/${locale}/${countryCode}${pathname}${searchParams.length > 0 ? `?${searchParams}` : ''}`, req.url),
    {
      headers: response.headers,
    },
  );

  return response;
}

```

evilive3000:
## Issue: Custom middleware breaks with 404 after this PR

  After upgrading from `4.3.10` to `4.3.11`, our custom middleware that adds currency segments to URLs started returning 404 errors.

  ### Setup
  We use a custom middleware wrapper around `next-intl` to inject currency into the URL:
  - Pattern: `[locale]/[currency]` (e.g., `/de/EUR`, `/ja/JPY`)
  - The middleware modifies `x-middleware-rewrite` header set by `next-intl`

  ### Problem
  Our code relied on `x-middleware-rewrite` always being set:
  ```typescript
  const rewriteUrl = response.headers.get("x-middleware-rewrite");
  if (rewriteUrl) {
    // Add currency to URL
  }
```
  After this PR, `x-middleware-rewrite` is null for simple locale-only requests like `/de`, causing 404s.

 ### Solution

  Following the [additional rewrites](https://next-intl.dev/docs/routing/middleware#example-additional-rewrites) pattern from the docs:
```typescript
if (response.ok) {
  const rewriteUrl = response.headers.get("x-middleware-rewrite") || req.url;
  const updatedUrl = insertCurrency(rewriteUrl, currency);

  return NextResponse.rewrite(new URL(updatedUrl), {
    headers: response.headers,
  });
}
```
 ### Suggestion

  Consider adding a migration note in the changelog about this breaking change for custom middleware that relies on `x-middleware-rewrite` always being present.

amannn:
Hey @evilive3000, thanks for sharing your solution! And yes, that's a good point, I gave this more visibility in the [release notes](https://github.com/amannn/next-intl/releases/tag/v4.3.11) and this PR description now!

ghost:
@amannn this has caused breaking changes across a number of projects for us. This being a patch version it took us a while to determine why our projects were suddenly showing a 404. I understand this is technically marked as a fix but should this change have been considered an opt-in solution until v5?

amannn:
@riley-pearce-airteam My apologies! In which way have you relied on the `x-middleware-rewrite` response header?

I have to admit this change caused more ripples than I had anticipated, I'm sorry about that. It's unfortunately also hard to revert now as people rely on the fixed behavior and the previous behavior had issues.

ghost:
@amannn no worries at all, in that case I'll go through and determine the necessary fixes on my end.

I use middleware slightly differently to your migration examples so once I've got it worked out I'll share an example here for others who might need it.

Looking at the code change you made I can completely understand why you didn't anticipate this, I don't believe I would have either!

Thanks for all your work on next-intl.

PR #1632, opened 2024-12-20
fix: Deprecate `setRequestLocale` and add blog post on `next/root-params`

amannn:
Related to https://github.com/vercel/next.js/pull/72837 and https://github.com/vercel/next.js/pull/80255

[→ Draft](https://next-intl-docs-git-docs-v4-rootparams-blog-post-next-intl.vercel.app/blog/nextjs-root-params)

**TODO**
- [x] Deprecate `setRequestLocale`?
- [x] Wait until `next/root-params` lands in Next.js stable (https://github.com/vercel/next.js/pull/93764)
- [x] Update publishing date
- [x] Update initial link when there are official docs
- [ ] https://github.com/amannn/next-intl/pull/2327

Closes https://github.com/amannn/next-intl/issues/663
Closes https://github.com/amannn/next-intl/issues/2075

PR #2084, opened 2025-11-08
feat: playground

### LINKED ISSUE #1623: Refactor playground to be more useful

Currently, https://github.com/amannn/next-intl/tree/main/examples/example-app-router-playground is used for e2e testing.

That's absolutely fine, but I think if we'd refactor it to something like https://app-router.vercel.app it could also be helpful to document more advanced use cases.

Refactor the playground example to:
1. Look nicer
2. Keep all existing tests
3. Group tests logically into different sections of the app (i.e. no home page that demonstrates all kinds of wild patterns)

-

Adebesin-Cell:
Closes #1623.

Reboot of the playground refactor with the v1 scope agreed in https://github.com/amannn/next-intl/issues/1623#issuecomment-3425567893: a single category (Translations) with two detail pages (Server / Client Components), MDX-driven content via Code Hike, locale-prefixed routing, and the labeled-boundary visual identity from the Vercel app-router playground applied to next-intl's design system.

## What's in this PR

- New `playground/` workspace (Next 15.5.15, React 19, Tailwind v4, shadcn primitives)
- `next-intl` middleware + `[locale]` routing with `en` / `de` messages
- `@next/mdx` + Code Hike pipeline (`remarkCodeHike` + `recmaCodeHike`) wired in `next.config.ts`, with a custom `<Code>` RSC and 6 annotation handlers (mark, callout, focus, link, fold, line-numbers)
- Sidebar / header / locale-switcher / theme-toggle chrome adapted from the design system spec — left-aligned brand, icon-prefixed section labels, quiet active-state accent, `bg-background` corner labels punching through hairline borders
- Two detail pages backed by `content.mdx` + a sibling live demo component:
  - **Translations / Server components** — async demo using `getTranslations` + `setRequestLocale`
  - **Translations / Client components** — `useTranslations` with an interactive ICU `{name}` interpolation
- Each page wraps content in `<PlaygroundBoundary label="Demo">` + a separate `<PlaygroundBoundary label="Output" variant="dotgrid">` so the rendered demo reads like a render preview
- Empty future categories (Formatting, Routing, Patterns) are reserved in the sidebar with a dimmed `coming soon` placeholder so adding them later is just a data change in `src/lib/nav.ts`
- Polished byline footer with Lucide-iconed source/docs/inspiration links
- Smoke Playwright tests covering landing + both detail pages in both locales
- Existing `examples/example-app-router-playground` is untouched — e2e coverage stays where it lives until the new playground absorbs it in a follow-up

## Screenshots

<details>
<summary>Landing — dark</summary>

01-landing-en-dark

</details>

<details>
<summary>Landing — light</summary>

02-landing-en-light

</details>

<details>
<summary>Client Components — dark</summary>

05-client-en-dark

</details>

<details>
<summary>Client Components — light</summary>

06-client-en-light

</details>

## Out of scope (explicit)

- Other categories (Formatting, Routing, Patterns) and their pages
- An `intl-explorer`-style interactive formatting playground
- Code Hike `tabs` / `CodeWithNotes` (no v1 page needs them)
- Migrating Playwright e2e off `examples/example-app-router-playground`

## Try it

```bash
pnpm install
pnpm --filter playground dev
```

Open http://localhost:3000 — middleware redirects to `/en`, then click into either Translations page. Locale switcher (top-right) flips to `/de`, theme toggle persists.

## Notes for review

- Turbopack is disabled for the playground because `@next/mdx`'s loader options aren't serializable for the current Turbopack pipeline (`pnpm dev`/`build` use webpack)
- Two `as any` casts in `next.config.ts` work around a peer-resolved Next 14 / 15 NextConfig type clash from `@next/mdx`
- A future iteration can hand-write per-page `README.md` files for GitHub-side browsing; currently the README slot just points at `content.mdx`


amannn:
Nice, let me know once it's ready to have a look!

amannn:
Oh, just one thing after having a quick look: The _patterns_ example I've set up was intended to _be_ the playground. You can just move all your code there! Let's call it "patterns" for now since there is already example-app-router-playground.

Adebesin-Cell:
Hey @amannn 👋 sorry for the long radio silence on #2084, but I'm back on it. I just pushed a full reboot.

What’s there now (matches the v1 scope you sketched in issue #1623 https://github.com/amannn/next-intl/issues/1623#issuecomment-3425567893):

* Translations category with Server Components + Client Components pages
* Locale-prefixed routing (`/[locale]/...`) with `next-intl/middleware`, en + de
* MDX content authored as `content.mdx` per page, rendered via Code Hike (`remarkCodeHike` + `recmaCodeHike`) with annotation handlers (mark, callout, focus, link, fold, line-numbers)
* Vercel-app-router-playground visual identity adapted to your design system: hairline borders with mono corner labels, sidebar nav with Lucide section icons + quiet active-state accent
* Output preview boxes use a subtle dot-grid backdrop, so the rendered greeting reads like a preview
* Lucide icons throughout, blue-300 logo on dark / blue-700 on light
* Existing `examples/example-app-router-playground` is untouched — Playwright coverage stays where it lives

Screenshots attached: landing (en/de, light/dark), both detail pages in both locales, mobile.

Ideas for next iteration (would love your read before scoping):

1. Formatting category — small intl playground (date + number for v1), live preview + copyable `next-intl` config
2. Routing category — mixed routing example (like https://next-intl.dev/examples#app-router-mixed-routing), showing locale-prefixed vs non-prefixed boundary
3. Patterns category — locale switcher, Server Actions, metadata, error files (each as its own page)
4. Auto-generate per-page `README.md` from `content.mdx` to avoid drift
5. Eventually move Playwright suite into the playground so it doubles as docs + E2E

No rush at all, happy to split any of these into separate PRs once v1 feels good 🙏

amannn:
Awesome, thanks for your hard work here @Adebesin-Cell! I'll make a note of this and will get back to you!

Adebesin-Cell:
Hi @amannn,

I picked up the work from [comment](https://github.com/amannn/next-intl/pull/2084#issuecomment-4406623441), was curious how those ideas would shape up. Attached a video so you can see how it looks.

Also did a quick a11y pass:

* `aria-pressed` on the chips
* `aria-live` on the Explorer Output so screen readers announce format changes
* proper `aria-labels' on the icon-only buttons
* `aria-current` on the active locale

Had fun doing this 😅

I’m also thinking of an accessibility/preferences panel similar to how [Intl Explorer Playground](https://intl-explorer.com/Playground) does theirs, toggles for reduced motion, muting the Explorer’s screen-reader announcements, accent color, etc

I have some ideas and would love to include them in the playground as well.

https://github.com/user-attachments/assets/ba1a6d13-9eff-402d-816b-66e729bb02a8

amannn:
Oh wow @Adebesin-Cell, that's awesome! 😮🔥

I'll get back to you with some more detailed feedback!

amannn:
Hey @Adebesin-Cell, sorry, I still haven't found the time to dig more into this.

But two things:
1. https://github.com/amannn/next-intl/pull/2084#issuecomment-3510955802
2. Can you split the PR, so there's an initial one with only the translations section with pages Server Components and then Client Components?

It'd be good to make sure the foundation and direction is solid first. Then, we can look into adding more pages.

But the integration of the `Intl` explorer is a really cool idea, I think we should add this along with the formatter pages in follow-up PRs!

Adebesin-Cell:
No worries @amannn 🙏 Did the split — this PR is now scoped to just the Translations foundation.

Moved everything else into a Draft follow-up: #2337. It stacks on the same base and contains:

- Formatting → Dates / Numbers / **Explorer** (the `useFormatter` integration covering all five methods + named formats registry + Code Hike snippet)
- Routing → Mixed routing
- Patterns → Locale switcher, Browser language, Cache components placeholder
- Settings dialog (reduce motion, announce output, accent palette) — persisted to `localStorage`
- Localized `/not-found` with a small `useFormatter`/`useTranslations` mini-playground
- Root-layout refactor so the 404 renders correctly
- Accessibility pass (`aria-pressed`, `aria-live`, `aria-current`, etc.)

playground/src/app/layout.tsx
```diff
@@ -0,0 +1,5 @@
+import type { ReactNode } from 'react';
+
+export default function RootLayout({ children }: { children: ReactNode }) {
+  return children;
+}
```
amannn: The playground should be the `example-app-router-patterns` that I've set up earlier (see https://github.com/amannn/next-intl/pull/2084#issuecomment-3510955802)

playground/src/i18n/request.ts
```diff
@@ -0,0 +1,15 @@
+import { getRequestConfig } from 'next-intl/server';
+import { hasLocale } from 'next-intl';
+import { routing } from './routing';
+
+export default getRequestConfig(async ({ requestLocale }) => {
```
amannn: Next.js 16.3 is on the horizon and is expected to ship with `next/root-params`. We should adjust for this once the release is out (related to https://github.com/amannn/next-intl/pull/1632 and https://github.com/amannn/next-intl/pull/2327)

playground/public/file.svg
amannn, line 1: Can we use the lucide icon library instead of providing custom icons?

playground/src/assets/logo.tsx
```diff
@@ -0,0 +1,70 @@
+
+export function Logo({ className = "w-12 h-12" }: { className?: string }) {
```
amannn: The logo appears cut off at the top and has this bubble at the bottom right end:

Image

Can you make sure this matches the docs?

playground/tests/playground.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+});
+
+test('Client Components page renders demo and reacts to input', async ({
+  page,
+}) => {
+  await page.goto('/en/translations/client-components');
+  const input = page.getByPlaceholder('Frodo');
+  await input.fill('Sam');
+  await expect(page.getByText('Hello, Sam!')).toBeVisible();
+});
```
amannn: I think we can even remove playwright tests for now to simplify this.

I've recently introduced an `/e2e` folder on `main` that I'll continue to use for e2e testing, I think we don't need to use the playground for e2e tests going forward.

playground/src/app/[locale]/translations/server-components/page.tsx
```diff
@@ -0,0 +1,45 @@
...
+      </PlaygroundBoundary>
+
+      <PlaygroundBoundary label="Output" variant="dotgrid">
+        <div className="flex min-h-[140px] items-center justify-center">
+          <ServerExample />
+        </div>
+      </PlaygroundBoundary>
+
+      <div className="!mt-6 flex justify-end pt-4">
+        <GitHubLink path="playground/src/app/[locale]/translations/server-components" />
```
Comment (amannn) [inline playground/src/app/[locale]/translations/server-components/page.tsx]: I think `[locale]` needs to be encoded (`%5Blocale%5D`), can you verify these links work?

playground/src/app/globals.css
```diff
@@ -0,0 +1,290 @@
...
+  --color-sidebar-primary-foreground: var(--sidebar-primary-foreground);
+  --color-sidebar-accent: var(--sidebar-accent);
+  --color-sidebar-accent-foreground: var(--sidebar-accent-foreground);
+  --color-sidebar-border: var(--sidebar-border);
+  --color-sidebar-ring: var(--sidebar-ring); /* Charts */
+  --color-chart-1: var(--chart-1);
+  --color-chart-2: var(--chart-2);
+  --color-chart-3: var(--chart-3);
+  --color-chart-4: var(--chart-4);
+  --color-chart-5: var(--chart-5); /* Gray scale */
```
amannn: Charts? Is this used somewhere?

playground/src/app/globals.css
```diff
@@ -0,0 +1,290 @@
...
+  --color-gray-200: var(--gray-200);
+  --color-gray-300: var(--gray-300);
+  --color-gray-400: var(--gray-400);
+  --color-gray-500: var(--gray-500);
+  --color-gray-600: var(--gray-600);
+  --color-gray-700: var(--gray-700);
+  --color-gray-800: var(--gray-800);
+  --color-gray-900: var(--gray-900); /* Blues */
+  --color-blue-300: var(--blue-300);
+  --color-blue-700: var(--blue-700); /* White */
```
amannn: Can we avoid this mapping, e.g. by managing our colours as `--color-*` directly?

playground/src/app/globals.css
```diff
@@ -0,0 +1,290 @@
...
+}
+@layer base {
+  button:not([disabled]),
+  [role='button']:not([disabled]) {
+    cursor: pointer;
+  }
+}
+
+/* Subtle dot-grid backdrop for the "output" boundary box */
+.dotgrid {
```
amannn: Maybe we can move this to a component instead to share this style via an import instead of a global class name?

playground/src/app/[locale]/layout.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { setRequestLocale, getMessages } from 'next-intl/server';
+import { notFound } from 'next/navigation';
+import { routing } from '@/i18n/routing';
+import { ClientProviders } from '@/components/playground/client-providers';
+import { PlaygroundSidebar } from '@/components/playground/sidebar';
+import { PlaygroundByline } from '@/components/playground/byline';
+import '../globals.css';
+
+const geistSans = Geist({ variable: '--font-geist-sans', subsets: ['latin'] });
+const geistMono = Geist_Mono({ variable: '--font-geist-mono', subsets: ['latin'] });
```
Comment (amannn) [inline playground/src/app/[locale]/layout.tsx]: Can we use Inter instead like the docs?

playground/src/app/[locale]/layout.tsx
```diff
@@ -0,0 +1,57 @@
...
+  children,
+  params,
+}: {
+  children: React.ReactNode;
+  params: Promise<{ locale: string }>;
+}) {
+  const { locale } = await params;
+  if (!hasLocale(routing.locales, locale)) notFound();
+
+  setRequestLocale(locale);
```
Comment (amannn) [inline playground/src/app/[locale]/layout.tsx]: This should also be gone/adapted once we switch to next/root-params.

playground/src/app/[locale]/layout.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const messages = await getMessages();
+
+  return (
+    <html lang={locale} suppressHydrationWarning>
```
Comment (amannn) [inline playground/src/app/[locale]/layout.tsx]: Why do we need to suppress hydration warnings?

playground/src/app/[locale]/layout.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <body
+        className={`font-sans ${geistSans.variable} ${geistMono.variable} antialiased`}
+      >
+        <NextIntlClientProvider locale={locale} messages={messages}>
+          <ClientProviders>
```
Comment (amannn) [inline playground/src/app/[locale]/layout.tsx]: This only contains `ThemeProvider`, can we render this here directly?

playground/src/app/[locale]/page.tsx
```diff
@@ -0,0 +1,73 @@
+import { setRequestLocale } from 'next-intl/server';
+import { useTranslations } from 'next-intl';
+import { ArrowRight } from 'lucide-react';
+import { Link } from '@/i18n/navigation';
+import { sections } from '@/lib/nav';
```
Comment (amannn) [inline playground/src/app/[locale]/page.tsx]: By using this pattern, we can't translate the navigation. Can we move this inline?

playground/src/app/[locale]/translations/server-components/content.mdx
```diff
@@ -0,0 +1,26 @@
+import {TwoColumn} from '@/components/playground/two-column';
+
+<TwoColumn>
```
Comment (amannn) [inline playground/src/app/[locale]/translations/server-components/content.mdx]: This cuts off important parts of the code snippet, maybe it's better to switch to a single column.

playground/src/app/[locale]/translations/server-components/content.mdx
```diff
@@ -0,0 +1,26 @@
...
+
+- In async Server Components, call `getTranslations()` from `next-intl/server` to read translated strings.
+- Messages stay on the server — no client-side JavaScript is shipped for the translation runtime.
+- Pair with `setRequestLocale()` near the top of the component so static rendering picks the right locale.
```
Comment (amannn) [inline playground/src/app/[locale]/translations/server-components/content.mdx]: Hmm, I know we previously discussed using markdown for this, but in the case of next-intl, it's a bit unfortunate that this will make it harder to translate.

What do you think of reverting the markdown focus of the playground, and instead rendering relevant labels directly in `page.tsx` so that we can call them with `t(…)`? In that case, we could remove `content.mdx` along with `README.md` and like to `page.tsx` instead.

playground/src/app/[locale]/translations/server-components/content.mdx
```diff
@@ -0,0 +1,26 @@
...
+```tsx app/page.tsx
+import {getTranslations, setRequestLocale} from 'next-intl/server';
+
+export default async function Page({
+  params
+}: {
+  params: Promise<{locale: string}>;
+}) {
+  const {locale} = await params;
+  setRequestLocale(locale);
```
Comment (amannn) [inline playground/src/app/[locale]/translations/server-components/content.mdx]: Let's avoid any `setRequestLocale` in snippets, it will be gone soon. Also the `params` add noise here, let's make this as minimal as possible.

playground/src/app/[locale]/translations/server-components/content.mdx
```diff
@@ -0,0 +1,26 @@
...
+
+  // !mark
+  const t = await getTranslations('ServerDemo');
+  return <h1>{t('greeting')}</h1>;
+}
+```
```
Comment (amannn) [inline playground/src/app/[locale]/translations/server-components/content.mdx]: We should also mention non-async components on this page.

playground/src/app/[locale]/translations/server-components/page.tsx
```diff
@@ -0,0 +1,45 @@
...
+  params,
+}: {
+  params: Promise<{ locale: string }>;
+}) {
+  const { locale } = await params;
+  setRequestLocale(locale);
+
+  return (
+    <div className="pb-12 -space-y-px">
+      <PlaygroundBoundary label="Demo">
```
Comment (amannn) [inline playground/src/app/[locale]/translations/server-components/page.tsx]: Does the "Demo" label really add anything here?

playground/src/app/[locale]/translations/server-components/page.tsx
```diff
@@ -0,0 +1,45 @@
...
+        <header className="mb-6">
+          <h1 className="text-2xl sm:text-[28px] font-semibold tracking-tight text-foreground leading-tight">
+            Server components
+          </h1>
+          <p className="mt-1 text-sm text-muted-foreground">
+            Translations
```
Comment (amannn) [inline playground/src/app/[locale]/translations/server-components/page.tsx]: Also these could be translated I guess.

One aspect to decide is if we should use `useTranslations`  or `useExtracted` for the implementation of the playground. My vote would be for the latter, it seems like many devs are migrating to it. WDYT?

playground/src/app/[locale]/translations/server-components/page.tsx
```diff
@@ -0,0 +1,45 @@
...
+          </p>
+        </header>
+        <Content />
+      </PlaygroundBoundary>
+
+      <PlaygroundBoundary label="Output" variant="dotgrid">
```
Comment (amannn) [inline playground/src/app/[locale]/translations/server-components/page.tsx]: Also here, generally I think we should move all these labels to translations to make a good demo of next-intl.

playground/src/app/[locale]/translations/server-components/server-example.tsx
```diff
@@ -0,0 +1,8 @@
+import { getTranslations } from 'next-intl/server';
+
+export async function ServerExample() {
+  const t = await getTranslations('ServerDemo');
+  return (
+    <p className="text-2xl font-semibold text-foreground">{t('greeting')}</p>
+  );
+}
```
Comment (amannn) [inline playground/src/app/[locale]/translations/server-components/server-example.tsx]: I think this is what we should link to and should match the snippet that we display. Maybe we can use something like a raw-loader to actually read this code as text to display on the page? That would ensure we're showcasing something up-to-date. Maybe we should move the text styles to a shared parent component like `<PlaygroundBoundaryText><ServerExample /></...>` so that we can avoid tailwind markup in the snippet.

playground/src/app/[locale]/translations/client-components/content.mdx
```diff
@@ -0,0 +1,27 @@
+import {TwoColumn} from '@/components/playground/two-column';
+
+<TwoColumn>
+
+- Use `useTranslations()` inside Client Components for any UI that reacts to user input.
+- Messages for client-side namespaces are sent to the browser — only the keys you actually call.
```
Comment (amannn) [inline playground/src/app/[locale]/translations/client-components/content.mdx]: "only the keys you actually call" — this is not entirely true, see the server/client docs.

playground/src/app/[locale]/translations/client-components/content.mdx
```diff
@@ -0,0 +1,27 @@
...
+- ICU placeholders (`{name}`) interpolate at render time and stay locale-aware.
```
Comment (amannn) [inline playground/src/app/[locale]/translations/client-components/content.mdx]: ~~ICU should be a distinct page I think, let's make the snippet as minimal as possible.~~

No, I think you have a good point of demonstrating this with client-side interactivity, making a point about when to use Client Components. We can later link to a separate page on ICU features.

playground/src/app/[locale]/translations/client-components/content.mdx
```diff
@@ -0,0 +1,27 @@
...
+import {useState} from 'react';
+import {useTranslations} from 'next-intl';
+
+export function Greet() {
+  const t = useTranslations('ClientDemo');
+  const [name, setName] = useState('Frodo');
+  return (
+    <>
+      <input value={name} onChange={(e) => setName(e.target.value)} />
+      // !mark
```
Comment (amannn) [inline playground/src/app/[locale]/translations/client-components/content.mdx]: This is rendered in the browser.

playground/src/components/code/annotations/callout.tsx
```diff
@@ -0,0 +1,33 @@
+import type { AnnotationHandler, InlineAnnotation } from 'codehike/code';
+
+export const callout: AnnotationHandler = {
```
amannn: There are quite a few components in the `components` folder. Do we need all of them? It would be good to stay with something as minimal as possible to avoid having to maintain more code than necessary.

playground/src/components/code/annotations/filename.tsx
```diff
@@ -0,0 +1,3 @@
+// Filename is rendered as a header strip in code.tsx using codeblock.meta.
+// This file exists only as a placeholder so importers stay consistent.
+export {};
```
amannn: E.g. this looks like boilerplate that doesn't add anything.

playground/src/components/code/annotations/focus.client.tsx
```diff
@@ -0,0 +1,36 @@
...
+      top = Math.min(top, rect.top - containerRect.top);
+      bottom = Math.max(bottom, rect.bottom - containerRect.top);
+    });
+    if (bottom > containerRect.height || top < 0) {
+      ref.current.scrollTo({
+        top: ref.current.scrollTop + top - 10,
+        behavior: firstRender.current ? 'instant' : 'smooth',
+      });
+    }
+    firstRender.current = false;
```
amannn: Is this code strictly necessary? (please also have a look at other components)

playground/src/components/code/annotations/focus.client.tsx
```diff
@@ -0,0 +1,36 @@
+'use client';
+
+import React, { useLayoutEffect, useRef } from 'react';
+import type { AnnotationHandler } from 'codehike/code';
+import { InnerPre, getPreRef } from 'codehike/code';
+
+export const PreWithFocus: AnnotationHandler['PreWithRef'] = (props) => {
```
amannn: Can we keep the conventions from the existing examples?
- Component names match file names
- PascalCase
- Function declarations instead of arrow functions
- Export components as default

(pls see examples for reference)

playground/src/components/ui/badge.tsx
```diff
@@ -0,0 +1,48 @@
...
+    <Comp
+      data-slot="badge"
+      data-variant={variant}
+      className={cn(badgeVariants({ variant }), className)}
+      {...props}
+    />
+  )
+}
+
+export { Badge, badgeVariants }
```
amannn: `badgeVariants` are not used anywhere outside. Can we only export what is actually shared?

playground/src/lib/utils.ts
```diff
@@ -0,0 +1,6 @@
+import { clsx, type ClassValue } from "clsx"
+import { twMerge } from "tailwind-merge"
+
+export function cn(...inputs: ClassValue[]) {
+  return twMerge(clsx(inputs))
+}
```
amannn: Is this needed for anything? Can we avoid passing class names to shared components so that we don't need merging behavior?

playground/components.json
```diff
@@ -0,0 +1,22 @@
...
+    "css": "src/app/globals.css",
+    "baseColor": "neutral",
+    "cssVariables": true,
+    "prefix": ""
+  },
+  "iconLibrary": "lucide",
+  "aliases": {
+    "components": "@/components",
+    "utils": "@/lib/utils",
+    "ui": "@/components/ui",
```
amannn: I haven't worked with shadcn so far. Can we for simplicity also work with a flat components folder without sub folders?

Would you suggest to keep shadcn here, or is the example minimal enough so that we can get away with a few hand-maintained components?

playground/src/app/globals.css
```diff
@@ -0,0 +1,290 @@
...
+  flex-direction: column;
+  gap: 0.5rem;
+}
+
+.mdx-content li {
+  color: var(--muted-foreground);
+}
+
+.mdx-content li::marker {
+  color: var(--blue-300);
```
amannn: Should be grey, let's use the accent color very sparingly.

playground/src/components/playground/byline.tsx
```diff
@@ -0,0 +1,43 @@
...
+    Icon: Github,
+  },
+  {
+    href: 'https://next-intl.dev/docs/getting-started',
+    label: 'Docs',
+    Icon: BookOpen,
+  },
+  {
+    href: 'https://app-router.vercel.app/',
+    label: 'Inspired by Next.js',
```
amannn: It's really cool to give credit, but I think we could do so in an announcement tweet for example and don't need to hardcode this.

playground/src/app/[locale]/page.tsx
```diff
@@ -0,0 +1,73 @@
...
+                <div className="flex items-center gap-2 text-[10px] font-mono font-semibold uppercase tracking-[0.18em] text-muted-foreground">
+                  <Icon className="h-3 w-3" strokeWidth={2} />
+                  {section.title}
+                </div>
+                <div className="grid grid-cols-1 sm:grid-cols-2 gap-px bg-border">
+                  {section.items.map((item) => (
+                    <Link
+                      href={item.slug}
+                      key={item.title}
+                      className="group flex flex-col gap-1.5 bg-background px-5 py-4 transition-colors hover:bg-card/50"
```
Comment (amannn) [inline playground/src/app/[locale]/page.tsx]: I think the background color doesn't have an effect here, but it causes a weird flash when switching the theme.

playground/src/components/ui/dropdown-menu.tsx
```diff
@@ -0,0 +1,257 @@
...
+  inset?: boolean
+  variant?: "default" | "destructive"
+}) {
+  return (
+    <DropdownMenuPrimitive.Item
+      data-slot="dropdown-menu-item"
+      data-inset={inset}
+      data-variant={variant}
+      className={cn(
+        "relative flex cursor-default items-center gap-2 rounded-sm px-2 py-1.5 text-sm outline-hidden select-none focus:bg-accent focus:text-accent-foreground data-[disabled]:pointer-events-none data-[disabled]:opacity-50 data-[inset]:pl-8 data-[variant=destructive]:text-destructive data-[variant=destructive]:focus:bg-destructive/10 data-[variant=destructive]:focus:text-destructive dark:data-[variant=destructive]:focus:bg-destructive/20 [&_svg]:pointer-events-none [&_svg]:shrink-0 [&_svg:not([class*='size-'])]:size-4 [&_svg:not([class*='text-'])]:text-muted-foreground data-[variant=destructive]:*:[svg]:text-destructive!",
```
amannn: I think we shouldn't use the accent color in the dropdown, gray is fine for here.

playground/src/components/playground/byline.tsx
```diff
@@ -0,0 +1,43 @@
...
+    Icon: ArrowUpRight,
+  },
+];
+
+export function PlaygroundByline() {
+  return (
+    <div className="mt-12 flex flex-col gap-4 border-t border-border pt-6 sm:flex-row sm:items-center sm:justify-between">
+      <div className="text-[11px] font-mono uppercase tracking-[0.18em] text-muted-foreground/60">
```
amannn: Shouldn't be the uppercase variant here I think, let's use the same style as the links on the right side.

amannn (commented):
Overall looks really cool, thanks @Adebesin-Cell! 🙌 

Here's a first review pass, hopefully there's something useful in this for you. Some remarks are questions to you, let me know what you think!

playground/src/app/[locale]/translations/server-components/page.tsx
Comment (Adebesin-Cell) [reply playground/src/app/[locale]/translations/server-components/page.tsx]: Yeah, that makes sense, actually. I think I did a general translation after I was done, and stripped it off after the split

playground/src/app/[locale]/translations/server-components/content.mdx
Comment (Adebesin-Cell) [reply playground/src/app/[locale]/translations/server-components/content.mdx]: Hmm, interesting. My idea was we would have a markdown per locale, say we'd have `server-components/en/content.mdx` and `server-components/de/content.mdx` if we would support more locales, we'd have more mdx files.

playground/src/app/globals.css
Adebesin-Cell: Cleaned this up — dropped the chart vars and the wholesale palette re-mapping. The palette now lives directly in `@theme` as `--color-gray-*` / `--color-blue-*`.

I'd suggest keeping the small semantic layer though (`background`, `foreground`, `muted`, etc.). It's the bit that flips under `.dark` for runtime theming via `next-themes`, and having one source of truth for the light/dark values makes theming much easier to maintain than scattering `dark:` variants across every component.

playground/src/app/[locale]/layout.tsx
Comment (Adebesin-Cell) [reply playground/src/app/[locale]/layout.tsx]: for sure

playground/src/components/playground/byline.tsx
Adebesin-Cell: Ha, make sense 👍

playground/src/components/code/annotations/filename.tsx
Adebesin-Cell: Good catch 👏

playground/src/app/layout.tsx
Adebesin-Cell: Done — moved everything into `examples/example-app-router-patterns/` and kept your `/design` page (middleware skips it).

playground/public/file.svg
Adebesin-Cell: Removed the unused Next template SVGs; icons are all lucide now.

playground/src/assets/logo.tsx
Adebesin-Cell: Matched the docs logo — `currentColor` + a padded viewBox so it's no longer clipped. (The little dot at the bottom-right is part of the official mark.)

playground/tests/playground.spec.ts
Adebesin-Cell: Removed the Playwright tests + config — good to lean on `/e2e` going forward.

playground/src/app/[locale]/translations/server-components/page.tsx
Comment (Adebesin-Cell) [reply playground/src/app/[locale]/translations/server-components/page.tsx]: Good catch — the GitHub links now encode each segment, so `[locale]` becomes `%5Blocale%5D` and resolves correctly.

playground/src/app/globals.css
Adebesin-Cell: Removed — the chart vars weren't used anywhere.

playground/src/app/globals.css
Adebesin-Cell: Left it as a one-rule utility class for now since it's tiny; happy to move it into the boundary component if you'd prefer no global.

playground/src/app/[locale]/layout.tsx
Comment (Adebesin-Cell) [reply playground/src/app/[locale]/layout.tsx]: Switched to Inter.

playground/src/app/[locale]/layout.tsx
Comment (Adebesin-Cell) [reply playground/src/app/[locale]/layout.tsx]: Yep — will adapt this once `next/root-params` ships.

playground/src/app/[locale]/layout.tsx
Comment (Adebesin-Cell) [reply playground/src/app/[locale]/layout.tsx]: It's required by `next-themes` — it sets the theme class on `<html>` before hydration, so without it React warns about the mismatch.

playground/src/app/[locale]/layout.tsx
Comment (Adebesin-Cell) [reply playground/src/app/[locale]/layout.tsx]: Inlined the `ThemeProvider` and dropped the `ClientProviders` wrapper.

playground/src/app/[locale]/page.tsx
Comment (Adebesin-Cell) [reply playground/src/app/[locale]/page.tsx]: Done — `lib/nav` now holds message keys and the labels render via `t()`, so the nav is translatable.

playground/src/app/[locale]/translations/server-components/content.mdx
Comment (Adebesin-Cell) [reply playground/src/app/[locale]/translations/server-components/content.mdx]: Switched to a single column so the snippet no longer gets cut off.

playground/src/app/[locale]/translations/server-components/content.mdx
Comment (Adebesin-Cell) [reply playground/src/app/[locale]/translations/server-components/content.mdx]: Done — snippet is minimal now (`useTranslations`, no `setRequestLocale`/`params`).

playground/src/app/[locale]/translations/server-components/content.mdx
Comment (Adebesin-Cell) [reply playground/src/app/[locale]/translations/server-components/content.mdx]: Added — the prose and a real non-async Server Component example both make the point that `useTranslations` works without `async`.

playground/src/app/[locale]/translations/server-components/page.tsx
Comment (Adebesin-Cell) [reply playground/src/app/[locale]/translations/server-components/page.tsx]: Agreed — dropped the "Demo" label.

playground/src/app/[locale]/translations/server-components/page.tsx
Comment (Adebesin-Cell) [reply playground/src/app/[locale]/translations/server-components/page.tsx]: Went with `useTranslations` here since these pages are specifically about reading messages in Server/Client Components and it's the most familiar entry point. I'd love to showcase `useExtracted` too — planning a dedicated extraction page as a follow-up so the playground teaches both. (Kept it off the foundation to avoid the Next 16 bump the extraction setup seems to want.)

playground/src/app/[locale]/translations/server-components/content.mdx
Comment (Adebesin-Cell) [reply playground/src/app/[locale]/translations/server-components/content.mdx]: kept mdx for now, let me know if you prefer `t()`

playground/src/app/[locale]/translations/server-components/page.tsx
Comment (Adebesin-Cell) [reply playground/src/app/[locale]/translations/server-components/page.tsx]: Done — these labels now come from messages too.

playground/src/app/[locale]/translations/server-components/server-example.tsx
Comment (Adebesin-Cell) [reply playground/src/app/[locale]/translations/server-components/server-example.tsx]: Made the live example match the snippet (both `useTranslations`, non-async). The raw-loader idea (reading the file as text so they can't drift) is nice — I'd do that as a follow-up rather than block here.

playground/src/app/[locale]/translations/client-components/content.mdx
Comment (Adebesin-Cell) [reply playground/src/app/[locale]/translations/client-components/content.mdx]: Reworded — dropped the "only the keys you actually call" claim.

playground/src/app/[locale]/translations/client-components/content.mdx
Comment (Adebesin-Cell) [reply playground/src/app/[locale]/translations/client-components/content.mdx]: Right — the prose now says the ICU args resolve in the browser as state changes.

playground/src/components/code/annotations/callout.tsx
Adebesin-Cell: Trimmed a lot here — dropped `demo-card`, `two-column`, and all the Code Hike annotation handlers except the one we actually use (`mark`).

playground/src/components/code/annotations/filename.tsx
Adebesin-Cell: Removed.

playground/src/components/code/annotations/focus.client.tsx
Adebesin-Cell: Removed — wasn't needed after trimming the annotations.

playground/src/components/code/annotations/focus.client.tsx
Adebesin-Cell: The remaining code (`code.tsx`, `mark.tsx`) follows the conventions now — function declarations, names matching files.

playground/src/components/ui/badge.tsx
Adebesin-Cell: Gone — dropped shadcn entirely, so no more `badge`/`badgeVariants`.

playground/src/lib/utils.ts
Adebesin-Cell: Removed `cn` — no class merging needed now that shared components aren't passed arbitrary class names.

playground/components.json
Adebesin-Cell: Dropped shadcn — it's just a few hand-maintained components now in a flat folder, no `components.json`.

playground/src/app/globals.css
Adebesin-Cell: Done — marker uses grey now, accent kept sparing.

playground/src/components/playground/byline.tsx
Adebesin-Cell: Removed the hardcoded credit — agreed, better as an announcement tweet.

playground/src/app/[locale]/page.tsx
Comment (Adebesin-Cell) [reply playground/src/app/[locale]/page.tsx]: Fixed at the source via `disableTransitionOnChange` on the ThemeProvider, so no flash on theme switch.

playground/src/components/ui/dropdown-menu.tsx
Adebesin-Cell: The dropdown is gone entirely — the locale switcher is a plain segmented control now, no accent.

playground/src/components/playground/byline.tsx
Adebesin-Cell: Cleaned up — dropped the uppercase label; the byline is just the links now in the same style.

examples/example-app-router-patterns/src/assets/logo.tsx
```diff
@@ -0,0 +1,32 @@
...
+          d="M8.847 28.785c16.248-1.974 32.318-2.949 48.211-2.923 15.893.025 31.156 1.05 45.789 3.073"
+          strokeLinecap="square"
+        />
+        <ellipse cx="57.421" cy="57.12" rx="28.035" ry="57.12" />
+        <path
+          d="M11.695 88.603c14.934 2 30.008 3 45.224 3 15.215 0 30.525-1 45.928-3M1.009 59.069h112.482"
+          strokeLinecap="square"
+        />
+      </g>
+    </svg>
```
amannn: Seems like there's still something off with this dot at the bottom right:

Image

Docs:

Image

examples/example-app-router-patterns/src/components/playground/github-link.tsx
```diff
@@ -0,0 +1,22 @@
+import {ArrowUpRight, Github} from 'lucide-react';
+
+export function GitHubLink({path}: {path: string}) {
+  // Encode segments so dynamic ones like `[locale]` work as a GitHub URL
+  const href = `https://github.com/amannn/next-intl/tree/main/${path
```
amannn: Can we move the base url to a config file at `src/config.tsx`? This is also used for `byline.tsx`.

examples/example-app-router-patterns/src/app/globals.css
```diff
@@ -0,0 +1,243 @@
...
+    @apply bg-background text-foreground antialiased;
+  }
+  button:not([disabled]),
+  [role='button']:not([disabled]) {
+    cursor: pointer;
+  }
+}
+
+/* Subtle dot-grid backdrop for the "output" boundary box */
+.dotgrid {
```
amannn: Coming from https://github.com/amannn/next-intl/pull/2084#discussion_r3324240880

Yep, I'd prefer keeping as much styles out of `global.css` as possible and move to component code. Otherwise this could end up as dead code during a refactor.

examples/example-app-router-patterns/src/app/[locale]/translations/client-components/page.tsx
```diff
@@ -0,0 +1,51 @@
...
+import {ClientExample} from './client-example';
+import Content from './content.mdx';
+
+type Props = {
+  params: Promise<{locale: string}>;
+};
+
+export default async function ClientComponentsPage({params}: Props) {
+  const {locale} = await params;
+  setRequestLocale(locale);
```
Comment (amannn) [inline examples/example-app-router-patterns/src/app/[locale]/translations/client-components/page.tsx]: Coming from https://github.com/amannn/next-intl/pull/2084#discussion_r3324246644: Ok to use `requestLocale` for now, but I'd already avoid `setRequestLocale` and just use dynamic rendering for now. Next.js 16.3 with `next/root-params` was expected for last week already, I think this should be out any minute now.

examples/example-app-router-patterns/src/app/layout.tsx
```diff
@@ -0,0 +1,40 @@
...
+  title: 'next-intl playground',
+  description:
+    'Explore translations, formatting, routing and patterns with next-intl.'
+};
+
+export default async function RootLayout({children}: {children: ReactNode}) {
+  const locale = await getLocale();
+
+  return (
+    <html lang={locale} suppressHydrationWarning className={inter.variable}>
```
amannn: This should be part of `[locale]/layout.tsx`. There's no need for a layout at `app/`, right?

examples/example-app-router-patterns/src/app/[locale]/page.tsx
```diff
@@ -0,0 +1,71 @@
...
+import {ArrowRight} from 'lucide-react';
+import {useTranslations} from 'next-intl';
+import {setRequestLocale} from 'next-intl/server';
+
+type Props = {
+  params: Promise<{locale: string}>;
+};
+
+export default async function HomePage({params}: Props) {
+  const {locale} = await params;
```
Comment (amannn) [inline examples/example-app-router-patterns/src/app/[locale]/page.tsx]: You can use `const {locale} = use(params)` to avoid the async/non-async split of components

examples/example-app-router-patterns/src/app/[locale]/page.tsx
```diff
@@ -0,0 +1,71 @@
...
+  setRequestLocale(locale);
+  return <Home />;
+}
+
+function Home() {
+  const t = useTranslations('Layout');
+  const tNav = useTranslations('Nav');
```
Comment (amannn) [inline examples/example-app-router-patterns/src/app/[locale]/page.tsx]: Coming from https://github.com/amannn/next-intl/pull/2084#discussion_r3324289640:

I think it's different aspects:
1. Which patterns to we demonstrate on the pages
2. How we build this app

We can keep `useTranslations` for the demonstrated patterns, but for the actual app code I'd switch to `useExtracted`. It naturally avoids issues like this `tNav` in this file if you just inline the labels where they are used (you'll need to remove `nav.ts` then as the labels will be inline in a component. the types from the file can go into `src/types.tsx` I'd say).

examples/example-app-router-patterns/src/app/[locale]/translations/client-components/content.mdx
```diff
@@ -0,0 +1,20 @@
+- `useTranslations` works the same in Client Components, so interactive UI can read messages too.
+- ICU arguments like `{name}` are resolved in the browser as state changes — try typing below.
```
Comment (amannn) [inline examples/example-app-router-patterns/src/app/[locale]/translations/client-components/content.mdx]: Coming from https://github.com/amannn/next-intl/pull/2084#discussion_r3324274771: Yeah, I'd switch to inline labels here. Otherwise the locale switcher will appear to be non-working if we have markdown that is not translated. Using inline labels with `t` should be easier to maintain (esp. with `useExtracted`).

We can also avoid the `@next/mdx` dependency then I think if we use `codehike` directly on a page and pass code samples like below directly as multiline strings. This will also enable Turbopack, since the way `next/mdx` is used currently causes a switch to webpack.

examples/example-app-router-patterns/src/app/[locale]/translations/server-components/content.mdx
```diff
@@ -0,0 +1,12 @@
+- Call `useTranslations` from a Server Component to read messages — it works in both `async` and non-async components.
```
Comment (amannn) [inline examples/example-app-router-patterns/src/app/[locale]/translations/server-components/content.mdx]: Coming from https://github.com/amannn/next-intl/pull/2084#discussion_r3324282307:

> it works in both `async` and non-async components

This is not true, please see the docs. This page should cover that `await getTranslations` can be called in async components, and `useTranslations` in non-async ones.

Should we cover both patterns in a single code sample/demo or should we split this with sub headlines into two sections?

examples/example-app-router-patterns/src/app/[locale]/translations/client-components/content.mdx
Comment (amannn) [inline examples/example-app-router-patterns/src/app/[locale]/translations/client-components/content.mdx]: I think the "try typing below." is not necessary.

There's this phrase I like which I think you already did very well with the input field: Show, don't tell.

examples/example-app-router-patterns/src/app/[locale]/translations/client-components/content.mdx
```diff
@@ -0,0 +1,20 @@
+- `useTranslations` works the same in Client Components, so interactive UI can read messages too.
```
Comment (amannn) [inline examples/example-app-router-patterns/src/app/[locale]/translations/client-components/content.mdx]:
```suggestion
- `useTranslations` can be called in Client Components to incorporate client-side state.
```

examples/example-app-router-patterns/src/app/[locale]/translations/client-components/content.mdx
```diff
@@ -0,0 +1,20 @@
...
+import {useState} from 'react';
+import {useTranslations} from 'next-intl';
+
+export function Greet() {
+  const t = useTranslations('ClientComponentsPage');
+  const [name, setName] = useState('Frodo');
+  return (
+    <>
+      <input value={name} onChange={(e) => setName(e.target.value)} />
+      // !mark
```
Comment (amannn) [inline examples/example-app-router-patterns/src/app/[locale]/translations/client-components/content.mdx]: This is rendered on the page:

Image

I guess the issue is that `//` comments don't work in JSX (only /* … */)

examples/example-app-router-patterns/src/components/code/annotations/mark.tsx
```diff
@@ -0,0 +1,35 @@
+import type {AnnotationHandler} from 'codehike/code';
+import {InnerLine} from 'codehike/code';
+
+export const mark: AnnotationHandler = {
```
amannn: Instead of the `annotations` folder, let's just move these inline into the `Code` component module.

We can still go the direction of adding a folder once this becomes more extensive, but inline/local wins for me for the current state of this app to avoid structural bloat.

examples/example-app-router-patterns/src/components/playground/boundary.tsx
```diff
@@ -0,0 +1,43 @@
+import clsx from 'clsx';
+import type React from 'react';
+
+export const PlaygroundBoundary = ({
```
amannn: Also no need for the `playground` folder I think.

Let's do:

```
components/
  Code.tsx
  PlaygroundBoundary.tsx
  ...
```

Just a simple flat list of components for now.

Also:
- PascalCase names (I know more people seem to be used kebab-case these days, but all of the other code in this repo uses the PascalCase file naming already)
- Make sure component name matches file name

examples/example-app-router-patterns/src/app/globals.css
```diff
@@ -0,0 +1,243 @@
...
+    circle at center,
+    color-mix(in oklch, var(--muted-foreground) 28%, transparent) 1px,
+    transparent 1px
+  );
+  background-size: 14px 14px;
+  background-position: 0 0;
+}
+
+/* MDX content typography (scoped — no @tailwindcss/typography needed) */
+.mdx-content {
```
amannn: Also these global `.mdx-content` styles could be moved to a `<Prose>{children}</Prose>` component that applies these styles for its children.

examples/example-app-router-patterns/src/components/playground/link-status.tsx
```diff
@@ -0,0 +1,15 @@
+'use client';
+
+import {useLinkStatus} from 'next/link';
+
+export function LinkStatus() {
```
amannn: If our whole app renders statically, there's no need for this—right?

See the [`useLinkStatus` docs](https://nextjs.org/docs/app/api-reference/functions/use-link-status): `loading.js` is preferred, but also that is only needed for dynamic pages.

WDYT?

examples/example-app-router-patterns/src/app/[locale]/translations/client-components/content.mdx
```diff
@@ -0,0 +1,20 @@
+- `useTranslations` works the same in Client Components, so interactive UI can read messages too.
+- ICU arguments like `{name}` are resolved in the browser as state changes — try typing below.
+
+```tsx app/greet.tsx
```
Comment (amannn) [inline examples/example-app-router-patterns/src/app/[locale]/translations/client-components/content.mdx]: I think `app/greet.tsx` doesn't add anything here. We could however use a generic "Code" label, related to "Output" below.

examples/example-app-router-patterns/src/components/playground/locale-switcher.tsx
```diff
@@ -0,0 +1,39 @@
+'use client';
+
+import {Link, usePathname} from '@/i18n/navigation';
+import {routing} from '@/i18n/routing';
+import {clsx} from 'clsx';
+import {useLocale} from 'next-intl';
+import {useParams} from 'next/navigation';
+
+export function LocaleSwitcher() {
```
amannn: This seems to be broken:

https://github.com/user-attachments/assets/164962d7-52ef-4048-ad6d-df421af11f9b

examples/example-app-router-patterns/src/components/playground/locale-switcher.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const locale = useLocale();
+  const pathname = usePathname();
+  const params = useParams();
+
+  return (
+    <div
+      role="group"
+      aria-label="Switch language"
```
amannn: There are still hardcoded inline labels around, we should switch all consistently to `t(…)`

examples/example-app-router-patterns/src/app/[locale]/page.tsx
```diff
@@ -0,0 +1,71 @@
...
+    <div className="pb-12">
+      <div className="mb-12 pt-8 text-center sm:mb-16 sm:pt-12">
+        <h1 className="text-foreground text-[34px] font-semibold tracking-tight sm:text-5xl">
+          {t('title')}
+        </h1>
+        <p className="text-muted-foreground mx-auto mt-4 max-w-xl text-base sm:text-lg">
+          {t('tagline')}
+        </p>
+      </div>
+      <PlaygroundBoundary label={t('examples')} className="space-y-10">
```
Comment (amannn) [inline examples/example-app-router-patterns/src/app/[locale]/page.tsx]: Is it `t('examples')`? Maybe a generic "pages" could fit better terminology-wise? (also the rendered labels)

examples/example-app-router-patterns/src/app/[locale]/page.tsx
```diff
@@ -0,0 +1,71 @@
...
+              <div className="text-muted-foreground flex items-center gap-2 font-mono text-[10px] font-semibold tracking-[0.18em] uppercase">
+                <Icon className="h-3 w-3" strokeWidth={2} />
+                {tNav(section.titleKey)}
+              </div>
+              <div className="bg-border grid grid-cols-1 gap-px sm:grid-cols-2">
+                {section.items.map((item) => (
+                  <Link
+                    href={item.slug}
+                    key={item.slug}
+                    className="group bg-background hover:bg-muted flex flex-col gap-1.5 px-5 py-4 transition-colors"
```
Comment (amannn) [inline examples/example-app-router-patterns/src/app/[locale]/page.tsx]: `hover:bg-muted`: I think that's too dark, can we use a lighter shade?

examples/example-app-router-patterns/src/app/globals.css
```diff
@@ -0,0 +1,243 @@
...
+  --color-input: var(--input);
+  --color-ring: var(--ring);
+  --color-sidebar: var(--sidebar);
+  --color-sidebar-foreground: var(--sidebar-foreground);
+  --color-sidebar-border: var(--sidebar-border);
+}
+
+:root {
+  --background: var(--color-gray-50);
+  --foreground: var(--color-gray-900);
```
amannn: There are two approaches to dark mode:
1. Using semantic tokens that switch their internal value (used here)
2. Use `dark:` class names to declare which specific values a given element has

I know the first one is tempting from an engineering perspective, but it creates a quite heavy abstraction IMO that doesn't allow for much fine-tuning on a case-by-case basis and also a lot of upfront declarations (e.g. specific ones like `--sidebar-foreground` below).

If you compare what e.g. the Next.js playground does:

```
bg-gray-900 hover:bg-gray-800
```

It takes a bit of discipline to apply colors consistently, but I think it's just dead-simple. Reuse of colors is typically achieved through component reuse then (e.g. by using `<Prose>` where relevant).

I think there was a tweet from Adam from the Tailwind team at some point that he also prefers this approach for the reasons listed here.

Can we switch to this simpler inline approach to remove another architectural layer and reduce complexity?

amannn (commented):
Really nice work @Adebesin-Cell! 👏 Thanks also for the helpful replies to my comments!

I've added a few more inline comments, let me know what you think!

PR #2086, opened 2025-11-10
docs: Generate blog RSS feed on build

bawpcwpn:
Hi @amannn !

Hope this PR is ok to propose! It adds basic RSS 2.0 XML feed support for the blog.

It's a build time scripts that parses a the `<BlogPostLink />` components from the blog `index.mdx` file, and generates a static XML file in `/public/feed.xml` 

No new dependencies are added, and linting and testing pass.

---

> [!NOTE]
> Adds build-time RSS feed generation for the blog and exposes it via an alternate link in the docs site head.
> 
> - **Build/Infra**:
>   - Add `scripts/generate-rss.mjs` to parse `BlogPostLink` entries from `src/pages/blog/index.mdx` and generate `public/feed.xml` (RSS 2.0).
>   - Update `package.json` `build` script to run `next-sitemap` then `node scripts/generate-rss.mjs`.
> - **Site Head**:
>   - Add RSS `<link rel="alternate" type="application/rss+xml" href="/feed.xml" title="next-intl Blog RSS Feed" />` in `src/theme.config.tsx`.
> - **Misc**:
>   - Ignore generated `public/feed.xml` in `docs/.gitignore`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f6b574e352fcf0ce5aaca80b1b93cb13e6bafd62. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/docs/scripts/generate-rss.mjs
+++ b/docs/scripts/generate-rss.mjs
@@ -1,17 +1,23 @@
 import {readFileSync, writeFileSync} from 'node:fs';
 import { dirname, join } from 'node:path';
 import { fileURLToPath } from 'node:url';
+import config from '../src/config.js'
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
 // Site configuration
-const SITE_URL = 'https://next-intl.dev';
+const SITE_URL = config.baseUrl;
 const FEED_URL = `${SITE_URL}/feed.xml`;
 const TITLE = 'next-intl Blog';
-const DESCRIPTION = 'Updates, guides, and insights about internationalization (i18n) in Next.js with next-intl';
+const DESCRIPTION = 'Updates, guides, and insights about internationalization in Next.js';
 const LANGUAGE = 'en';
-const IMAGE_URL = `${SITE_URL}/og-image.png`;
+const IMAGE_URL = `${SITE_URL}/api/og-image?params=${encodeURIComponent(
+  JSON.stringify({
+    title: 'Blog',
+    subtitle: DESCRIPTION
+  })
+)}`
 
 function parseBlogPostLinks(mdx) {
   const results = [];
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -8,3 +8,4 @@ tsconfig.tsbuildinfo
 public/robots.txt
 public/sitemap.xml
 public/sitemap-0.xml
+public/feed.xml


docs/scripts/generate-rss.mjs
```diff
@@ -0,0 +1,135 @@
+import {readFileSync, writeFileSync} from 'node:fs';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+
+// Site configuration
+const SITE_URL = 'https://next-intl.dev';
```
amannn: This can be read from `./src/config.js`.

docs/scripts/generate-rss.mjs
```diff
@@ -0,0 +1,135 @@
...
+const FEED_URL = `${SITE_URL}/feed.xml`;
+const TITLE = 'next-intl Blog';
+const DESCRIPTION = 'Updates, guides, and insights about internationalization (i18n) in Next.js with next-intl';
+const LANGUAGE = 'en';
+const IMAGE_URL = `${SITE_URL}/og-image.png`;
```
amannn: Doesn't seem to exist: https://next-intl.dev/og-image.png

We can use:


```suggestion
const IMAGE_URL = `${SITE_URL}/api/og-image?params=${encodeURIComponent(
  JSON.stringify({
    title: 'Blog',
    subtitle: DESCRIPTION
  })
)}`;
```

docs/scripts/generate-rss.mjs
```diff
@@ -0,0 +1,135 @@
...
+import { fileURLToPath } from 'node:url';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+
+// Site configuration
+const SITE_URL = 'https://next-intl.dev';
+const FEED_URL = `${SITE_URL}/feed.xml`;
+const TITLE = 'next-intl Blog';
+const DESCRIPTION = 'Updates, guides, and insights about internationalization (i18n) in Next.js with next-intl';
```
amannn: If we use:

```suggestion
const DESCRIPTION = 'Updates, guides, and insights about internationalization in Next.js';
```

… then we could encode this into the OG image and it fits quite nice:

Image

docs/scripts/generate-rss.mjs
```diff
@@ -0,0 +1,135 @@
...
+  // RSS 2.0 uses RFC-822/RFC-1123 format; toUTCString() is acceptable.
+  return new Date(d).toUTCString();
+}
+
+/**
+ * blogPosts: Array<{ title, date, author?, description?, href? , slug? }>
+ *  - date: Date | string
+ *  - link is resolved from href (preferred) or `/blog/${slug}`
+ */
+export function generateRSS(blogPosts) {
```
amannn, line 97: If we move this to something like [`src/app/rss.xml/route.tsx`](https://nextjs.org/docs/app/guides/backend-for-frontend#content-types) we can't run it at build time and therefore it might cause trouble on serverless functions, right? Just thought that might be slightly cleaner, but we can keep it like this if it's easier.

docs/src/theme.config.tsx
```diff
@@ -142,6 +142,13 @@ export default {
         <meta content="#da532c" name="msapplication-TileColor" />
         <meta content="#ffffff" name="theme-color" />
 
+        <link
+          href="/feed.xml"
```
amannn, line 146: Can we move this to `.gitignore`?

docs/scripts/generate-rss.mjs
bawpcwpn: With NextJS 14 it'd have to be an API route that would need a rewrite, and even then I'm not sure if it can reliably read from the index.mdx file unless it was explicitly included at build. 
Though also because the blog is static and not coming from a CMS, it's perhaps more efficient to have create the feed at build time, though I definitely agree, a route would be much cleaner! It's initially what I planned to do when I noticed this morning I couldn't add the blog to my RSS reader.

amannn: Yeah, let's keep it like this—thanks for confirming 👍

amannn (approved):
Awesome, thanks a bunch!

PR #2090, opened 2025-11-11
fix: Avoid partial matches in overlapping localized pathnames

### LINKED ISSUE #2089: Segments are wrongly parsed by middleware adding part of the canonical route as a segment of the localized route
### Description

I have a route defined as `/stock[[...slug]]/page.tsx`, in `routing.tsx` I've added the pathname as:

```ts
    '/stock/[[...slug]]': {
      en: '/stock/[[...slug]]',
      ro: '/stoc/[[...slug]]',
    }
````

Notice how the Romanian translation is a subset of the english version "stock" -> "stoc" this is important.

When I visit /en/stock/truck and switch to romanian, the next path gets written as `/ro/stoc/k` it happens more frequently when there's a slug but also happens if you visit /en/stock and switch to romanian.

this is resulting in an infinite loop of redirects.

So far the only think that seems to prevent this is by defining the localized romanian route as `  ro: '/stock/[[...slug]]'` in routing.ts so that its a 1-1 match with the canonical, I've also tested with different values and it works, its the partial match (stoc) that seems to create issues.

### Verifications

https://codesandbox.io/p/sandbox/strange-wave-fxis37

### Reproduction description

1) go to `/en/stock/truck`
2) use the language switcher in the navbar to switch to romanian
3) check the tracing logs in the console, notice how it logs the route as /ro/stoc/k 

In case of a route that doesn't exist under /stock my middleware logic redirects back to /stock but since the middleware rewrites /ro/stoc as /ro/stoc/k I end up with an infinite loop of redirects because /ro/stoc/k is not found and redirects to /ro/stock which then gets rewritten as /ro/stoc/k again and ends up redirected again.

I've digged into the middleware code and the culprit seems to be the regex match in `getRouteParams` that matches `stock` in `/ro/stock` returning 'k' as the slug and the path becomes `/ro/stoc/k

### Expected behaviour

the route should be translated correctly

### PR DESCRIPTION
**Pull request by amannn**
Fixes #2089

Fixes a middleware loop by preventing partial matches for optional catch-all segments and ensuring missing optional parameters default to an empty string.

The previous regex for optional catch-all segments like `[[...slug]]` was too permissive, allowing routes such as `/stoc/[[...slug]]` to partially match static paths like `/stock`. This caused `next-intl`'s middleware to incorrectly identify a route, leading to an infinite redirect loop as reported in #2089. The updated regex now requires a clear boundary for optional segments, and missing optional parameters are explicitly set to an empty string for consistent handling.

---
<a href="https://cursor.com/background-agent?bcId=bc-77fee808-9ea4-46f1-b93c-1136508b87d2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-cursor-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-cursor-light.svg">Open in Cursor</picture></a>&nbsp;<a href="https://cursor.com/agents?id=bc-77fee808-9ea4-46f1-b93c-1136508b87d2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-web-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-web-light.svg">Open in Web</picture></a>

### DIFF
--- a/packages/next-intl/src/shared/utils.tsx
+++ b/packages/next-intl/src/shared/utils.tsx
@@ -127,8 +127,8 @@ export function getLocaleAsPrefix(locale: string) {
 export function templateToRegex(template: string): RegExp {
   const regexPattern = template
     // Replace optional catchall ('[[...slug]]')
-    .replace(/\/\[\[(\.\.\.[^\]]+)\]\]/g, '(?:/(.*))?')
-    .replace(/\[\[(\.\.\.[^\]]+)\]\]/g, '(?:/(.*))?')
+    .replace(/\/\[\[(\.\.\.[^\]]+)\]\]/g, '(?:/(.*))?') // With leading slash
+    .replace(/\[\[(\.\.\.[^\]]+)\]\]/g, '(?:/(.*))?') // Without leading slash
     // Replace catchall ('[...slug]')
     .replace(/\[(\.\.\.[^\]]+)\]/g, '(.+)')
     // Replace regular parameter ('[slug]')

### CODE REVIEW

File: packages/next-intl/.size-limit.ts [hunk 1]
```diff
@@ -42,7 +42,7 @@ const config: SizeLimitConfig = [
   {
     name: "import * from 'next-intl/middleware'",
     path: 'dist/esm/production/middleware.js',
-    limit: '9.7 KB'
+    limit: '9.71 KB'
```
amannn, line 45: @cursoragent Still an error:

```
  import {createNavigation} from 'next-intl/navigation' (react-client)
  Package size limit has exceeded by 1 B
  Size limit: 2340 B
  Size:       2341 B with all dependencies, minified and brotlied
```

Bump this as well.

packages/next-intl/src/middleware/utils.test.tsx
```diff
@@ -73,6 +73,13 @@ describe('getRouteParams', () => {
       '...categories': 'clothing/t-shirts'
     });
   });
+
+  it('does not partially match similar optional catch-all segments', () => {
+    expect(getRouteParams('/stoc/[[...slug]]', '/stock')).toBeUndefined();
+    expect(
+      getRouteParams('/stoc/[[...slug]]', '/stock/truck')
+    ).toBeUndefined();
+  });
```
amannn, line 80: @cursoragent 

```
> prettier src --check

Checking formatting...
[warn] src/middleware/utils.test.tsx
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
```

PR #2091, opened 2025-11-11
fix(useExtracted): Ensure deterministic translation key order

amannn:
Sorts message references alphabetically to ensure deterministic PO file output.

Previously, when multiple files referenced the same message, the order of these references in the PO file was nondeterministic. This caused PO files to change unnecessarily between builds, even when no actual translations or code changed, leading to "noisy" diffs for users. This fix ensures that shared message references are always sorted alphabetically, providing stable and predictable PO file generation.

---
<a href="https://cursor.com/background-agent?bcId=bc-92413a35-439b-4252-baea-8217fe88d4dc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-cursor-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-cursor-light.svg">Open in Cursor</picture></a>&nbsp;<a href="https://cursor.com/agents?id=bc-92413a35-439b-4252-baea-8217fe88d4dc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-web-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-web-light.svg">Open in Web</picture></a>


packages/next-intl/src/extractor/catalog/CatalogManager.test.tsx
amannn, line 1: @cursoragent Add this test for ExtractionCompiler.tsx instead for more e2e relevance. I don't want to test CatalogManager in isolation. Also make sure to fix eslint errors before commiting. Make sure to follow existing code style in ExtractionCompiler test suite.

Also, the test should verify we don't have race conditions:
- file a uses msg a
- file b uses msg b
- file c uses msg c
- file d uses msg b

Compiled in this order:
- file a
- file d
- file c
- file b

`next-intl/src/extractor/formatters/utils.tsx`: getSortedMessages should produce this order in the written file:
- file a: msg a
- file b & d: msg b
- file c: msg c

packages/next-intl/src/extractor/catalog/CatalogManager.tsx
```diff
@@ -202,11 +202,18 @@ export default class CatalogManager {
         // reference, which is the current file. We need to merge this with
         // potentially existing references.
         const references = [...(prevMessage.references ?? [])];
-        message.references.forEach((ref) => {
-          if (!references.some((cur) => cur.path === ref.path)) {
-            references.push(ref);
+        message.references.forEach((reference) => {
```
amannn: @cursoragent Seems like you only renamed the variable here, revert this.

packages/next-intl/src/extractor/catalog/CatalogManager.test.tsx
amannn: @cursoragent Any update?

PR #2093, opened 2025-11-11
fix(useExtracted): Filter common directories from `srcPath`, if not explicitly requested

amannn:
Ignore `node_modules`, `.next` and `.git` by default, allowing consumers to use `srcPath: './'` in case they're not using a `src` directory. In a future iteration, we might infer these directories from `.gitignore`.

Note that you can still use:

```
srcPath: ['./src', './node_modules/@acme/components'],
```

… in case you want to extract from source files inside of such a directory.

This PR also contains a fix that allows using trailing slashes in `srcPath` (e.g. `./src/app/`).

### DIFF
--- a/docs/src/pages/docs/usage/plugin.mdx
+++ b/docs/src/pages/docs/usage/plugin.mdx
@@ -128,7 +128,7 @@ If your project is split into multiple folders, you can provide an array of path
 
 ```tsx
 // Not using a `src` folder
-srcPath: ['./'],
+srcPath: './',
 ```
 
 ```tsx


packages/next-intl/src/extractor/source/SourceFileFilter.tsx
```diff
@@ -2,9 +2,41 @@ import path from 'path';
 
 export default class SourceFileFilter {
   static readonly EXTENSIONS = ['ts', 'tsx', 'js', 'jsx'];
+  private static readonly IGNORED_DIRECTORIES = ['node_modules', '.next'];
```
amannn:
```suggestion
  // Will not be entered, except if explicitly asked for
  private static readonly IGNORED_DIRECTORIES = ['node_modules', '.next'];
```

docs/src/pages/docs/usage/plugin.mdx
```diff
@@ -128,7 +128,7 @@ If your project is split into multiple folders, you can provide an array of path
 
 ```tsx
 // Not using a `src` folder
-srcPath: ['./app', './components'],
+srcPath: ['./'],
```
amannn:
```suggestion
srcPath: './',
```

PR #2094, opened 2025-11-11
fix: Move AST transformer of `useExtracted` to SWC plugin & handle source maps

amannn:


packages/next-intl/src/extractor/extractor/MessageExtractor.tsx
```diff
@@ -54,28 +63,41 @@ export default class MessageExtractor {
...
+      absoluteFilePath,
+      relativeFilePath
+    );
 
     const finalResult = (
       processResult.source ? processResult : {...processResult, source}
-    ) as {messages: Array<StrictExtractedMessage>; source: string};
+    ) as {
+      messages: Array<StrictExtractedMessage>;
+      source: string;
```
amannn: @cursoragent I don't want to call this `source`, rename consistently across package to `code`.

amannn: @cursoragent It was a bit too much. Keep it "source" for when it's source in the sense that it's actual source (i.e. the user writes this). Call it generic "code" if it's something we return from a compilation.

packages/next-intl/src/extractor/extractor/MessageExtractor.tsx
```diff
@@ -505,11 +527,27 @@ export default class MessageExtractor {
...
+    });
+
+    // Fix the source map to include the correct filename
+    let map = output.map;
+    if (map && this.sourceMap) {
+      map = map.replace(
+        '{"version":3,"sources":["<anon>"]',
+        `{"version":3,"file":"${filePath}","sources":["${absoluteFilePath}"]`
+      );
+    }
```
amannn: Investigate if we can avoid patching this and instead use a public API from SWC.

Ideas:
- Use `swc.transform` instead
- Some kind of config for print

Try both approaches, see what works, and compare both approaches with `MessageExtractor.bench.tsx`

packages/next-intl/src/extractor/extractor/MessageExtractor.test.tsx
```diff
@@ -34,8 +37,15 @@ it('can extract a simple message', async () => {
     )
   ).toMatchInlineSnapshot(`
     {
+      "code": "import { useTranslations } from 'next-intl';
+    function Component() {
+        const t = useTranslations();
+        t("+YJVTi", void 0, void 0, "Hey!");
+    }
+    ",
```
amannn: - [ ] Move (most) tests to Rust

packages/next-intl/src/extractor/extractor/MessageExtractor.tsx
```diff
@@ -51,465 +45,82 @@ export default class MessageExtractor {
...
-                visit(item.expression as Node);
-              } else if ('type' in item) {
-                visit(item as Node);
+  ): Promise<{
+    messages: Array<StrictExtractedMessage>;
+    code?: string;
+    map?: string;
+  }> {
+    // TODO: Depend on this package and use require.resolve to get the path
+    const swcPluginPath = path.join(
```
amannn: - [ ] Use package name

packages/next-intl/src/extractor/extractor/MessageExtractor.test.tsx
amannn, line 1: @cursoragent These tests should be moved to Rust in swc-plugin. At least the vast majority which test simple transforms. Leave some strategic ones here as an integration test and move the others. See the test structure with fixture/await-get-extracted for example. Use succinct folder names per test.

packages/next-intl/src/extractor/extractor/MessageExtractor.tsx
```diff
@@ -51,465 +46,50 @@ export default class MessageExtractor {
...
-        }
-      }
-    }
-
-    visit(ast);
+    // TODO: Improve the typing of @swc/core
+    const output = (result as any).output as string;
+    const messages = JSON.parse(
+      JSON.parse(output).results
+    ) as Array<StrictExtractedMessage>;
```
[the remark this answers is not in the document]
amannn: @cursoragent Use a simple try/catch here for both steps without this elaborate ceremony that's outlined here. Use a simple error message just in case this happens.

PR #2101, opened 2025-11-12
fix(useExtracted): Fix inconsistent translation message ordering

amannn:
Use message ID as a tie-breaker in `getSortedMessages` in case reference paths match for a given message.

The previous tie-breaker, `a.message.localeCompare(b.message)`, was ineffective for untranslated messages (where `message` is an empty string), causing `getSortedMessages` to return `0` and rely on non-deterministic insertion order from parallel file extraction. Another issue is that `message` can be different across locales, therefore the order could vary across catalogs.

This fix ensures consistent ordering for messages, particularly those used in multiple files, by using a stable `message.id` for tie-breaking.

### DIFF
--- a/packages/next-intl/src/extractor/formatters/utils.test.tsx
+++ b/packages/next-intl/src/extractor/formatters/utils.test.tsx
@@ -1,26 +1,43 @@
-import {expect, it} from 'vitest';
-import type {ExtractedMessage} from '../types.js';
+import {describe, expect, it} from 'vitest';
 import {getSortedMessages} from './utils.js';
 
-it('uses message ids to break ties when reference paths match', () => {
-  const messages: Array<ExtractedMessage> = [
-    {id: 'beta', message: '', references: [{path: 'components/Footer.tsx'}]},
-    {id: 'alpha', message: '', references: [{path: 'components/Footer.tsx'}]},
-    {id: 'gamma', message: '', references: [{path: 'components/Header.tsx'}]}
-  ];
+describe('getSortedMessages', () => {
+  it('sorts by reference path', () => {
+    expect(
+      getSortedMessages([
+        {
+          id: 'a',
+          message: 'a',
+          references: [{path: 'components/B.tsx'}]
+        },
+        {
+          id: 'b',
+          message: 'b',
+          references: [{path: 'components/A.tsx'}]
+        }
+      ]).map((message) => message.id)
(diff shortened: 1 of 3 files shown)


packages/next-intl/src/extractor/formatters/utils.tsx
```diff
@@ -3,13 +3,14 @@ import type {ExtractedMessage} from '../types.js';
...
-  return messages.toSorted((a, b) => {
-    const aPath = a.references?.[0]?.path ?? a.message;
-    const bPath = b.references?.[0]?.path ?? b.message;
-    if (aPath === bPath) {
-      return a.message.localeCompare(b.message);
-    } else {
-      return aPath.localeCompare(bPath);
+  return messages.toSorted((messageA, messageB) => {
+    const pathA = messageA.references?.[0]?.path ?? messageA.message;
+    const pathB = messageB.references?.[0]?.path ?? messageB.message;
```
amannn: @cursoragent The fallback here is purely theoretical, we always have one. Use `''` instead of the message to avoid confusion.

amannn:
@cursoragent Also investigate and fix failing tests.

packages/next-intl/src/extractor/formatters/utils.test.tsx
```diff
@@ -0,0 +1,28 @@
+import {describe, expect, it} from 'vitest';
+import type {ExtractedMessage} from '../types.js';
+import {getSortedMessages} from './utils.js';
+
+describe('getSortedMessages', () => {
```
amannn: @cursoragent Remove describe block (just put "it" on top level)

packages/next-intl/src/extractor/formatters/utils.test.tsx
```diff
@@ -0,0 +1,28 @@
...
+      {id: 'alpha', message: '', references: [{path: 'components/Footer.tsx'}]},
+      {id: 'gamma', message: '', references: [{path: 'components/Header.tsx'}]}
+    ];
+
+    const sorted = getSortedMessages(messages).map((message) => message.id);
+
+    expect(sorted).toEqual(['alpha', 'beta', 'gamma']);
+  });
+
+  it('sorts by id when messages share the same inferred path', () => {
```
amannn: @cursoragent Remove this test and instead add one that shows that we sort by reference and not be id if possible.

PR #2103, opened 2025-11-13
fix: Fix inconsistent ordering of messages for `useExtracted` pt. 2

amannn:
**Prerelease:**

```
next-intl@0.0.0-canary-8f1d542
```

---

https://github.com/amannn/next-intl/pull/2091 was a first step, but more work was necessary to really fix this.

Use message ID as a tie-breaker in `getSortedMessages` in case reference paths match for a given message.

The previous tie-breaker, `a.message.localeCompare(b.message)`, was ineffective for untranslated messages (where `message` is an empty string), causing `getSortedMessages` to return `0` and rely on non-deterministic insertion order from parallel file extraction. Another issue is that `message` can be different across locales, therefore the order could vary across catalogs.

This fix ensures consistent ordering for messages, particularly those used in multiple files, by using a stable `message.id` for tie-breaking. This now also ensures that the order is the same across all catalogs.

### DIFF
--- a/packages/next-intl/src/extractor/ExtractionCompiler.test.tsx
+++ b/packages/next-intl/src/extractor/ExtractionCompiler.test.tsx
@@ -686,15 +686,13 @@ describe('json format', () => {
     simulateFileEvent('/project/messages', 'rename', 'de.json');
 
     await waitForWriteFileCalls(2);
-    expect(vi.mocked(fs.writeFile).mock.calls.slice(1)).toMatchInlineSnapshot(`
+    expect(vi.mocked(fs.writeFile).mock.calls.at(-1)).toMatchInlineSnapshot(`
       [
-        [
-          "messages/de.json",
-          "{
+        "messages/de.json",
+        "{
         "7kKG3Q": "",
         "OpKKos": ""
       }",
-        ],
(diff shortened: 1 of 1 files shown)


amannn:
@ceolinwill and @johanneskares: You two reported that ordering of messages wasn't deterministic yet, thanks again for the helpful reports and supplemental context! 🙏 

I found a bug and I think now everything should work as expected. Can you try `next-intl@0.0.0-canary-8f1d542` in your apps and report back here if everything works fine now?

Note that you might see an initial change after installing this version, but all subsequent extractions should match the previous one now. If this looks good to you, I'll release this to stable.

ceolinwill:
@amannn Thanks for looking into this. It's working as expected now.

amannn:
@ceolinwill Thanks, that's great to hear! @johanneskares gave me some details about a case where there is apparently still a bug happening. I'll see if I can reproduce this before merging this fix.

amannn:
Ok, false alarm—also @johanneskares is getting deterministic results now. Will publish to stable, thanks everyone!

johanneskares:
amazing, thanks @amannn 🚀

PR #2115, opened 2025-11-16
feat: convert `t.has` to type guard

### LINKED ISSUE #2104: `t.has` should act as a type guard

When getting translations from `useTranslations` hook via keys of broad or unknown type in Typescript, it's highly likely to run into an issue that the key cannot be used unless its type is narrowed down to `IntlMessages` keys. There is a helpful `t.has` method that checks if the given key exists, but as it only accepts `IntlMessages` keys and returns `boolean`, hence the variable needs to be cast twice:

```ts
const t = useTranslations();
const key: string = 'foo';

// Error - Argument of type 'string' is not assignable to parameter of type MessageKeys...
if (t.has(key)) {
  // Error - Argument of type 'string' is not assignable to parameter of type MessageKeys...
  return t(key);
}
```

`t.has` can accept a broader type (`string` perhaps?) and act as [a type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards):

```ts
function has(key: string): key is keyof MessageKeys
```

Right now, I've written the following wrapper to achieve it:

```ts
export const hasTranslation = <TranslateFn extends ReturnType<typeof useTranslations>>(
  t: TranslateFn,
  key: string
): key is Parameters<typeof t>[0] => t.has(key as any);
```

G07cha:
Resolves #2104

### DIFF
--- a/packages/use-intl/src/core/createTranslator.test.tsx
+++ b/packages/use-intl/src/core/createTranslator.test.tsx
@@ -163,8 +163,6 @@ describe('type safety', () => {
         // @ts-expect-error
         t('unknown');
         // @ts-expect-error
-        t.has('unknown');
-        // @ts-expect-error
         t.markup('unknown');
         // @ts-expect-error
         t.rich('unknown');
@@ -195,12 +193,26 @@ describe('type safety', () => {
         // @ts-expect-error
         t('unknown');
         // @ts-expect-error
-        t.has('unknown');
-        // @ts-expect-error
         t.markup('unknown');
         // @ts-expect-error
         t.rich('unknown');
       };
+    });
+
+    it('allows dynamic keys after they were checked by t.has', () => {
+      const t = createTranslator({
+        locale: 'en',
+        messages: {title: 'test'}
+      });
+      const key = 'title' as string;
+
+      if (t.has(key)) {
+        expect(t(key)).not.toBeUndefined();
+      } else {
+        expect.fail(
+          `\`t.has\` was not able to find ${key}, check if it still exists in messages`
+        );
+      }
     });
   });
 
@@ -610,7 +622,10 @@ describe('type safety', () => {
       () => {
         t('param', {unknown: 'Jane'});
         t.rich('param', {unknown: 'Jane', p: (chunks) => <p>{chunks}</p>});
-        t.markup('param', {unknown: 'Jane', p: (chunks) => `<p>${chunks}</p>`});
+        t.markup('param', {
+          unknown: 'Jane',
+          p: (chunks) => `<p>${chunks}</p>`
+        });
       };
     });
 
@@ -880,5 +895,27 @@ describe('t.raw', () => {
     });
 
     expect(t.raw('rich')).toBe('<b>Hello <i>{name}</i>!</b>');
+  });
+});
+
+describe('t.has', () => {
+  it('returns true when key exists in messages', () => {
+    const t = createTranslator({
+      locale: 'en',
+      namespace: 'Home',
+      messages
+    });
+
+    expect(t.has('title')).toBe(true);
+  });
+
+  it('returns false when key does not exist in messages', () => {
+    const t = createTranslator({
+      locale: 'en',
+      namespace: 'Home',
+      messages
+    });
+
+    expect(t.has('invalid')).toBe(false);
   });
 });
--- a/packages/use-intl/src/core/createTranslator.tsx
+++ b/packages/use-intl/src/core/createTranslator.tsx
@@ -139,9 +139,7 @@ export type Translator<
   ): any;
 
   // `has`
-  has<TargetKey extends NamespacedMessageKeys<TranslatorMessages, Namespace>>(
-    key: TargetKey
-  ): boolean;
+  has(key: string): key is NamespacedMessageKeys<TranslatorMessages, Namespace>;
 };
 
 /**
--- a/examples/example-app-router-playground/src/components/AsyncComponent.tsx
+++ b/examples/example-app-router-playground/src/components/AsyncComponent.tsx
@@ -32,6 +32,8 @@ export async function TypeTest() {
   // @ts-expect-error
   t.markup('unknown');
 
-  // @ts-expect-error
-  t.has('unknown');
+  const key = 'unknown' as string;
+  if (t.has(key)) {
+    t(key);
+  }
 }


packages/use-intl/src/core/createTranslator.test.tsx
```diff
@@ -195,13 +193,27 @@ describe('type safety', () => {
...
         t.markup('unknown');
         // @ts-expect-error
         t.rich('unknown');
       };
     });
+
+    it('allows dynamic keys after they were checked by t.has', () => {
+      const t = createTranslator({
+        locale: 'en',
+        messages: {title: 'test'}
```
G07cha, line 205: Using a different set of messages here, as the current one contains parameterized messages so the type will have to be narrowed down further before any of the keys can be used.

packages/use-intl/src/core/createTranslator.tsx
```diff
@@ -139,9 +139,7 @@ export type Translator<
   ): any;
 
   // `has`
-  has<TargetKey extends NamespacedMessageKeys<TranslatorMessages, Namespace>>(
-    key: TargetKey
-  ): boolean;
+  has(key: string): key is NamespacedMessageKeys<TranslatorMessages, Namespace>;
```
amannn, line 142: Ah, that would be the only minor change: You don't get type-safety anymore for the `key` that's passed here.

I was in fact considering `string` for the initial implementation, but went for the stricter variant.

I'd be a non-breaking change to adjust this, I can ask some other users of `t.has` if this matches their understanding. But let's see if there's a solution to the other comment first.

packages/use-intl/src/core/createTranslator.test.tsx
amannn: Isn't that what would happen in basically any non-trivial app? How would you do that in userland?

This seems to fail as soon as I add this:

```suggestion
        messages: {title: 'test', foo: '{foo} bar'}
```

G07cha: This happens because all values of `messages` object are treated as constants. When importing values from JSON, the leaf nodes will just have `string` value types unless [the experimental messages declarations generation](https://next-intl.dev/docs/workflows/typescript#messages-arguments) is enabled. 

It is a trade-off of a stricter typing, requiring the developer to manually narrow down the types if one of the keys in the namespace has parameters:

```ts
if(t.has(key) && key !== 'foo') {
  t(key);
}
```

G07cha:
Updated usage in `example-app-router-playground` to fix previously failing build in the CI pipeline.

packages/use-intl/src/core/createTranslator.test.tsx
amannn: Yep, it doesn't work well with strict messages. I'd hope that at some point https://github.com/microsoft/TypeScript/issues/32063 will be resolved, so it could be that strict messages might be the default at some point.

> `key !== 'foo'` 

How would you do this in a real-word codebase?

G07cha: In our case, we're not using messages declarations for strictly typed parameters yet, but even then, the usage of parameters is rather sparse in our translations. We usually resort to checking keys in runtime when they come from an external source (e.g. API), then we define a separate object for them:

```ts
// messages.json:
// {
//   ...
//   "status": {
//     "success": "Success",
//     "failure": "Failed",
//     "unknown": "Unknown"
//   }
//   ...
// }

const response = await fetch('https://myapi.com')
const data = await response.json();
const translationKey = `status.${data.result}`

return t.has(translationKey) ? t(translationKey) : t('status.unknown');
```

amannn: > the usage of parameters is rather sparse in our translations

But it will cause trouble with every argument that you add. Even if fixed with the above approach, this will break again for every new message that you add that uses arguments. Maintaining this for non-trival apps just doesn't work out.

Is there any way to avoid this?



> We usually resort to checking keys in runtime when they come from an external source (e.g. API), then we define a separate object for them

Right, my stance here is that you can avoid these problems like this:

```tsx
const message = {
  success: t('success'),
  failure: t('failure')
}[data.result] || t('unknown');
```

G07cha: > Is there any way to avoid this?

To be frank, I don't know. This works for our codebase, but you made a clear case for when it wouldn't. Fundamentally, we're adding more type safety by converting `t.has` into a type guard, which inevitably may cause type errors in the applications that previously weren't covering all scenarios.

amannn: Yes, but without having any ergonomic way of using this with strict arguments, I think this is unfortunately not something we can realistically change—sorry :-/. Maybe the workaround you shared in the issue is the better solution here for apps which don't use strict arguments.

G07cha: No worries, thanks for taking the time to review and provide the feedback!

G07cha:
Closing as there doesn't seem to be an ergonomic way to handle messages with typed arguments, which makes it difficult to use.

packages/use-intl/src/core/createTranslator.test.tsx
amannn: Sure, thank you too for your effort in exploring this!

packages/use-intl/src/core/createTranslator.tsx
AndyClausen: Just wanted to chime in here: I don't think the decision to use strict key types here is the best. If you are sure that the key is the correct type, then there is no point to calling `t.has` - you already know the result ahead of runtime. The only reason to use `t.has` currently is either if you aren't using typed messages or if you're casting the type in `t.has` (and any other calls to `t`). In either case, `key` might as well just be of type `string`.

This is why I really dislike the current implementation of `Array.includes` in TS as well, it has the same problem when using const values.

I know this PR has been closed now, but I wanted to point this out in case it's picked up again.

amannn: >  If you are sure that the key is the correct type, then there is no point to calling t.has

It depends a bit, you know that it exists for the locale where you're sampling your `Messages` type from. The message could still be missing for another locale.

AndyClausen: That's true, it can be used to add a fallback for other locales, but it makes it less useful with this restriction. Maybe the better solution would be to add a `t.exists` method, which could take in string or unknown and use a type predicate to narrow the type to `keyof Messages` or a more complex type that includes nested keys.

I know that dynamic keys are discouraged, and I understand and agree on the why's, but sometimes it's unavoidable, and then we have to either type cast with `t.has` or make a custom type predicate ourselves to make sure the translation will show correctly. A built-in solution would definitely be preferred.

PR #2143, opened 2025-12-01
fix: Skip `accept-language` parsing when locale cookie is already present

### LINKED ISSUE #2116: next-intl middleware extremely slow with large number of locales (~140+)
### Description

When using `createMiddleware(routing)` with a large number of locales (100+), the next-intl middleware becomes extremely slow even in a minimal setup.

With ~140 locales in `routing.locales`, the middleware takes ~120–180 ms per request.  
With only a few locales (<5), it runs in ~1–3 ms.

This slowdown happens:

- before any dictionary loading
- without any additional middleware logic
- in both dev and production builds
- with or without other integrations (Supabase, Auth.js, etc.)

This suggests that locale negotiation or internal matching scales linearly with the number of locales. Because only prefix detection + cookie + Accept-Language is needed, the middleware should not degrade significantly with a larger locale list.

This makes next-intl difficult to use in multilingual applications that support many regional variants (e.g. `fr-FR`, `de-CH`, `en-GB`, etc.).

### Verifications

https://github.com/lxup/next-intl-middleware-slow-repro

### Reproduction description

Steps to reproduce:

1. Open the reproduction repository.
2. Notice that `supportedLocales` contains ~140 generated locales.
3. Start the dev server or run a production build.
4. Load any page and check the console output in the server logs.

Observed:
- The line `Intl middleware took: XXX ms` consistently reports ~150 ms.
- Reducing the locale list to 3 items makes the middleware drop to ~1–2 ms.

### Expected behaviour

The next-intl middleware should remain performant regardless of the number of supported locales.

Locale negotiation should be fast because it normally involves:

1. Detecting a prefix
2. Reading a cookie
3. Optionally parsing Accept-Language

None of these operations require scanning every locale or performing heavy operations.

Expected:
- Middleware execution time should remain close to ~1 ms even with many locales (100+).
- Locale negotiation should not scale linearly with the number of locales.
- Apps supporting many regional locales should not suffer a 100–150 ms overhead per request.

lxup:
Fixes #2116

### Description
This PR addresses a severe performance bottleneck in the `syncCookie` middleware logic where `getAcceptLanguageLocale` (and consequently locale negotiation) was called unconditionally on every request.

As detailed in the linked issue, this caused linear performance degradation as the number of locales increased (~130ms overhead for ~140 locales).

### Changes
I optimized the `syncCookie` function to verify the cookie state *before* attempting to derive the locale from headers.
- **Before:** Header parsing & locale negotiation ran on every request.
- **After:** Header parsing only runs if:
1. No locale cookie exists.
2. OR the cookie is outdated/invalid.

### Benchmarks
Tested with ~140 locales in the reproduction repository provided in the issue.

**Before:**
```text
[next-intl-timing] syncCookie: 130ms
[next-intl-timing] total: 132ms
```
**After:**
```text
[next-intl-timing] syncCookie: 1ms
[next-intl-timing] total: 3ms
```

**Checklist**
- [x] I have verified that the issue is resolved locally.
- [x] I have verified that existing tests pass.

### DIFF
--- a/packages/next-intl/src/middleware/syncCookie.tsx
+++ b/packages/next-intl/src/middleware/syncCookie.tsx
@@ -38,19 +38,26 @@ export default function syncCookie<
   if (!routing.localeCookie) return;
 
   const {name, ...rest} = routing.localeCookie;
-  const acceptLanguageLocale = getAcceptLanguageLocale(
-    request.headers,
-    domain?.locales || routing.locales,
-    routing.defaultLocale
-  );
   const hasLocaleCookie = request.cookies.has(name);
   const hasOutdatedCookie =
     hasLocaleCookie && request.cookies.get(name)?.value !== locale;
 
-  if (hasLocaleCookie ? hasOutdatedCookie : acceptLanguageLocale !== locale) {
+  if (hasOutdatedCookie) {
     response.cookies.set(name, locale, {
       path: request.nextUrl.basePath || undefined,
       ...rest
     });
+  } else if (!hasLocaleCookie) {
+    const acceptLanguageLocale = getAcceptLanguageLocale(
+      request.headers,
+      domain?.locales || routing.locales,
+      routing.defaultLocale
+    );
+    if (acceptLanguageLocale !== locale) {
+      response.cookies.set(name, locale, {
+        path: request.nextUrl.basePath || undefined,
+        ...rest
+      });
+    }
   }
 }


amannn (approved):
Careful analysis + simple fix + a lot of impact = dream PR — thanks a bunch @lxup ❤️

lxup:
@amannn thx! That was my first PR ever 😅

PR #2154, opened 2025-12-03
fix: Fail loud if messages can't be read for usage with `useExtracted`

amannn:
Prevent silent translation wipes by propagating read errors instead of returning empty arrays.

Previously, `loadLocaleMessages()` would silently catch all errors during file reads and return an empty array. This empty array would then be used by `saveLocale()` to overwrite all existing translations with empty strings. This change ensures that only `ENOENT` (file not found) errors result in an empty array (expected for new locales), while other read errors (e.g., file corruption, I/O issues) are now propagated, preventing accidental data loss.

---
<a href="https://cursor.com/background-agent?bcId=bc-5c8e526e-4861-411d-b241-a73664d14fa6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-cursor-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-cursor-light.svg">Open in Cursor</picture></a>&nbsp;<a href="https://cursor.com/agents?id=bc-5c8e526e-4861-411d-b241-a73664d14fa6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-web-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-web-light.svg">Open in Web</picture></a>

---

> [!NOTE]
> Propagates non-ENOENT read/parse errors when loading message catalogs (only ENOENT returns empty), with tests and mocks updated to prevent silent translation wipes.
> 
> - **Extractor/Catalog loading**:
>   - `CatalogManager.loadLocaleMessages`: remove silent catch; propagate errors from persister.
>   - `CatalogPersister`:
>     - Return `[]` only on `ENOENT` in `read`; otherwise throw `Error while reading ...`.
>     - Throw `Error while parsing ...` on formatter parse failures.
>     - Add `getFileName` and use in path/error messages.
> - **Tests**:
>   - `ExtractionCompiler.test.tsx`:
>     - Add tests asserting propagation of read errors and parser errors; ensure only `ENOENT` yields empty.
>     - Enhance `fs/promises` mock to throw realistic `ENOENT` and support read interceptors.
>   - `MessageExtractor.test.tsx`: update inline snapshots (source map mapping and aliased `useTranslations`).
> 
[...]

Issue #2083, opened 2025-11-07
Add optional support for generating psuedo localization


priyajeet:
### Is your feature request related to a problem? Please describe.

In order to check the UI for localized overflow problems (string that may break the UI in another language), but still keeping the words mostly english (or default locale) having a psuedo localization helps. If we have to use something today, it needs to be a build step with a script running after looking and reading the en.json source.

Source
```json
{
  "Foo": "Foo",
  "Bar": "Bar"
}
```
Psuedo localized
```json
{
  "Foo": "[### Ƒőō ###]",
  "Bar": "[### Ɓǻŕ ###]"
}
```

### Describe the solution you'd like

With the new auto-extraction it would be nice if a new optional psuedo localized locale is also written alongside the en.json. A really old library that does psuedo localization is https://www.npmjs.com/package/pseudoloc. If some logic of this can be brought over or a hook given to call into a 3P library/function, both work. But since this can be done at the time of message extraction, having it inbuilt will be nice.

### Describe alternatives you've considered

Write your own pre-commit hook script that converts en.json to en-x-psuedo.json

amannn:
Custom formatters are on the roadmap for `useExtracted` (see also [this discussion](https://github.com/amannn/next-intl/discussions/2036#discussioncomment-14923974)).

I guess you could add one that when writing to `en-x-pseudo.json` auto-populates the text, I made a note about this in https://github.com/amannn/next-intl/issues/2087.

PR #2155, opened 2025-12-03
feat: Custom formats for `useExtracted`, consistency fixes for file references, pruning of messages and sorting of keys

amannn:
**Notes**

- Docs: [**`useExtracted`: Custom format**](https://next-intl-docs-git-canary-next-intl.vercel.app/docs/usage/extraction#formats-custom)
- Examples: [`codecs`](https://github.com/amannn/next-intl/tree/canary/packages/next-intl/src/extractor/format/codecs) (see also the `fixtures`)
- Prerelease: `next-intl@0.0.0-canary-6876534`

**Changes**

- https://github.com/amannn/next-intl/pull/2158
- https://github.com/amannn/next-intl/pull/2154
- https://github.com/amannn/next-intl/pull/2163
- https://github.com/amannn/next-intl/pull/2166
- https://github.com/amannn/next-intl/pull/2169
- https://github.com/amannn/next-intl/pull/2170

---

> [!NOTE]
> Introduces codec-based message formats (built-in JSON/PO + custom), overhauls the `useExtracted` pipeline with file watching, consistent reference handling and pruning, and updates docs, tests, plugin loaders, examples, workflows, and dependencies.
> 
> - **Extractor/Core**:
>   - Add codec system (`defineCodec`) with built-in `json`/`po` codecs and support for custom formats; expose via `next-intl/extractor`.
>   - Refactor extraction flow: new `extractAll`, `CatalogManager` improvements (metadata merge, reference dedupe/sort, pruning obsolete messages), robust error handling, and source watching via `@parcel/watcher`.
>   - Update loaders (`extractionLoader`, `catalogLoader`) and format resolution (`getFormatExtension`, `resolveCodec`).
>   - Expand tests extensively for JSON/PO/custom formats, references, races, and errors; add `SourceFileWatcher`.
> - **Docs**:
>   - Revise `useExtracted` docs with Formats section (JSON/PO/Custom) and examples; minor plugin/config wording tweaks.
> - **Examples**:
>   - Rename example `Client` → `Counter`; increase Playwright retries on CI.
> - **Tooling/CI**:
>   - Adjust release/prerelease workflows (setup-node config, trusted publishers note).
>   - Build: add CJS `chunkFileNames` in Rollup.
> - **Deps**:
>   - Add `@parcel/watcher`; bump `po-parser` to v2; update Lerna and conventional-changelog packages.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d43359c83737c5340d413eeb20c3788290855f72. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/next-intl/src/extractor/source/SourceFileWatcher.tsx
+++ b/packages/next-intl/src/extractor/source/SourceFileWatcher.tsx
@@ -23,12 +23,17 @@ export default class SourceFileWatcher implements Disposable {
     for (const root of this.roots) {
       const sub = await subscribe(
         root,
-        async (err, events) => {
+        (err, events) => {
           if (err) {
-            console.error('next-intl source watcher error:', err);
+            console.error(err);
             return;
           }
-          await this.onChange(events);
+          const filteredEvents = events.filter((event) =>
+            SourceFileFilter.isSourceFile(event.path)
+          );
+          if (filteredEvents.length > 0) {
+            void this.onChange(filteredEvents);
+          }
         },
         {ignore}
       );


amannn:
I'm tagging a few people who have already expressed interest in this feature: @ceolinwill, @kieranm, @priyajeet

This PR adds support for custom formats, so with this you can implement e.g.:
- Any JSON-based file format that persists metadata like a description and file references
- A modified version of the built-in PO format (e.g. using the source message as `msgid`)
- Auto-population of certain locales like `en-x-pseudo.json` when the file is written (related to https://github.com/amannn/next-intl/issues/2083)

I think the feature should be sound and I'm looking to merge it early next week. If some of you are interested in having an early look and let me know if this looks good to you, I'd be really grateful for that. Esp. API-related feedback is welcome as this is harder to change later. Please see the PR description for docs and a prerelease.

As a side note, as I've already mentioned in the related discussion, I'm thinking about building some kind of tooling like a VSCode extension that helps with tasks like manual editing of messages across locales and potentially even AI-based translations. I've written a bit on this in https://github.com/amannn/next-intl/discussions/2036#discussioncomment-14923974 and I hope that provides some background on why e.g. the default PO format uses the generated key as `msgid`. I think once the tooling is better in this area that should make the default PO format even more useful. But that being said, you're of course welcome to use whatever format you prefer and use custom tooling as desired!

Let me know what you think!

ceolinwill:
Thanks for working on this.

I’ll review it more carefully later, but at first glance it looks good.

I noticed you added two custom codecs as test fixtures. Would it make sense to expose them as supported formats (in addition to the default json and po) or at least export them so others can reuse them?

The custom PO codec is exactly what I need. Otherwise, I’d end up copying it into my own codebase. I’d prefer importing it from next-intl and contributing fixes upstream if anything comes up.

Totally fine if you don’t want to support this codec in the main library, though. Just wanted to raise the idea.

ceolinwill:
I just tested it and it works well. This is exactly what I needed.

If you ever decide to export the `POCodecSourceMessageKey` codec, let me know. There's a bug in there that would need fixing.

It would also help to expose some internal utilities like `getSortedMessages` and `setNestedProperty`, so they don’t need to be copy-pasted when building a custom codec like `POCodecSourceMessageKey`.

amannn:
@ceolinwill I see your point! And thanks for the early feedback!

I think people will have quite a bit of opinions on how to configure the PO codec, e.g. different meta content, sorting of messages, message ids … I think it will be pretty tough to support the permutations of this with a descriptive name and without the library getting bloated. So another approach would be to allow configuration, but this is pretty limited as we need to serialize over a Turbopack loader bridge (no functions, …).

What I did for the PO codec is to extract the meat to [`po-parser`](https://github.com/amannn/po-parser/). My hope is that with this being available, the codec itself should be a rather simple mapping layer. So I hope it shouldn't feel too cumbersome to configure a bit of this for developers, while in turn giving you a lot of flexibility. And we can still share fixes via the `po-parser` dependency in case it's relevant to all implementations!

But that being said, it's certainly ok to publish codecs as standalone libraries in case people share a lot of preferences. And also, we could still decide to include other formats/variants in the future. As it's still early for this feature, I'd like to explore preferences of a people a bit more before adding APIs that are hard to change in the future. For the time being, I think it might be nice that there's a single OOTB codec per file extension, so the typical developer never has to think about what a "codec" is.

Does that sound right to you?

On a different note, I just noticed that the `toJSONString` method might need a minor overhaul to allow for better performance, I made a note about that in the PR comment …

amannn:
Oh, just commented at the same time! :D

> There's a bug in there that would need fixing.

Can you tell me which one?

`setNestedProperty` is btw. pretty much lodash/set, I just wanted to avoid the dependency.

ceolinwill:
> Can you tell me which one?

When there's a dot, it splits the message into `msgctxt` and `msgid`:

```po
msgctxt "The code you entered is incorrect. Please try again or contact hello@zoonk"
msgid "com"
```

Similarly, when a sentence ends with a dot, it removes the last dot, moves that sentence to `msgctxt` and keeps `msgid` empty:

```po
msgctxt "Checking if you're logged in.."
msgid ""
```

amannn:
> When there's a dot, it splits the message into `msgctxt` and `msgid`

@ceolinwill Ah yep, that's a great point! I've also discovered that [namespaces](https://next-intl.dev/docs/usage/extraction#namespaces) were not supported by the custom PO codec. I've now fixed both of these bugs in https://github.com/amannn/next-intl/pull/2163.

This also required [a change in `po-parser`](https://github.com/amannn/po-parser/pull/4) which removed some hardcoded logic from the shared library that was only relevant to the default PO codec. In turn, keys can now be persisted e.g. to `msgctxt` to avoid regenerating them during `decode`.


- Latest fixture: [`POCodecSourceMessageKey.tsx`](https://github.com/amannn/next-intl/blob/canary/packages/next-intl/src/extractor/format/codecs/fixtures/POCodecSourceMessageKey.tsx) 
- Latest prerelease: `next-intl@0.0.0-canary-6876534`
- ([uses `po-parser@2.0.0-alpha.2`](https://github.com/amannn/next-intl/blob/6876534a0851e0c42d58b76805f677944c579a89/packages/next-intl/package.json#L131C6-L131C33))

Could you give this another go? If this looks fine to you, I can publish `po-parser@2` along with this PR.

My comment above about `toJSONString` turned out to be false alarm, I think we're good here.

ceolinwill:
> keys can now be persisted e.g. to `msgctxt` to avoid regenerating them during `decode`.

@amannn Great idea.

The formatting issues I had before are gone. Running `next build` now correctly puts the id in `msgctxt` and the source in `msgid`.

In my custom codec I’d likely invert that: keep the id in `msgid` and put the source in `msgctxt`, since the source text is mainly useful for context when reading PO files. It feels more natural:

- `msgid` → id
- `msgctxt` → source/context

If the source message were placed in `msgctxt` by default, I probably wouldn’t need a custom codec at all.

---

I did notice a potential new issue but I'm not sure:

1. I deleted all PO contents to start fresh with the new codec.
2. `next build` generated the PO files correctly.
3. I ran `pnpm dev` (which runs `next dev` across all apps). Everything looked fine.
4. After navigating through a few pages, the 3rd page threw: `Error: msgctxt is required`.
5. Checking the PO files, all `msgctxt` entries were gone and the id had been moved back to `msgid`.

Not sure yet if this is a real bug because I couldn't reproduce it again. I didn’t clear `.next`, so maybe some stale cache caused it. I’ll report back if it happens again.

amannn:
@ceolinwill Thanks for the feedback!

> The formatting issues I had before are gone.

Nice! 👍 

> In my custom codec I’d likely invert that: keep the id in `msgid` and put the source in `msgctxt`, since the source text is mainly useful for context when reading PO files. It feels more natural:
> 
> * `msgid` → id
> * `msgctxt` → source/context

Yeah, that's surely fine too!

[The PO "spec"](https://www.gnu.org/software/gettext/manual/html_node/PO-File-Entries.html) has some hints here:

- `msgid` is described as "the original string" (that's also what some TMS interpret)
- On the other hand, `msgctxt` is used to disambiguate messages with the same original string. But still, both allow arbitrary strings.

Many i18n libs like `react-intl` for example also allow you to define an explicit ID for certain cases, which can be helpful to disambiguate identical strings—which is also [what `next-intl` offers](https://next-intl.dev/docs/usage/extraction#explicit-ids). And then, `msgid` is no longer the "original string", but an actual ID. Also, in this case, `msgctxt` becomes more like a namespace. AFAIK, originally, GNU gettext wasn't really designed to work with any IDs at all, so it needs some interpretation.





> Not sure yet if this is a real bug because I couldn't reproduce it again. I didn’t clear `.next`, so maybe some stale cache caused it. I’ll report back if it happens again.

Thanks! I also tried reproducing it in a project and investigated a bit, but no luck so far. If you encounter it again, definitely let me know—thanks!

amannn:
@ceolinwill I've updated this PR with a few fixes, some of them quite significant. It's only internal changes though.

This PR now also includes fixes for file references to be consistently updated and messages pruned when no longer used (both for builds as well as dev). Clearing `.next` or similar should surely not be necessary.

Would you mind giving this a quick sanity check in your app? That would be really helpful 🙏 Hopefully I can finally release to stable then, including the new feature for custom formats.

```
next-intl@0.0.0-canary-d4f4cde
```

ceolinwill:
@amannn Just tested it and everything looks great

- Deleting a file correctly removed the translation
- Renaming a file correctly updated the reference
- Tested several scenarios with `next dev` running for all apps in the monorepo while also running `next build`, and everything behaved as expected

One promising note:

In this monorepo, I upgraded only one app to the new version and left another on the old version.

When I ran `pnpm build` at the root, the app still on the old version showed a diff where some translations changed order. The app on the new version did not show this issue, which is a good sign.

Great job!

amannn:
Nice, that's really good to hear! Ok, let's get this over the finish line then …

PR #2166, opened 2025-12-10
fix: Update references of `useExtracted` after renames / deletions

amannn:


amannn:
@cursoragent Review this PR

packages/next-intl/src/extractor/ExtractionCompiler.test.tsx
```diff
@@ -1310,6 +1475,35 @@ msgstr "Hallo!"
...
+      return <div>{t('Message C')}</div>;
+    }
+    `;
+    filesystem.project.src['d.tsx'] = `
+    import {useExtracted} from 'next-intl';
+    export default function D() {
+      const t = useExtracted();
+      return <div>{t('Message B')}</div>;
+    }
+    `;
```
amannn: @cursoragent Take some redundancy out of this test with a fn like:

```tsx
createFile(componentName: string, message: string)

// E.g.
filesystem.project.src['d.tsx'] =  = createFile('D', 'Message B')
```

And then below, for all the `.compile` calls, write the file to filesystem and then reuse e.g. `filesystem.project.src['d.tsx']` for the compile call.

Also temporarily add some logging and verify the `waitForWriteFileCalls(1)` is what we want.

amannn: @cursoragent It's ok, but only add files to filesystem just before their corresponding compile invocation. Previous compilations shouldn't "see" the files that only become relevant later. Again, make sure write file calls number is up to date.

PR #2170, opened 2025-12-11
fix: Prune unused messages from `useExtracted` (during `dev`)

amannn:
## Problems discussed and identified
- Catalog entries stayed stale in dev when files were deleted because the extractor loader only handled new/changed files; deletions weren’t observed.
- Extraction and catalog persistence were coupled; transforms wrote catalogs, making dev flow and race handling harder.

## Changes implemented
- Introduced a `CatalogManager`-owned watcher (`SourceFileWatcher` with `@parcel/watcher`) to handle add/change/delete, treating `ENOENT` as no messages, pruning removed files, and saving catalogs.
- Loader now only transforms in dev using the shared extractor; persistence in dev is watcher-driven. Builds await the initial `extractAll` once.
- Shared extractor has shared caching for transforms/extracts, therefore sharing a single process is still beneficial.

### DIFF
--- a/packages/next-intl/src/extractor/catalog/CatalogManager.tsx
+++ b/packages/next-intl/src/extractor/catalog/CatalogManager.tsx
@@ -267,11 +267,7 @@ export default class CatalogManager {
 
       // Merge with previous message if it exists
       if (prevMessage) {
-        const validated = await this.validateExistingReferences(
-          message.id,
-          prevMessage.references ?? [],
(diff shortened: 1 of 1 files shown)


amannn:
@cursoragent Check the changes in this PR, esp. for suspicious code that might be left over or looks inconsistent with the rest of the changes.

amannn:
@cursoragent Review this PR again and leave a review here

Issue #2171, opened 2025-12-11
Locale cookie not set when visiting a URL matching the browser’s preferred locale → forces full locale negotiation on every request


lxup:
### Description

When no `NEXT_LOCALE` cookie exists, `next-intl` correctly falls back to `Accept-Language` to determine the user’s preferred locale.

However, when the detected locale _matches_ the browser’s preferred locale (e.g., browser = `fr-FR`, app supports `fr-FR`), visiting a localized URL such as:
```
http://localhost:3000/fr-FR
```
does not set a locale cookie.
In contrast, visiting a different locale such as:
```
http://localhost:3000/en-US
```
does set the cookie, because it differs from the browser’s preferred locale.

This leads to a regression in performance when using the optimization introduced in my previous PR (https://github.com/amannn/next-intl/pull/2143), where `syncCookie` avoids re-running `Accept-Language` parsing when a cookie is already present.

Since no cookie is stored in the “same-locale-as-browser” scenario, the optimization never applies and locale negotiation runs on every request, negating the improvement.

### Verifications

- [x] I've verified that the problem I'm experiencing isn't covered in [the docs](https://next-intl.dev).
- [x] I've searched for similar, existing issues on [GitHub](https://github.com/amannn/next-intl/issues) and [Stack Overflow](https://stackoverflow.com/search?q=next-intl).
- [x] I've compared my app to [a working example](https://next-intl.dev/examples) to look for differences.

### Mandatory reproduction URL

https://github.com/lxup/react-intl-cookie-not-set

### Reproduction description

Steps to reproduce:
1. Ensure no `NEXT_LOCALE` cookie is set.
2. Configure browser language to fr-FR.
3. Visit http://localhost:3000/fr-FR.
4. Observe: No locale cookie is created.
5. Reload a few times → Accept-Language parsing runs on every request.

### Expected behaviour

Visiting a locale-matching URL should still result in a locale cookie being stored so that the app benefits from the optimization in `syncCookie`. This would avoid repeated locale negotiation and restore the performance improvements from PR https://github.com/amannn/next-intl/pull/2143.

amannn:
Hey again, thanks for the question!

[...]

PR #2172, opened 2025-12-11
fix: Always set locale cookie when missing to ensure stable locale detection and preserve performance optimizations

lxup:
Fixes https://github.com/amannn/next-intl/issues/2171

## Description
This PR fixes a logic gap in `syncCookie` where the locale cookie was not being written if:
- No locale cookie existed, and
- The resolved locale matched the locale derived from the `Accept-Language` header.
This resulted in a situation where users whose browser language matched the app’s current locale (e.g. browser fr-FR and route /fr-FR) would never receive a locale cookie.

### Impact
Because no cookie was set in this scenario:
- Locale negotiation (`getAcceptLanguageLocale`) had to run on every request
- The performance benefit introduced in my previous PR (https://github.com/amannn/next-intl/pull/2143) — was lost for these users

This became especially visible in setups with many locales where the `Accept-Language` negotiation cost grows linearly.``

### Before
`syncCookie` only set the cookie if:
```tsx
acceptLanguageLocale !== locale
```
This prevented cookie creation whenever the header matched the active locale, which is a common and valid case.
### After
`syncCookie` now sets the cookie whenever it is missing, regardless of the Accept-Language header value:
- If cookie is outdated → update it
- If cookie is missing → write it
- No header logic involved in this decision
This aligns with the fact that the locale has already been resolved before reaching syncCookie, and the cookie must simply reflect that final state.

The `domain` parameter is also removed since it became unused after simplifying the logic.

### Result
- Stable locale persistence
- Correct behavior for first-time visitors
- Performance optimization preserved for all users
- Simpler and more predictable code

### Checklist
- [x] I have verified that the fix behaves correctly across all routing modes (always, never, as-needed).
- [x] I have confirmed that performance improvements from the previous PR remain active in all scenarios.
- [x] Existing tests pass successfully.

### DIFF
--- a/packages/next-intl/src/middleware/middleware.tsx
+++ b/packages/next-intl/src/middleware/middleware.tsx
@@ -319,7 +319,7 @@ export default function createMiddleware<
       }
     }
 
-    syncCookie(request, response, locale, resolvedRouting, domain);
+    syncCookie(request, response, locale, resolvedRouting);
 
     if (
       !hasRedirected &&
--- a/packages/next-intl/src/middleware/syncCookie.tsx
+++ b/packages/next-intl/src/middleware/syncCookie.tsx
@@ -5,13 +5,11 @@ import type {
   ResolvedRoutingConfig
 } from '../routing/config.js';
 import type {
-  DomainConfig,
   DomainsConfig,
   LocalePrefixMode,
   Locales,
   Pathnames
 } from '../routing/types.js';
-import {getAcceptLanguageLocale} from './resolveLocale.js';
 
 export default function syncCookie<
   AppLocales extends Locales,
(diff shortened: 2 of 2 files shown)


lxup:
Hi, I see the tests are failing. This is expected with the current implementation.

The goal here is to fix the performance issue from https://github.com/amannn/next-intl/issues/2171, where not setting a cookie (when the visited locale matches the browser's preference) forces expensive locale negotiation on every request.

My proposed change fixes this by always setting the cookie. The tests are failing because they currently assert the opposite behavior.

This raises a question: should we favor the performance gain and update the tests to reflect this new logic? I wanted to bring this up for discussion before proceeding with test modifications.

Let me know what you think

amannn:
Hey, I replied over here: https://github.com/amannn/next-intl/issues/2171#issuecomment-3647457854

PR #2180, opened 2025-12-16
fix: Consistently handle changes to entire folders for `useExtracted`

amannn:


amannn:
@cursoragent Add tests for extraction compiler in `ExtractionCompiler.test.tsx` based on changes in this pr. Cover the critical parts, don't overdo it with overly verbose test suites and assertions.

packages/next-intl/src/extractor/ExtractionCompiler.test.tsx
```diff
@@ -2840,6 +2838,121 @@ describe('custom format', () => {
...
+    using compiler = createCompiler();
+    await compiler.extractAll();
+    await waitForWriteFileCalls(1);
+
+    // Simulate deleting the directory
+    delete (filesystem.project.src as any).components;
+    fileTimestamps.delete('/project/src/components/Button.tsx');
+
+    const callback = parcelWatcherCallbacks.get('/project/src');
+    if (!callback) throw new Error('No watcher found for /project/src');
```
amannn: @cursoragent You can just assert non-null (`!`) here since we are in a test (check also other tests)

packages/next-intl/src/extractor/ExtractionCompiler.test.tsx
```diff
@@ -2840,6 +2838,121 @@ describe('custom format', () => {
...
+
+    callback(null, [{type: 'delete', path: '/project/src/components'}]);
```
amannn: @cursoragent There are prettier errors in this file, fix them.

packages/next-intl/src/extractor/ExtractionCompiler.test.tsx
```diff
@@ -2840,6 +2838,121 @@ describe('custom format', () => {
   });
 });
 
+describe('folder operations', () => {
```
amannn: @cursoragent move all these tests to .po format describe block, but you can use a nested describe block.

packages/next-intl/src/extractor/ExtractionCompiler.test.tsx
```diff
@@ -2840,6 +2838,121 @@ describe('custom format', () => {
...
+      {type: 'delete', path: '/project/src/old'}
+    ]);
+
+    await waitForWriteFileCalls(2);
+
+    expect(vi.mocked(fs.writeFile).mock.calls.at(-1)).toMatchInlineSnapshot(`
+      [
+        "messages/en.json",
+        "{
+        "cfI2fq": "Click me updated"
```
amannn: @cursoragent related to the other comment, when we use .po, we can assert file reference to be correct here.

amannn:
@cursoragent 

```
> next-intl@4.6.0 lint:source /Users/runner/work/next-intl/next-intl/packages/next-intl
> eslint src test && tsc --noEmit && pnpm run lint:prettier
/Users/runner/work/next-intl/next-intl/packages/next-intl/src/extractor/ExtractionCompiler.test.tsx
Error:   2552:14  error  'createCompiler' is already declared in the upper scope on line 790 column 12  @typescript-eslint/no-shadow
✖ 1 problem (1 error, 0 warnings)
```

be sure to run eslint and prettier before you commit, fix any issues.

do you even need to create another createCompiler function here? can we just reuse the original one?

PR #2183, opened 2025-12-16
fix: Gracefully handle errors for `useExtracted` parse failures

amannn:


packages/next-intl/src/extractor/catalog/CatalogManager.tsx
```diff
@@ -395,10 +395,12 @@ export default class CatalogManager implements Disposable {
...
       const extractStart = Date.now();
-      const extraction = await this.extractor.extract(
-        absoluteFilePath,
-        content
-      );
+      let extraction: Awaited<ReturnType<typeof this.extractor.extract>>;
+      try {
+        extraction = await this.extractor.extract(absoluteFilePath, content);
+      } catch {
+        return false;
```
amannn, line 402:
```suggestion
        // Treat parse errors as ignored files, wait until we get a sound update
        return false;
```

PR #2210, opened 2026-01-13
feat: Integrate `icu-minify` into `next-intl`

amannn:
<details>
<summary>Prompt</summary>

```
i've implemented @packages/icu-minify/README.md. we now want to get this into next-intl

what i imagine is an integration like this:

- in @packages/use-intl/ we have this import for intl-messageformat. mainly the one in @packages/use-intl/src/core/createBaseTranslator.tsx is crucial, the others are secondary.
- currently, compile and format happens both in @packages/use-intl/src/core/createBaseTranslator.tsx
- use-intl should continue working like it does currently
- however, in next-intl we can plug in icu-minify parser in @packages/next-intl/src/plugin/catalog/catalogLoader.tsx. this will compile all messages at build time. something to potentially later investigate is that we might need caching here to avoid recompiling all messages for every single reload, e.g. if a single message changes
- and then, we should only use the runtime feature of icu-minify when rendering messages that were precompiled in the catalog loader
- to do this, in PluginConfig in @packages/next-intl/src/plugin/types.tsx we could add a flag like messages.precompile: true.
- when this is done, the catalog loader should precompile
- now the challenge is to get the runtime from icu-minify in. we want this to work in useTranslations (across react-server and react-client, check @packages/next-intl/package.json) as well as getTranslations.
- the way i believe we can get this working is that in use-intl, we currently have two exports: /core and /react. i think we could add two more /format-message/compile-and-format and then /format-message/format (naming tbd, if you have better ideas let me know). now the first one would be the existing logic with intl-messageformat and the latter would utilize icu-minify.
- now, in @packages/use-intl/src/core/createBaseTranslator.tsx we need to call into 'use-intl/format-message/compile-and-format'. we need to ensure this is not compiled away when running build, you should check the dist folder when you're at this point. and then the trick is: in next-intl, we can register a webpack or turbopack alias that when the plugin precompile flag is on, then we change the import to only the /format variant. and tada, that should do the trick
- side note: you might need to refactor createBaseTranslator a bit for the compile/format part to be moved to a new module/import, so the rest of the code can be shared
- maybe use-intl could remove the dependency on intl-messageformat and we only depend on icu-minify? just to have a single codepath. but: evaluate if this has perf tradeoff. and also something to explore in future

- work in new branch, commit regularly and keep a plan.md with what you're doing and what you've done

i know this is a significant ask. write down a plan how to tackle this in multiple steps and ask me as much questions upfront so i can give this to a cloud agent afterward to autonomously implement. if i've overlooked something based on your perspective, let me know
```

</details>

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -50,7 +50,7 @@ In either case, you can focus individual tests during development via `it.only`.
 
 This project uses ESLint for detecting issues in code, as well as Prettier for formatting.
 
-It's recommended to use [an editor integration](https://github.com/molindo/eslint-config-molindo?tab=readme-ov-file#editor-integration) for both, so that autofixable issues are automatically resolved. For editor plugins to work correctly, you should open individual packages you work on as the workspace root (e.g. `next-intl/packages/next-intl` instead of `next-intl`).
+It's recommended to use an editor integration for both, so that autofixable issues are automatically resolved. For editor plugins to work correctly, you should open individual packages you work on as the workspace root (e.g. `next-intl/packages/next-intl` instead of `next-intl`).
 
 Alternatively, you can run ESLint and Prettier via the command line:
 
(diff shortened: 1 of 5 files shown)


amannn:
@cursoragent review changes in this pr

PR #2245, opened 2026-02-02
fix: Avoid throwing config errors for non-Next.js consumers of `next.config.ts`

amannn:
Fixes https://github.com/amannn/next-intl/discussions/2209

Nx’s `@nx/next/plugin` builds the project graph by importing `next.config.ts` directly, outside of a real `next dev`/`next build` invocation. Since `next-intl/plugin` now configures Turbopack more broadly (`shouldConfigureTurbo` via `isNextJs16OrHigher`), the plugin runs early during this analysis step and `resolveI18nPath()` validates `i18n/request` relative to the current CWD (often the monorepo root). In Nx monorepos this can’t be resolved and breaks the graph/build with “Could not locate request configuration module”, even though the app works when Next.js actually runs.

This change makes the plugin only throw config-path errors when Next.js is actually running (`next dev`/`next build`). For other consumers that merely import the config (Nx graph, tooling), it avoids throwing while still returning a stable alias target.

### DIFF
--- a/packages/next-intl/src/plugin/config.tsx
+++ b/packages/next-intl/src/plugin/config.tsx
@@ -1,0 +1,9 @@
+// Avoid rollup's `replace` plugin to compile this away
+const nodeEnvKey = 'NODE_ENV'.trim();
+
+// We avoid reading `argv.includes('dev')` related to
+// https://github.com/amannn/next-intl/issues/2006
+export const isDevelopment = process.env[nodeEnvKey] === 'development';
+
+export const isNextBuild = process.argv.includes('build');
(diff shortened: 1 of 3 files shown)


amannn:
@creage I've fixed the issue you raised at https://github.com/amannn/next-intl/discussions/2209#discussioncomment-15650927 here. In my reproduction it works fine now.

Can you give this pre-release a quick try in your app to confirm?

```
next-intl@0.0.0-canary-6b9a772
```

creage:
@amannn works fine - no more errors during the build. Thank you!

PR #2252, opened 2026-02-10
feat: Automatic tree-shaking of messages (experimental)

### LINKED ISSUE #1: Automatic tree-shaking of messages

amannn:
Resolves #1 (but very, very WIP to experiment for now)

**TODO**
- [ ] Implementation
  - [ ] Test with https://github.com/vercel/next.js/pull/96631 (not documented yet)
  - [ ] Compiler orchestration
    - [ ] Manifest loader only has a single compile pass. Can we do something similar to integrate with extraction loader?
    - [ ] Offload more to SWC / Rust? Maybe a single SWC plugin that returns all analysis (`useExtracted`, deps, `'use client'`, …)
    - [ ] Can we emit extractor and tree shaking result at same time to avoid redundant work?
  - [ ] API
    - [ ] Should we add a lower-level API? The usage is _slightly_ misleading, because not everything below the provider is collected, but everything that is imported into the module. E.g. `import manifest from 'next-intl/client-manifest?path=optional' with {type: 'macro'}` would be another candidate, but not sure if the verbosity is worth it (or if it's even clearer). It's surely more explicit, but `useExtracted` is already implicitly subject to AST transforms—so not sure if we need to do it here.
    - [ ] Pages Router?
    - [ ] Any way to easily cut off server messages without more work?
  - [ ] Edge cases
    - [ ] Solution for `error.tsx`
    - [ ] https://github.com/amannn/next-intl/pull/2262
    - [ ] Check Next.js docs thoroughly for other use cases to consider
    - [ ] Configurable tsconfig? ([ref](https://nextjs.org/docs/app/api-reference/config/typescript)) — forward this
    - [ ] Test what happens where there are syntax errors.
  - [ ] Rework implementation / deslop
  - [ ] Performance
- [ ] Tests
  - [ ] Can we reuse this for an integration test?
- [ ] First user testing
- [ ] Docs
  - [ ] A playground for toggling layout boundaries?


LuHaikun:
When can this new feature be launched?

amannn:
@LuHaikun I've delayed this feature for a bit, related to https://github.com/vercel/next.js/pull/96631. This could potentially be a built-in API in Turbopack that allows to implement this feature more cleanly and integrated. It's undocumented and not API-stable for now though.

Issue #2047, opened 2025-10-07
☂️ Umbrella: List of potential changes for a major version


amannn:
This is a list of ideas for API changes that we could potentially introduce in a major version:

- https://github.com/amannn/next-intl/issues/2049 (ideally this is non-breaking, but to be safe we could schedule it for a major version)
- Investigate if we could remove `t.rich` and `t.markup` in favor of a universal `t` function that works for all cases.
- Investigate if we could get rid of `t.raw`. It leads users to toward bad patterns that might not work with future optimizations (see https://github.com/amannn/next-intl/discussions/2076#discussioncomment-14867566 for details).
- Fix casing `getTimeZone` (`next-intl`) vs `Timezone` (`use-intl`)
- Consider making `NextIntlClientProvider` server-only (with `IntlProvider` being available on the client). The environment makes a difference here and it might be to easy to get wrong. In this case, we'd also move it to `next-intl/server`. Not sure if the benefit is big enough though.
- Rename "middleware" in APIs to "proxy"
- Should `Link` that changes a `locale` include `nofollow` by default? This could help search engines avoid indexing this link while allowing us to use a [safe href before hydration](https://next-intl.dev/docs/routing/navigation#link-locale). Search engines should pick up localized variants from alternate links or a sitemap.
- Remove "node10" types support
- Only a loose thought for now, but I wonder if it could be helpful to return routing config from `request.ts` instead of passing it to a factory function like `createNavigation`. Really not sure yet if it's worth it, but could be something to investigate.
  - **Pro's**: Avoids a [hook factory](https://react.dev/reference/eslint-plugin-react-hooks/lints/component-hook-factories), for apps with [mixed routing](https://next-intl.dev/examples#app-router-mixed-routing) users could use `import {Link} from 'next-intl/navigation'` for both parts of the app, avoids https://github.com/amannn/next-intl/issues/1271
  - **Con's**: Type safety of routes would be opt-in (similar to `formats`), this is especially annoying for routes with dynamic params
- Should we use UTC as a default for the time zone instead of the one the server provides? A benefit might be that you get the same results locally as when deploying to a production server.
- Should we make the `now` argument mandatory for `format.relativeTime(…)` to be explicit and avoid errors?
- Once `rootParams` is a thing, we could consider removing the `{locale}` argument from `getNow`. Other async functions are TBD, we might want to keep it for some like `getTranslations` and `getFormatter`.
[...]

PR #2281, opened 2026-03-09
fix: Set redirect domain if `x-forwarded-host` header exists

FourwingsY:
## Summary

Fix redirect URL resolution behind reverse proxies when `domains` is **not** configured in `defineRouting`.

This ensures middleware redirects can use forwarded request metadata instead of falling back to the deployment/internal host from `request.url`.

Related context: Discussion [#2206](https://github.com/amannn/next-intl/discussions/2206).

## Problem

In reverse-proxy setups (e.g. CloudFront -> Vercel), redirects could resolve to the backend/deployment host unless domain-based routing was enabled.  
Even when `x-forwarded-host` was present, redirect origin handling was effectively tied to the `redirectDomain` path.

## Changes

- Updated `redirect()` in middleware to make origin selection explicit:
  1. `redirectDomain` (domain-routing decision)
  2. `x-forwarded-host`
  3. fallback to `request.url`
- Added a shared redirect response helper to reduce duplication and keep flow clearer.
- Added test coverage for proxy behavior without `domains`.

## Tests

Added a new test case under prefix-based routing (`localePrefix: 'always'`):

- Redirect uses `x-forwarded-host` + `x-forwarded-proto`
- Redirect uses `x-forwarded-host` + `x-forwarded-port` + `x-forwarded-proto`

Validation run:

- `pnpm --filter next-intl exec vitest run src/middleware/middleware.test.tsx`
- Result: `221 passed`

## Impact

- Fixes proxy-host redirect behavior for non-domain routing setups.
- Preserves existing domain-routing behavior and expectations.

---

⚠️ If you're using a setup behind a reverse proxy and your proxy sets `x-forwarded-port`, make sure the value is correct (typically 443).


amannn:
Hey, thanks for opening this PR!

I had a pretty busy week, I just wanted to let you know that I still have this on my list to look into. I'll get back to you!

amannn (approved):
Hey, thanks a lot for looking into this!

Sorry that it took a bit, I was pretty [busy](https://x.com/jamannnnnn/status/2046172101904605555) lately 😅

I added a commit that moves the code around a bit, but it keeps the essence of your fix.

ckarli:
My prod redirects to :3000 after this. What do i need to accomodate this change?

amannn:
@ckarli really? Can you share details about your setup?

ckarli:
What info do you want me to supply?

amannn:
Where/how you host your app and relevant config

ckarli:
Caddy for ssl, nginx on the container. No access to caddy unfortunately. 

proxy.ts

```
import createMiddleware from "next-intl/middleware";
import { NextRequest, NextResponse } from "next/server";
import { locales, routing } from "./i18n/routing";
import { validateSession } from "./lib/iron-session";

const publicPages = ["auth/"]; // your public routes
	`^(/(${locales.join("|")}))?\/(${publicPages
		.flatMap((p) => (p === "/" ? ["", "/"] : p))
		.join("|")})/?`,
	"i",
);

const handleI18nRouting = createMiddleware(routing)

function isPublicPage(pathname: string) {
	return publicPathnameRegex.test(pathname);
}

export default async function proxy(req: NextRequest) {
	const pathname = req.nextUrl.pathname;

	// Check if it's a public page
	if (isPublicPage(pathname)) {
		const response = handleI18nRouting(req);
		return response;
	}

	// Check for authentication on protected routes
	const session = await validateSession();
	if (!session.isLoggedIn) {
		// No session cookie, redirect to signin
		const signInUrl = new URL("/auth/signin", req.url);
		signInUrl.searchParams.set("callbackUrl", req.nextUrl.pathname);
		return NextResponse.redirect(signInUrl);
	}

	// User is authenticated, continue with i18n routing
	const response = handleI18nRouting(req);
	return response;
}

export const config = {
	matcher: [
		"/((?!api|_next|_vercel\\\\.|health|icons\\/|images\\/|profile\\/|service-worker\\.js|robots\\.txt|manifest\\.webmanifest|favicon\\.ico).*)",
	],
};
```

nginx.conf

```
upstream nextjs_upstream {
  server frontend:${FRONTEND_PORT};
}

server {
    access_log /var/log/nginx/access.log main;
    server_name ${FRONTEND_HOST};
        index index.html index.htm;
        root /frontend/; #Make sure your using the full path

        # Serve any static assets with NGINX
        location /.next/static {
            alias /frontend//_next/static;
            add_header Cache-Control "public, max-age=3600, immutable";
        }

        location / {
            try_files $uri.html $uri/index.html # only serve html files from this dir
            @public
            @nextjs;
            add_header Cache-Control "public, max-age=3600";
        }

        location @public {
            add_header Cache-Control "public, max-age=3600";
        }

        location @nextjs {
            # reverse proxy for next server
            proxy_pass http://nextjs_upstream; #Don't forget to update your port number
            proxy_http_version 1.1;
            proxy_set_header Upgrade $http_upgrade;
            proxy_set_header Connection 'upgrade';
            proxy_set_header Host $host;
            proxy_set_header X-Forwarded-Host $host;
            proxy_set_header X-Forwarded-Proto ${PROTOCOL};
            proxy_cache_bypass $http_upgrade;
            proxy_buffer_size 128k;
            proxy_buffers 8 128k;
            proxy_busy_buffers_size 256k;

        }

        listen 80 default_server;
        listen [::]:80;
}

```

amannn:
Thanks!

Can you add logging in your middleware which headers your Next.js instance receives?

These are relevant:

- 'x-forwarded-host'
- 'x-forwarded-port'
- 'x-forwarded-proto'

proto should be 80 in your case.

Mr-Vipi:
> My prod redirects to :3000 after this. What do i need to accomodate this change?

@amannn  I have the same issue

- x-forwarded-host: thevansurfers.com
- request.details.headers.x-forwarded-port: 3000
- request.details.headers.x-forwarded-proto: https

proxy.ts
```import type { NextFetchEvent, NextRequest } from "next/server"
import { Logger } from "next-axiom"
import createMiddleware from "next-intl/middleware"

import { routing } from "./i18n/routing"

export async function proxy(request: NextRequest, event: NextFetchEvent) {
  const logger = new Logger({ source: "middleware" })
  logger.middleware(request, { logRequestDetails: true })
  event.waitUntil(logger.flush())
  // Call next-intl middleware for i18n handling
  return createMiddleware(routing)(request)
}

export const config = {
  // Match all pathnames except for
  // - … if they start with `/api`, `/trpc`, `/_next` or `/_vercel`
  // - … the ones containing a dot (e.g. `favicon.ico`)
  matcher: "/((?!api|trpc|_next|_vercel|_axiom|.*\\..*).*)",
}
```
routing.ts
```
import { defineRouting } from "next-intl/routing"

  // A list of all locales that are supported
  locales: ["en", "de", "sq", "it"],

  // Used when no locale matches
  defaultLocale: "en",
})

```

amannn:
> * request.details.headers.x-forwarded-port: 3000

That looks off, this header should not be set if you don’t want to add the port. x-forwarded-port represents the port the client has originally requested.

Sorry that this broke for you, but this looks like a bug in your proxy. next-intl now correctly honors this header.

ckarli:
Fixed it by setting 
proxy_set_header X-Forwarded-Port 443;

It was 3000 in the logs.

Mr-Vipi:
> Fixed it by setting proxy_set_header X-Forwarded-Port 443;
> 
> It was 3000 in the logs.

I fixed it too in the same way

thilak-rao:
@amannn Posting another data point. DigitalOcean App Platform

Same pattern as @ckarli and @Mr-Vipi  after the next-intl@4.9.1 → 4.11.0 bump. Prefix routing, no redirectDomain.

https://example.app started redirecting to https://example.app:3003/en, where 3003 is the container http_port we configure in our DO app spec. This took our marketing sites down until I rolled back.

amannn:
My apologies, this change clearly had a bigger impact than I thought. It seems like we tolerated some inconsistency around request headers previously, but at the same time this caused a bug for other users (see original report).

If you're already configuring `x-forwarded-port` on your proxy, you should make sure it's valid (typically `443`).

Jaynarol:
Same issue as @thilak-rao after updating to next-intl 4.11.0 on DigitalOcean App Platform.

Showfom:
Same issue here, I have asked Claude Code for a quick fix on `proxy.ts`

```
const handleI18nRouting = createMiddleware(routing);

// Next.js 16.2.5+ builds middleware redirect URLs from req.nextUrl, whose origin
// reflects the internal listener (e.g. 0.0.0.0:3000) rather than the public host
// the reverse proxy serves. Force the Location header back onto a trusted origin
// derived from request headers so cookie-based locale redirects don't leak ":3000".
export default function proxy(request: NextRequest) {
  const response = handleI18nRouting(request);
  const location = response.headers.get('location');
  if (!location) return response;

  const forwardedHost = request.headers.get('x-forwarded-host');
  const host = forwardedHost ?? request.headers.get('host');
  if (!host) return response;
  const proto =
    request.headers.get('x-forwarded-proto')?.split(',')[0].trim() ||
    request.nextUrl.protocol.replace(':', '');

  try {
    const target = new URL(location, `${proto}://${host}`);
    const [hostname, port = ''] = host.split(':');
    target.protocol = `${proto}:`;
    target.hostname = hostname;
    target.port = port;
    response.headers.set('location', target.toString());
  } catch {
    // Leave non-URL Location values untouched.
  }
  return response;
}
```

EveraertAlexander:
Same issue on Azure App Service (Linux, standalone Next.js 16.2.6, next-intl 4.11.1). Azure's managed reverse proxy forwards HTTPS (443) → internal port 8080, and sets x-forwarded-port: 8080 instead of 443. 

We have no control over this header. Root / redirect goes to https://myapp.azurewebsites.net:8080/en-US. Workaround: rewriting the Location header in our proxy function using only x-forwarded-host + x-forwarded-proto (ignoring x-forwarded-port).

Showfom:
@amannn please consider adding something to next-intl, or people will be confused.

amannn:
Yes, it seems like based on the amount of people who chimed in, merging this PR was shortsighted—my apologies.

As a first step, this will be reverted as of https://github.com/amannn/next-intl/pull/2322 to avoid that more projects run into this.

I'm lacking a bit of context here currently on why there are so many (seemingly?) misconfigured proxies around, e.g. Azure's managed reverse proxy setting the header to an internal port. @FourwingsY would you be interested on investigating this? I think we need more context for introducing such a change, maybe for a future major version.

Thanks everyone who has left a note here!

goetzc:
I'm experiencing the same issue with AWS App Runner where the redirect is being sent to port 3000. In this particular setup, there's not much configuration to do.

Don't believe there's any misconfiguration on AWS or Azure build-in reverse proxies.

amannn:
@goetzc if you upgrade to [4.11.2](https://github.com/amannn/next-intl/releases/tag/v4.11.2) then this should be resolved for you.

EveraertAlexander:
> @goetzc if you upgrade to [4.11.2](https://github.com/amannn/next-intl/releases/tag/v4.11.2) then this should be resolved for you.

Just tested this out with Azure App Service, can confirm this resolved my issue. Thanks for the quick response

FourwingsY:
I've read through the thread. @amannn 

Given that trusting `x-forwarded-*` headers can be deployment-sensitive, I think this should probably be an explicit opt-in rather than default behavior.

One possible approach could be adding a option like `respectForwardedHeaders`, so apps that are behind a trusted reverse proxy can opt into using `x-forwarded-host`, `x-forwarded-proto` and `x-forwarded-port` for redirect URL construction.

This would keep the current default behavior working as-is, while giving reverse-proxy setups a supported way to avoid redirects to the internal/deployment host.

amannn:
> adding a option like `respectForwardedHeaders`



Sorry, but I really don't want to include such an option. What we should do is have a default that works for the absolute majority of setups, and where not, there can be programmatic configuration with [middleware composition](https://next-intl.dev/docs/routing/middleware#composing-other-middlewares) (some people have shared their approaches in this thread).

What I'd be open to is to research why this has affected so many clients negatively, and to re-consider if we should potentially adjust something here for an upcoming major version (https://github.com/amannn/next-intl/issues/2047). I'm lacking quite some context here to make an informed decision, so this research is probably the hardest part of this.

PR #2287, opened 2026-03-17
feat: stack all descriptions from multiple files in PO extraction

### LINKED ISSUE #2286: Non-deterministic description extraction for identical messages in unstable_extractMessages
### Description

I've run into an issue regarding message extraction (specifically generating .po files via unstable_extractMessages). When the same default message is used in multiple places but with different descriptions, the extraction is non-deterministic. Only one description is kept, and the one that "wins" changes depending on the execution method.

In our codebase, we have the same string used in two different files, but with different descriptions to give translators proper context:

```tsx
t({
  message: 'Message1',
  description: 'Description A'
})
```

```tsx
t({
  message: 'Message1',
  description: 'Description B'
})
```

#### Current Behavior

- Only one message entry is created in [file].po.
- The extracted description is non-deterministic. Depending on whether the extraction is triggered via the standard dev server (yarn dev) or by running the explicit unstable_extractMessages script, the .po file flips between the two descriptions.
- This creates unnecessary git diffs and churn in our translation files, as the file changes depending on how the build/extraction was executed.

I'm curious what the intended design is for this edge case. I'd love to get your thoughts on a couple of potential approaches: Should we allow unique entries per message + description pair? Should the extraction order be deterministic? How do you recommend handling this? Happy to help open a PR if we agree on a direction!

### Verifications

https://github.com/baylesa-dev/next-intl-repro

### Reproduction description

Steps to reproduce:
1. Open reproduction
2. Run npm run dev
3. Commit changes
4. Run npm run extract
5. messages/en.po got changes

### Expected behaviour

Let's discuss it :)

baylesa-dev:
### Description

This PR fixes an issue where extracting identical messages with different descriptions into a .po file resulted in non-deterministic behavior. Previously, the extractor only kept one description (first-wins), which led to missing translator context and unnecessary git diffs depending on the extraction order.

### Changes in this PR:

- All unique descriptions for a given message are now accumulated.
- They are output as separate #.  lines above the translation block.
- The descriptions are sorted alphabetically to guarantee a deterministic .po file output every time.

###### Before:

```po
#. First description
msgid ""
msgstr ""
```

###### After:

```po
#. First description
#. Second description
msgid ""
msgstr ""
```

### Related Issue
Closes #2286

### DIFF
--- a/packages/next-intl/src/extractor/ExtractionCompiler.test.tsx
+++ b/packages/next-intl/src/extractor/ExtractionCompiler.test.tsx
@@ -435,6 +435,502 @@ describe('po format', () => {
     expect(relativeSpy).toHaveBeenCalled();
   });
 
+  it('tracks all line numbers when same message appears multiple times in one file', async () => {
+    filesystem.project.src['Greeting.tsx'] =
+      `import {useExtracted} from 'next-intl';
+    function Greeting() {
+      const t = useExtracted();
+      return (
+        <div>
+          {t('Hello!')}
+          {t('Hey!')}
+          <span>{t('Hello!')}</span>
+        </div>
+      );
+    }
+    `;
+    filesystem.project.src['Greeting2.tsx'] =
+      `import {useExtracted} from 'next-intl';
+    function Greeting2() {
+      const t = useExtracted();
+      return t('Hello!');
+    }
+    `;
+    filesystem.project.messages = {};
+
+    using compiler = createCompiler();
+    await compiler.extractAll();
+    await waitForWriteFileCalls(1);
+    expect(vi.mocked(fs.writeFile).mock.calls[0][1]).toMatchInlineSnapshot(`
+      "msgid ""
+      msgstr ""
+      "Language: en\\n"
+      "Content-Type: text/plain; charset=utf-8\\n"
+      "Content-Transfer-Encoding: 8bit\\n"
+      "X-Generator: next-intl\\n"
+      "X-Crowdin-SourceKey: msgstr\\n"
+
+      #: src/Greeting.tsx:6
+      #: src/Greeting.tsx:8
+      #: src/Greeting2.tsx:4
+      msgid "OpKKos"
+      msgstr "Hello!"
+
+      #: src/Greeting.tsx:7
+      msgid "+YJVTi"
+      msgstr "Hey!"
+      "
+    `);
+  });
+
+  it('saves messages initially', async () => {
+    filesystem.project.src['Greeting.tsx'] =
+      `import {useExtracted} from 'next-intl';
+    function Greeting() {
+      const t = useExtracted();
+      return <div>{t('Hey!')}</div>;
+    }
+    `;
+    filesystem.project.messages = {
+      'en.po': `
+      #: src/Greeting.tsx:4
+      msgid "+YJVTi"
+      msgstr "Hey!"
+      `,
+      'de.po': `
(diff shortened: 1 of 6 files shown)


amannn:
Hey, thanks for proposing this change!

Can you check how other PO extractors like gettext and lingui handle this case? Also it'd be interesting if there are discussions around this topic. I'm wondering if there's a standard behavior here …

baylesa-dev:
Hey @amannn,

Both gettext and lingui merge identical messages and stack descriptions and file sources together :

```c
/* Description A */
gettext("Message");

/* Description B */
gettext("Message");
```

```bash
$ xgettext --add-comments test.c -o output.po && cat output.po


# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
# ...
#. Description A
#. Description B
#: test.c:2 test.c:5
msgid "Message"
msgstr ""
```


There's a difference compared to this implementation : gettext sorts the descriptions in chronological order of discovery. I sorted them for determinism, since useExtracted may encounter files in different orders depending on the bundler.


Note: they also provide a context ([pgettext in gettext](https://www.gnu.org/software/gettext/manual/html_node/Contexts.html) and [context in lingui/i18next](https://lingui.dev/misc/i18next#context)), to distinguish identical strings with different meanings. 


```c
/* Description A */
pgettext("Context A", "Message");

/* Description B */
pgettext("Context B", "Message");
```

```bash
$ xgettext --add-comments test.c -o output.po && cat output.po
 
# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
# ...
#. Description A
#: test.c:2
msgctxt "Context A"
msgid "Message"
msgstr ""

#. Description B
#: test.c:5
msgctxt "Context B"
msgid "Message"
msgstr ""
```



Let me know what you want to do :)

baylesa-dev:
May be related to https://github.com/amannn/next-intl/pull/2292#issuecomment-4372196015 💡

amannn:
Hey, I just had a moment to have a closer look. I agree with your reasoning that descriptions should be stacked, thanks for checking out what related tools are doing here!

However there's a problem with the implementation that requires a slightly bigger refactor, reaching into the Rust code.

I've added https://github.com/amannn/next-intl/pull/2318 to achieve the same functionality, let's continue there! Would love if you could provide your input there if this matches what you're looking for!

---

**From a chat with GPT 5.5 on this PR:**

Partially correct, but I would not merge as-is.

The PR does stack descriptions on initial extraction and PO encode/decode now supports multiple `#. ...` comments. The focused extractor test file passes:

```sh
pnpm --filter next-intl exec vitest run src/extractor/ExtractionCompiler.test.tsx
```

The main issue is that descriptions are merged into the global message object and lose provenance. If one file contributes description A and another contributes description B, then the first file is edited or deleted, description A remains forever. Cleanup only removes references, not descriptions.

Similarly, if a source occurrence removes its description, the generic metadata merge copies the previous `description` back. Descriptions become sticky.

Best approach: keep per-file extracted messages unmerged in `messagesByFile`, then derive or recompute the aggregate `messagesById` for affected IDs from all current file entries. Merge unique descriptions during that aggregation, same as references. This preserves deterministic stacked output without stale descriptions.

Secondary concern: widening `ExtractorMessage.description` to `string | Array<string>` makes this a cross-codec/public shape change. That might be acceptable, but custom codecs should be expected to handle arrays intentionally.

Overall: the reasoning is sound and the happy path works, but the state model is wrong for incremental updates.

PR #2292, opened 2026-03-24
feat: add `devExtraction` option to skip catalog writes during dev

sezze:
Add `devExtraction` option to the `extract` config. When set to `false`, the `ExtractionCompiler` (which scans source files and writes message catalogs) is skipped during `next dev`, while the SWC extraction loader (which compiles `useExtracted` into `useTranslations`) continues to run.

This enables a workflow where:
1. Developers use `useExtracted` in code as normal
2. `next dev` does not write PO/JSON files, keeping the working tree clean
3. Message extraction runs via `unstable_extractMessages` in CI or a separate script (e.g. a GitHub Action)
4. `next build` still runs extraction as before

Usage:
```ts
createNextIntlPlugin({
  experimental: {
    srcPath: './src',
    extract: {
      sourceLocale: 'en',
      devExtraction: false
    },
    messages: { path: './messages', format: 'po', locales: 'infer' }
  }
});
```

Defaults to `true` for full backwards compatibility.

### DIFF
--- a/packages/next-intl/src/plugin/extractor/initExtractionCompiler.tsx
+++ b/packages/next-intl/src/plugin/extractor/initExtractionCompiler.tsx
@@ -25,9 +25,10 @@ export default function initExtractionCompiler(pluginConfig: PluginConfig) {
   // - lint
   //
   // What remains are:
-  // - dev (NODE_ENV=development)
+  // - dev (NODE_ENV=development), skipped when devExtraction is false
   // - build (NODE_ENV=production)
-  const shouldRun = isDevelopment || isNextBuild;
+  const devExtraction = experimental.extract?.devExtraction !== false;
+  const shouldRun = (isDevelopment && devExtraction) || isNextBuild;
   if (!shouldRun) return;
 
   runOnce(() => {
--- a/packages/next-intl/src/plugin/types.tsx
+++ b/packages/next-intl/src/plugin/types.tsx
@@ -27,6 +27,12 @@ export type PluginConfig = {
     /** Enables the usage of `useExtracted`, to be used in combination with `srcPath` and `messages`. */
     extract?: {
       sourceLocale: string;
+      /**
+       * Whether to automatically extract messages during `next dev`.
+       * The SWC transform for `useExtracted` runs regardless.
+       * @default true
+       */
+      devExtraction?: boolean;
     };
   };
 };


amannn:
Hey @sezze, thanks for opening this PR!

I've just replied to your question here: https://github.com/amannn/next-intl/discussions/2036#discussioncomment-16753577. 

I think the current feature set might already support this for .json, and for .po I'm planning to introduce another option that would help to support this workflow. Due to this, I'd like to not introduce a separate config flag for this currently.

Hope that makes sense! Which file format are you using?

Somewhat off-topic: I saw in your bio that you work for x.ai, are you using `next-intl` somewhere there? 👀

sezze:
> Hey @sezze, thanks for opening this PR!
> 
> I've just replied to your question here: [#2036 (reply in thread)](https://github.com/amannn/next-intl/discussions/2036#discussioncomment-16753577).
> 
> I think the current feature set might already support this for .json, and for .po I'm planning to introduce another option that would help to support this workflow. Due to this, I'd like to not introduce a separate config flag for this currently.
> 
> Hope that makes sense! Which file format are you using?

Hi, I'm using .po files.

When I first introduced next-intl, I received complaints of unexpected diffs touching .po files, when no changes were made to localization (simply line numbers changing, as the PO files include this). And that diffs were too large (which is unrelated to next-intl). My current idea is to decouple logic changes and translations, by using stacked PRs, where a separate PR in the stack will contain translation changes, so it's easier to review. This then requires extraction to not happen in real time, as it would be in a separate PR.

> Somewhat off-topic: I saw in your bio that you work for x.ai, are you using `next-intl` somewhere there? 👀

Yes, some things are using `next-intl` 🙂

amannn:
> simply line numbers changing, as the PO files include this

Yep, other developers also reported this feedback (https://github.com/amannn/next-intl/discussions/2036#discussioncomment-15770359, https://github.com/amannn/next-intl/discussions/2036#discussioncomment-15930728) — I might include an option to remove line numbers.

Would you prefer to have it turned off?

> This then requires extraction to not happen in real time, as it would be in a separate PR.

Do you think you'd still use this if there were no line numbers in the diff and diffs less common as a result of that?

But apart from that, I think this could be supported with the existing API with a small tweak that kind of makes sense anyway for consistency: https://github.com/amannn/next-intl/pull/2317. Would this PR work for you too?


> Yes, some things are using `next-intl` 🙂

That's amazing! Is it internal tooling or something public? In case the library works well for you, do you think there's a chance I could include a mention about x.ai using next-intl on the docs website?

amannn:
This has shipped as `locales: []` in [4.12](https://github.com/amannn/next-intl/releases/tag/v4.12.0)

PR #2295, opened 2026-03-28
docs: add example for translated error messages in non-React utilities

Andrew1326:
## Summary
- Adds a practical example to the "Non-React apps" section showing how to use `createTranslator` in API interceptors and utility functions

## Context
Users commonly ask how to use translations outside of React components — for example in Axios interceptors that need to show translated error messages ([Discussion #2246](https://github.com/amannn/next-intl/discussions/2246)).

The existing docs show the basic `createTranslator` API, but don't include a real-world example of this common pattern. This small addition bridges that gap.

### DIFF
--- a/docs/src/pages/docs/environments/core-library.mdx
+++ b/docs/src/pages/docs/environments/core-library.mdx
@@ -105,3 +105,21 @@ const format = createFormatter({locale: 'en'});
 // Result: "Oct 17, 2022"
 format.dateTime(new Date(2022, 9, 17), {dateStyle: 'medium'});
 ```
+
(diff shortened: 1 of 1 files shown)


amannn:
Hey, thanks for leaving your comment at https://github.com/amannn/next-intl/discussions/2246 with your approach.

Personally, I don't endorse this approach however—I've covered this in https://next-intl.dev/blog/translations-outside-of-react-components. Also newer features like `useExtracted` don't work with this pattern.

Opinions differ of course, I think the GitHub discussions are a good place for people to find user-suggested patterns.

PR #2315, opened 2026-05-04
fix: Make extracted catalog order deterministic on tied references

anthlasserre:
Follow-up to the discussion in #2036.

## Problem

`getSortedMessages` (in `extractor/utils.tsx`) sorts catalog entries by `references[0]` only and returns `0` on ties (same path + line, or missing references). Since `Array.prototype.toSorted` is stable, ties fall back to `messagesById` insertion order — which is non-deterministic because source files are scanned via `fs.readdir` and processed concurrently.

In practice, two messages sharing the same first source location (e.g. two `t()` calls on the same JSX line, or messages in tightly-bound JSX expressions) swap positions across runs and across machines. For repos that re-extract on every commit through a pre-commit hook, this produces reorder-only diffs that pollute unrelated PRs.

This is the same family of issue as the one reported by @ceolinwill earlier in #2036 and partially fixed in 4.5.6 ("Stable, environment-agnostic key sorting"). That fix made the comparator locale-independent, but didn't add a tiebreaker — so messages tied on `references[0]` still flip.

@LouisCuvelier raised the same symptom in [#2036 (comment)](https://github.com/amannn/next-intl/discussions/2036#discussioncomment-16766583), asking whether `unstable_extractMessages` and `next dev` / `next build` produce the same output. They share this same comparator, so once it's deterministic, all paths converge.

## Fix

Add `localeCompare(messageA.id, messageB.id)` as the final tiebreaker in `getSortedMessages`. The `id` is already deterministic (it's the existing extraction hash), so this gives a fully stable order. ~3-line change.

```ts
// extractor/utils.tsx
if (refA && refB) {
  const refCompare = compareReferences(refA, refB);
  if (refCompare !== 0) return refCompare;
}
return localeCompare(messageA.id, messageB.id);
```

## Tests

- **Unit**: renamed `preserves original order when reference paths and lines match` → `uses id as tiebreaker when reference paths and lines match` (now asserts `['a', 'b', 'c']` instead of insertion order). Added `uses id as tiebreaker when references are missing`.
- **E2E** (`extracted-po`): new test `orders messages deterministically when they share a first reference` — places two `t()` calls on the same JSX line and asserts the catalog entries appear in id-sorted order.
- **Snapshot**: one inline snapshot in `ExtractionCompiler.test.tsx > po format > removes flags when externally deleted` updated. The test deliberately produces 5 messages all at `Greeting.tsx:5` (a tied-references case), so the order shifts under the new tiebreaker. Pure reorder, no content change.

## Caveat

The new e2e test verifies the fix's *outcome* (catalog order matches id-sort), not its *causality* — since the msgid is a hash of the source string, there's a chance the hash order coincides with insertion order, in which case the test would also pass without the fix. The unit tests cover the regression directly with controlled input.

## Scope

Intentionally minimal: only adds a tiebreaker. Does not change behavior for messages with mixed reference states (one has refs, the other doesn't) — those now sort by id alongside everything else, which is also deterministic. Happy to refine if you'd prefer a different rule there.

### DIFF
--- a/packages/next-intl/src/extractor/utils.test.tsx
+++ b/packages/next-intl/src/extractor/utils.test.tsx
@@ -41,7 +41,7 @@ describe('getSortedMessages', () => {
     ).toEqual(['a', 'b', 'c']);
   });
 
-  it('preserves original order when reference paths and lines match', () => {
+  it('uses id as tiebreaker when reference paths and lines match', () => {
     expect(
       getSortedMessages([
         {
@@ -60,7 +60,17 @@ describe('getSortedMessages', () => {
           references: [{path: 'components/A.tsx', line: 10}]
         }
       ]).map((message) => message.id)
-    ).toEqual(['c', 'a', 'b']);
+    ).toEqual(['a', 'b', 'c']);
+  });
+
+  it('uses id as tiebreaker when references are missing', () => {
+    expect(
+      getSortedMessages([
+        {id: 'c', message: 'c'},
+        {id: 'a', message: 'a'},
+        {id: 'b', message: 'b'}
+      ]).map((message) => message.id)
+    ).toEqual(['a', 'b', 'c']);
   });
 });
 
--- a/e2e/extracted-po/tests/main.spec.ts
+++ b/e2e/extracted-po/tests/main.spec.ts
@@ -78,6 +78,37 @@ export default function Greeting() {
   expect(greetingRefs.length).toBeGreaterThanOrEqual(2);
 });
 
+it('orders tied messages by id', async ({page}) => {
+  // Two distinct messages on the same line tie on `references[0]`.
+  // 'apple' hashes to "hE2HeR" and 'banana' to "+OMYPT", so localeCompare
+  // sorts banana before apple — opposite to source order. Without the id
+  // tiebreaker the catalog preserves source order (apple first); with it,
+  // banana wins.
+  await using _ = await withTempEditApp(
+    'src/components/Greeting.tsx',
+    `'use client';
+
+import {useExtracted} from 'next-intl';
+
+export default function Greeting() {
+  const t = useExtracted();
+  return <div>{t('apple')}{t('banana')}</div>;
+}
+`
+  );
+
+  await page.goto('/');
+  const content = await expectCatalog(
+    'en.po',
+    (c) => c.includes('msgid "+OMYPT"') && c.includes('msgid "hE2HeR"'),
+    {timeout: 15_000}
+  );
+
+  const bananaPos = content.indexOf('msgid "+OMYPT"');
+  const applePos = content.indexOf('msgid "hE2HeR"');
+  expect(bananaPos).toBeLessThan(applePos);
+});
+
 it("saves catalog when it's missing", async ({page}) => {
   await page.goto('/');
   await expectCatalog(
--- a/packages/next-intl/src/extractor/ExtractionCompiler.test.tsx
+++ b/packages/next-intl/src/extractor/ExtractionCompiler.test.tsx
@@ -1165,6 +1165,10 @@ msgstr ""
       msgstr "Hallo!"
 
       #: src/Greeting.tsx:5
+      msgid "I5NMJ8"
+      msgstr ""
+
+      #: src/Greeting.tsx:5
       msgid "jqdzk6"
       msgstr ""
 
@@ -1174,10 +1178,6 @@ msgstr ""
 
       #: src/Greeting.tsx:5
       msgid "pE58D7"
-      msgstr ""
-
-      #: src/Greeting.tsx:5
-      msgid "I5NMJ8"
       msgstr ""
       ",
       ]
--- a/packages/next-intl/src/extractor/utils.tsx
+++ b/packages/next-intl/src/extractor/utils.tsx
@@ -55,11 +55,16 @@ export function getSortedMessages(
     const refA = messageA.references?.[0];
     const refB = messageB.references?.[0];
 
-    // No references: preserve original (extraction) order
-    if (!refA || !refB) return 0;
+    if (refA && refB) {
+      const refCompare = compareReferences(refA, refB);
+      if (refCompare !== 0) return refCompare;
+    }
 
-    // Sort by path, then line. Same path+line: preserve original order
-    return compareReferences(refA, refB);
+    // Tiebreaker for messages tied on (or missing) `references[0]`. Without
+    // this, the sort falls back to insertion order in `messagesById`, which
+    // is non-deterministic because source files are scanned via `fs.readdir`
+    // and processed concurrently.
+    return localeCompare(messageA.id, messageB.id);
   });
 }
 


amannn:
Hey, thanks for opening this PR!

I just had a chat with GPT 5.5 about this, and I'm unfortunately not confident that this is the right direction:

---

GPT-5.5's read:

`getSortedMessages` currently returns `0` when two messages have the same first reference, so the final order then falls back to the input order from `messagesById`.

The PR description's same-file example does not seem to be explained by `fs.readdir` / parallel processing. For two unique messages on the same JSX line in the same file, the flow should be:

1. The SWC plugin records messages in an insertion-preserving `IndexMap`.
2. `processFile` iterates the extracted messages sequentially.
3. `messagesById.values()` preserves that insertion order.
4. `toSorted` is stable, so when the comparator returns `0`, that same order is preserved.

So `toSorted` does not reshuffle tied messages. It keeps the previous order. If every extracted message has a reference, the relevant question is whether that previous order can be nondeterministic. For the same-file same-line case, there does not seem to be a reason to believe it is, unless SWC traversal itself is nondeterministic.

The `fs.readdir` / `Promise.all` concern applies more clearly to global `messagesById` insertion order across files. But the final reference sort already removes cross-file insertion order for ordinary unique messages with different `references[0]`.

For duplicate message IDs that are discovered in multiple files, the reference order should also be deterministic: `mergeReferences` sorts merged refs with `compareReferences`, so whichever file finishes first should not affect `references[0]` after the duplicate has been merged. For messages that only appear in one file, refs come from SWC traversal order.

Verdict: this does not look like a correct fix without a repro showing nondeterministic current behavior. It changes the tie policy from source appearance order to ID order, and the existing test that asserted source-order preservation was changed accordingly. If source appearance order for same-line messages was intentional, the PR is a behavior change rather than just deterministic hardening.

Separate note: sorting references again during persistence does not seem necessary if the current invariants hold. Merged multi-file refs are already sorted in `mergeReferences`; single-file refs follow SWC traversal order. A useful improvement, if desired, would be to document or test the invariant that `references[0]` is stable before `getSortedMessages` relies on it.

Also, the added e2e test currently uses strings whose generated IDs sort in the same order as source appearance (`tied-zebra` -> `GboqqG`, `tied-apple` -> `RQGz35`), so it would likely pass even without the tiebreaker. If that test stays, the fixture strings should have IDs that sort opposite to their source order.

packages/next-intl/src/extractor/utils.test.tsx
```diff
@@ -41,7 +41,7 @@ describe('getSortedMessages', () => {
     ).toEqual(['a', 'b', 'c']);
   });
 
-  it('preserves original order when reference paths and lines match', () => {
+  it('uses id as tiebreaker when reference paths and lines match', () => {
```
amannn, line 44: I think I previously added this test specifically for ensuring we keep the "discovery order" from the SWC plugin for messages where both the path and line match.

anthlasserre: Thanks for the context — useful to know the test was intentional.

I think the gap is this: `getSortedMessages` is a pass-through on ties — it doesn't *impose* SWC discovery order, it just doesn't shuffle. Same fixture, three different input orders, three different outputs (all pass against current `main`):

```ts
const a = {id: 'a', message: 'a', references: [{path: 'components/A.tsx', line: 10}]};
const b = {id: 'b', message: 'b', references: [{path: 'components/A.tsx', line: 10}]};
const c = {id: 'c', message: 'c', references: [{path: 'components/A.tsx', line: 10}]};

expect(getSortedMessages([a, b, c]).map(m => m.id)).toEqual(['a', 'b', 'c']);
expect(getSortedMessages([c, a, b]).map(m => m.id)).toEqual(['c', 'a', 'b']);
expect(getSortedMessages([b, c, a]).map(m => m.id)).toEqual(['b', 'c', 'a']);
```

So the *catalog-level* "preserve SWC discovery order" guarantee depends entirely on the input to `getSortedMessages` being in SWC discovery order. For same-file ties it is — they all flow through one `processFile` for-loop in `IndexMap` order. For ties that arise across files, it isn't.

### Repro showing it isn't

Two files, same two msgids each, on the same line, reversed source order:

```jsx
// src/A.tsx (line 4): <div>{t('foo')}{t('bar')}</div>
// src/B.tsx (line 4): <div>{t('bar')}{t('foo')}</div>
```

After `mergeReferences`:
- `foo.references = [A.tsx:4, B.tsx:4]` → `references[0] = A.tsx:4`
- `bar.references = [A.tsx:4, B.tsx:4]` → `references[0] = A.tsx:4`

They tie. `messagesById` insertion position then comes down to whichever `processFile` called `.set(id, …)` first — which depends on `Promise.all` over `fs.readdir`-ordered files.

Running `unstable_extractMessages` against this fixture 5 times against current `main` (catalog cleared between runs):

```
Source files (constant across runs):
  src/A.tsx (line 4): <div>{t('foo')}{t('bar')}</div>
  src/B.tsx (line 4): <div>{t('bar')}{t('foo')}</div>

Run 1: foo before bar
Run 2: foo before bar
Run 3: bar before foo   ← flipped
Run 4: bar before foo
Run 5: foo before bar

Distinct outputs: 2 / 5
```

<details>
<summary>Standalone repro script</summary>

```js
// repro.mjs — run with `node repro.mjs` from your next-intl checkout
import {createRequire} from 'node:module';
import fs from 'node:fs/promises';

const require = createRequire(import.meta.url);
const {unstable_extractMessages} = require('next-intl/extractor');

const dir = '/tmp/next-intl-tie-repro';
await fs.rm(dir, {recursive: true, force: true});
await fs.mkdir(`${dir}/src`, {recursive: true});
await fs.writeFile(
  `${dir}/src/A.tsx`,
  `import {useExtracted} from 'next-intl';
export default function A() {
  const t = useExtracted();
  return <div>{t('foo')}{t('bar')}</div>;
}
`
);
await fs.writeFile(
  `${dir}/src/B.tsx`,
  `import {useExtracted} from 'next-intl';
export default function B() {
  const t = useExtracted();
  return <div>{t('bar')}{t('foo')}</div>;
}
`
);

const outputs = new Set();
for (let i = 0; i < 5; i++) {
  await fs.rm(`${dir}/messages`, {recursive: true, force: true});
  process.chdir(dir);
  await unstable_extractMessages({
    sourceLocale: 'en',
    messages: {path: './messages', format: 'po', locales: 'infer'},
    srcPath: './src'
  });
  outputs.add(await fs.readFile(`${dir}/messages/en.po`, 'utf8'));
}
console.log(`${outputs.size} distinct outputs across 5 runs`);
```

</details>

So "SWC discovery order" isn't really the invariant today — for cross-file ties it's "discovery order from whichever file's `processFile` finished first", which is non-deterministic. The existing test passes because it stays in the single-file regime where the input to `getSortedMessages` *is* the IndexMap order.

The PR trades that partial determinism for full determinism via an `id` tiebreaker. The real cost is the behavior change for same-file ties: they go from source order to id order.

Happy to look at a narrower fix that keeps same-file SWC order specifically, but at sort time both cases produce identical `references` arrays after merge, so I couldn't find a signal to distinguish them. If you have an idea (e.g. tagging messages with the originating `processFile` call id, or sorting `messagesById` deterministically upstream), I'm game.

amannn: Hey, that's a great analysis—thanks! I chatted a bit more with GPT 5.5 on this, and I think this could be a reasonable fix that avoids the ID tie breaker: https://github.com/amannn/next-intl/pull/2324

Can you have a look and let me know what you think?

amannn: I'll merge https://github.com/amannn/next-intl/pull/2324 for the time being. If you have time to have a look, I'd still really appreciate your thoughts!

amannn:
Closed in favor of https://github.com/amannn/next-intl/pull/2324/

PR #2313, opened 2026-04-29
feat: Support multiple message catalog paths for `useExtracted` in monorepos and external packages

amannn:
## Features

- Overload `messages.path` as array to allow consuming external messages
- Add optional `extract.path` to define where to extract to in case `messages.path` is an array

→ [Monorepos and external packages](https://next-intl-docs-git-cursor-monorepo-catalog-pat-99d0bf-next-intl.vercel.app/docs/usage/extraction#monorepos-external-packages)

### `createNextIntlPlugin`

- `messages.locales` → `extract.locales` 
- `srcPath` → `extract.srcPath` 

```diff
-    srcPath: './src',
     extract: {
+      locales: 'infer',
       sourceLocale: 'en',
+      srcPath: './src'
     },
     messages: {
       path: './messages',
-      locales: 'infer',
       format: 'po'
     }
```

### `unstable_extractMessages`

- `srcPath` → `extract.srcPath` 
- `sourceLocale` → `extract.sourceLocale` 

```diff
 await unstable_extractMessages({
-  srcPath: './src',
-  sourceLocale: 'en',
   messages: {
     path: './messages',
-    locales: 'infer',
     format: 'po'
+  },
+  extract: {
+    sourceLocale: 'en',
+    locales: 'infer',
+    srcPath: './src'
   }
 });
```

PR #2316, opened 2026-05-05
feat: Improvements for `useExtracted`

amannn:
⚠️ A small config change is required to upgrade, please see the migration section below (or point your agent of choice at this comment and ask to upgrade).

### Improved monorepo support

- `messages.path` now optionally accepts `Array<string>` to load catalogs from multiple roots
- `extract.path` was added to allow loading messages from multiple roots while extracting only first-party messages

See: [Monorepos and external packages](https://next-intl-docs-git-canary-next-intl.vercel.app/docs/usage/extraction#monorepos-external-packages)

### Stacked descriptions

If a message has multiple, different descriptions, they are now stacked when serializing messages (e.g. multiple `#.` entries for `.po` files). Previously, only the first one was used and other ones discarded.

**Example:**

```po
#. First description
#. Second description
msgid "5VpL9Z"
msgstr "Right"
```

### Disabled development extraction

You can now configure `locales` to `[]` to disable extraction during development. This is useful when you prefer to only extract messages at a later stage with `unstable_extractMessages`.

### Path-only references in default PO codec (line numbers removed)

The built-in PO codec now writes location comments as `#: path` only: line numbers are not serialized, and duplicate paths for the same message are merged. This matches common gettext “file” location style and cuts noise and merge pain on committed catalogs.

```diff
 #. Advance to the next slide
-#: src/components/Carousel.tsx:12
-#: src/components/Carousel.tsx:44
+#: src/components/Carousel.tsx
 msgid "carousel.next"
 msgstr "Right"
```

To keep line numbers, define a [custom codec](https://next-intl.dev/docs/usage/plugin#formats-custom) modeled on the built-in [`POCodec.tsx`](https://github.com/amannn/next-intl/blob/canary/packages/next-intl/src/extractor/format/codecs/POCodec.tsx) (see `pathOnlyRefs`).

### Stabilize extracted catalog order

Yet another fix was merged with https://github.com/amannn/next-intl/pull/2324 that contributes to fully deterministic ordering of messages. Hopefully this will be the one 🤞

### `createNextIntlPlugin`

```diff
 const withNextIntl = createNextIntlPlugin({
   experimental: {
+    extract: true,
     srcPath: './src',
-    extract: {sourceLocale: 'en'},
     messages: {
       path: './messages',
       format: 'po',
-      locales: 'infer'
+      locales: 'infer',
+      sourceLocale: 'en'
     }
   }
 });
```

### `unstable_extractMessages`

```diff
 await unstable_extractMessages({
   srcPath: './src',
-  sourceLocale: 'en',
   messages: {
     path: './messages',
     format: 'po',
-    locales: 'infer'
+    locales: 'infer',
+    sourceLocale: 'en'
   }
 });
```

### Custom codecs (`defineCodec`)

- `ExtractorMessage.description` was changed from `string | undefined` to a mandatory `Array<string>` (use `[]` when there are no descriptions).
- `ExtractorMessage.references` was changed to be mandatory (use `[]` when the format does not store locations).
- Line numbers are no longer emitted in the default `.po` codec. To keep line numbers, define a [custom codec](https://next-intl.dev/docs/usage/plugin#formats-custom) modeled on the built-in [`POCodec.tsx`](https://github.com/amannn/next-intl/blob/canary/packages/next-intl/src/extractor/format/codecs/POCodec.tsx) (see `pathOnlyRefs`).

Please see changes to [`packages/next-intl/src/extractor/format/codecs`](https://github.com/amannn/next-intl/tree/canary/packages/next-intl/src/extractor/format/codecs) for reference.

## Prerelease

```
next-intl@0.0.0-canary-1a95c34
```

---

**PRs**
- [x] https://github.com/amannn/next-intl/pull/2313
- [x] https://github.com/amannn/next-intl/pull/2318
- [x] https://github.com/amannn/next-intl/pull/2317
- [x] https://github.com/amannn/next-intl/pull/2323
- [x] https://github.com/amannn/next-intl/pull/2325
- [x] https://github.com/amannn/next-intl/pull/2326/
- [x] https://github.com/amannn/next-intl/pull/2324


amannn:
@stefanprobst @sss-Mihail-sss @ceolinwill @c0nsoleg @LouisCuvelier Since you asked about better monorepo support, I've added this in https://github.com/amannn/next-intl/pull/2313 now along with [some docs](https://next-intl-docs-git-canary-next-intl.vercel.app/docs/usage/extraction#monorepos-external-packages).

Can you have a look and let me know what you think?

Please also note that there's been a (backwards-compatible) config change, please see the linked PR for details. Would be curious if this works well for you!

A prerelease can be found in the PR description.

LouisCuvelier:
At the moment, our monorepo doesn't have messages in packages but it seems pretty nice to have this option ! Thanks 🔥

ceolinwill:
@amannn This is great! Thanks for working on this.

I asked Codex to implement this for an internal package by providing this PR and the docs link. It did it successfully (see [draft PR](https://github.com/zoonk/zoonk/pull/1466)) without manual intervention needed.

I asked it to provide feedback on the refactor (eg if it hit any roadblocks/issues). This is what it wrote:

- The “ship messages with the external package” approach worked well with a sibling workspace package, custom `.po` codec, precompiled messages, and Turbopack.
- The migration from the old config shape is easy to trip over: once `messages.path` becomes an array, `extract.path` is required, and `messages.locales` must move to `extract.locales`.
- The docs correctly mention merging in `getRequestConfig`, but this is the remaining manual part. A small nested-merge example would help, especially for namespace collisions and deciding app-vs-package override order.
- `transpilePackages` is essential. Without it, the external package can keep raw `useExtracted` and hit the production fallback. The docs call this out, and it matched what we needed.
- For Turborepo users, package extraction should be declared as a build output such as `messages/**`; otherwise the task succeeds but Turbo warns that the package build produced no outputs.

---

I also manually tested this by:

- Extracting translations
- Running `next dev` and `next build`
- Making changes to PO files in the package

Everything seems to be working properly. I didn't notice any bugs so far.

amannn:
@ceolinwill Awesome, thanks for giving it a shot! The draft PR looks really solid to me.

Also thanks for the feedback!



> A small nested-merge example would help

True, I've included this now.



> especially for namespace collisions and deciding app-vs-package override order

Also a good point, I've included a link to the namespaces section now.

> `transpilePackages` is essential.

I have to admit, I'm still not 100% certain what Next.js is doing here. As per [their docs](https://nextjs.org/docs/app/api-reference/config/next-config-js/transpilePackages), my understanding is that this would have already been needed for your `player` package, since you import raw, uncompiled TypeScript. It seems like Next.js is still capable of compiling 3rd party packages if necessary.

I'm not entirely sure what their heuristic is, so I'm mentioning it for now in the `useExtracted` docs. I could imagine for non-TS consumers it might be necessary.

I'd be curious though if it works in your example even without `transpilePackages`. But when in doubt, I'd leave it, as it's certainly what Next.js is already doing in your case.

ceolinwill:
> I'd be curious though if it works in your example even without `transpilePackages`

@amannn You're right. I doubled check this and it does work without `transpilePackages`. I haven't seen this documented anywhere but I think Next.js/Turbopack automatically transpile internal packages we have in a monorepo without the need for `transpilePackages` since I haven't added any of my packages to `transpilePackages` and they're all being handled correctly at both build and dev time.

Edit: btw, I [opened a discussion in the next.js repo](https://github.com/vercel/next.js/discussions/93542) regarding what's the intended behavior around `transpilePackages` (eg when we need this).

stefanprobst:
not sure if i have misconfigured something, but i get this when ``await import(`@repo/ui/i18n/${language}.po`)``:

```
./packages/ui/messages/en.po
Unknown module type
This module doesn't have an associated type. Use a known file extension, or register a loader for it.
```

i'll try to put together a repro

amannn:
> i'll try to put together a repro

That'd be helpful thanks! You can probably start from [`e2e/extracted-monorepo-app`](https://github.com/amannn/next-intl/tree/canary/e2e/extracted-monorepo-app) and adjust it based on your setup until it breaks.

---

A separate note for those following along here: A few more PRs have landed, which resulted in some minor changes for consumers. Please see the updated PR description for details in case you want to try out a canary release.

Thanks a lot everyone for your help here! 🙏

amannn:
@stefanprobst I'm looking to merge this PR pretty soon. Did you have any luck with the reproduction? Otherwise, we can also follow up in a future patch release.

stefanprobst:
@amannn follow up in patch release is totally fine :+1:, i cant promise that i'll get to it this week still.

PR #2317, opened 2026-05-05
feat: Accept `locales: []` to disable extraction

amannn:
Alternative to https://github.com/amannn/next-intl/pull/2292


amannn:
@cursoragent investigate why ci is failing

PR #2318, opened 2026-05-06
feat: Stack descriptions for `useExtracted`

amannn:
## Features

- If a message has multiple, different descriptions, they are now stacked when serializing messages (e.g. multiple `#.` entries for `.po` files). Previously, only the first one was used and other ones discarded.

## Breaking changes for custom codecs

- `ExtractorMessage.description` was changed from `string | undefined` to a mandatory `Array<string>` (use `[]` when there are no descriptions).
- `ExtractorMessage.references` was changed to be mandatory (use `[]` when the format does not store locations).

Please see changes to `packages/next-intl/src/extractor/format/codecs` for reference.

PR #2320, opened 2026-05-08
fix: ensure deterministic message ordering in extracted catalogs

ellisio:
## Summary

`getSortedMessages` used a comparator that wasn't a total order, causing extracted message catalogs (e.g. `messages/en.json`, `messages/en.po`) to drift between extractor runs even when source code hadn't meaningfully changed. The drift surfaced as spurious PR diffs — typically a key swapping with one of its neighbors — that often couldn't be reproduced by re-running extraction on a stable tree.

## Root cause

```ts
if (!refA || !refB) return 0;
return compareReferences(refA, refB);
```

The `return 0` branch fires whenever either side lacks a `references` entry, which violates transitivity. Counterexample:

- A: `{ path: "auth.ts", line: 50 }`
- B: no ref
- C: `{ path: "auth.ts", line: 10 }`

Then `cmp(A, B) === 0` and `cmp(B, C) === 0`, but `cmp(A, C) > 0`. `Array.prototype.toSorted` is stable, but stability doesn't rescue a non-transitive comparator — output depends on input order and engine-specific Timsort pivot choices.

In practice, `JSONCodec.decode` doesn't round-trip references, so every fresh extractor process starts with reference-less messages and adds references as files are processed. Any save that fires before all files have been processed (incremental dev writes, watcher events, the first save after a partial pass) snapshots a mixed state and exposes the bug.

## Fix

Promote the comparator to a proper total order:

1. Push reference-less messages after referenced ones.
2. Tiebreak by `id` for both reference-less pairs and same-`(path, line)` pairs.

This preserves the existing intent (group by file path, then line) while making output independent of input order and parallel extraction timing.

## Impact

- **One-time reordering** of existing catalogs where the buggy comparator left reference-less messages or same-`(path, line)` ties in input-order-dependent positions. After that diff, catalogs converge to a stable canonical order.
- **No behavioral change** for catalogs where every message has a unique `(path, line)` reference.
- Applies identically to JSON and PO output (both go through `getSortedMessages`).

## Tests

- Added a regression test in `utils.test.tsx` that sorts a mixed set (referenced + reference-less) twice with shuffled input and asserts the outputs match.
- Updated the existing same-`(path, line)` test to expect id-based tiebreaking instead of input order.
- Refreshed 3 PO snapshots in `ExtractionCompiler.test.tsx` that depended on the old insertion-order behavior; they now reflect deterministic id ordering at same `(path, line)`.

### DIFF
--- a/packages/next-intl/src/extractor/ExtractionCompiler.test.tsx
+++ b/packages/next-intl/src/extractor/ExtractionCompiler.test.tsx
@@ -1027,12 +1027,12 @@ msgstr "World"
       msgstr "Hey!"
 
       #: src/Greeting.tsx:5
-      msgid "jqdzk6"
-      msgstr "World"
-
-      #: src/Greeting.tsx:5
       msgid "ODGmph"
       msgstr "!"
+
+      #: src/Greeting.tsx:5
+      msgid "jqdzk6"
+      msgstr "World"
       ",
       ]
     `);
@@ -1093,11 +1093,11 @@ msgstr ""
       msgstr "Hallo!"
 
       #: src/Greeting.tsx:5
-      msgid "jqdzk6"
+      msgid "ODGmph"
       msgstr ""
 
       #: src/Greeting.tsx:5
-      msgid "ODGmph"
+      msgid "jqdzk6"
       msgstr ""
 
       #: src/Greeting.tsx:5
@@ -1165,7 +1165,7 @@ msgstr ""
       msgstr "Hallo!"
 
       #: src/Greeting.tsx:5
-      msgid "jqdzk6"
+      msgid "I5NMJ8"
       msgstr ""
 
       #: src/Greeting.tsx:5
@@ -1173,11 +1173,11 @@ msgstr ""
       msgstr ""
 
       #: src/Greeting.tsx:5
-      msgid "pE58D7"
+      msgid "jqdzk6"
       msgstr ""
 
       #: src/Greeting.tsx:5
-      msgid "I5NMJ8"
+      msgid "pE58D7"
       msgstr ""
       ",
       ]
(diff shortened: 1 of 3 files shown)


amannn:
Coming from here: https://github.com/amannn/next-intl/discussions/2036#discussioncomment-16857791

amannn:
Hey, I just chatted with GPT-5.5 about this PR, and it seems like a missing reference either points to a different issue or that something else has caused the non-determinism.

That being said, https://github.com/amannn/next-intl/pull/2318 might improve things here (planning to release this week).

---

From gpt-5.5:

Before the recent extractor changes, `getSortedMessages` allowed messages without references and returned `0` whenever either side of the comparison had no reference. As described here, that means the comparator was not a total order. In a mixed set of referenced and reference-less messages, the final order could therefore depend on the input order.

The suggested fallback ordering does not look like the best direction for the current implementation. In the current extractor pipeline, active extracted messages are expected to have references. If a message reaches sorting without a reference, that points to an invariant violation earlier in the pipeline rather than an ordering case that should be normalized silently.

This was improved in https://github.com/amannn/next-intl/pull/2318. `CatalogManager` now treats source files as the source of truth for extraction-owned fields: it scans the source tree, keeps per-occurrence `SourceMessage`s, and rebuilds aggregated catalog messages, including references, from those source messages. Decoded catalog data is only merged for ancillary metadata, not for references.

So the remaining check added in https://github.com/amannn/next-intl/commit/3754cb31 is diagnostic: if sorting ever sees a message without a file reference, it warns once for that message id. This should give signal if the invariant is still violated somewhere, without adopting reference-less messages as a normal persisted state.

ellisio:
Closed due to feedback given over in the discussion. We'll bump the package when you release and report back if it still happens. 🍻

PR #2324, opened 2026-05-12
fix: Stabilize extracted catalog order

amannn:
## Summary

When multiple messages tie on the same primary reference (`references[0]`), catalog order fell through to `messagesById` insertion order. Files were processed with `Promise.all`, so that order depended on which source file finished first—leading to non-deterministic `.po`/`.json` output.

This PR keeps extraction parallel but **applies** extracted messages to internal maps in **sorted source-file order**. Watcher batches also sort events by path before processing, so dev/watch behaves consistently.

## Relation to #2315

[PR #2315](https://github.com/amannn/next-intl/pull/2315) discussed the same tie-breaking issue and proposed sorting ties by message `id` inside `getSortedMessages`. That fixes churn globally but changes semantics for same-line ties within one file (source order vs id order).

Here we fix the cross-file race **without** changing `getSortedMessages`: stable ordering comes from deterministic merge order instead of an id tiebreaker.

## Testing

- New regression: `orders same-reference messages independently of file processing order` in `ExtractionCompiler.test.tsx`.
- Ran: `pnpm --filter next-intl test run src/extractor/ExtractionCompiler.test.tsx` (all passing).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/next-intl/src/extractor/ExtractionCompiler.test.tsx
+++ b/packages/next-intl/src/extractor/ExtractionCompiler.test.tsx
@@ -1014,6 +1014,58 @@ describe('po format', {timeout: 20_000}, () => {
     `);
   });
 
+  it('orders same-reference messages independently of file processing order', async () => {
+    async function extractWithDelayedFile(delayedFileName: string) {
+      filesystem.project = {
+        src: {
+          'A.tsx': `
+          import {useExtracted} from 'next-intl';
+          export default function A() {
+            const t = useExtracted();
+            return <div>{t('apple')}{t('banana')}</div>;
+          }
+          `,
+          'B.tsx': `
+          import {useExtracted} from 'next-intl';
+          export default function B() {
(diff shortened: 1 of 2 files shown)


packages/next-intl/src/extractor/ExtractionCompiler.test.tsx
```diff
@@ -1014,6 +1014,58 @@ describe('po format', {timeout: 20_000}, () => {
...
+      await waitForWriteFileCalls(1);
+
+      return filesystem.project.messages!['en.po'];
+    }
+
+    const aDelayedContent = await extractWithDelayedFile('A.tsx');
+    const bDelayedContent = await extractWithDelayedFile('B.tsx');
+
+    expect(aDelayedContent).toEqual(bDelayedContent);
+  });
```
amannn, line 1067: This test failed consistently before adding the fix.

PR #2330, opened 2026-05-18
feat: Use URL-safe base64 for auto-generated message keys of `useExtracted`

### LINKED ISSUE #2250: useExtracted auto-generated message keys starting with `/` cause Google crawler 404 errors
### Description

When using `useExtracted` with auto-generated message keys, some keys begin with the `/` character. Google's crawl bot interprets these as URL paths and attempts to crawl them, resulting in **Not Found (404)** errors reported in Google Search Console.

This is because the auto-generated key character set includes `/`, which is a valid URL path character. When these keys appear in the rendered HTML or page source, Googlebot treats them as relative URLs.

**Screenshot from Google Search Console:**

Image

### Verifications

- [x] I've read the [docs](https://next-intl.dev). In particular, I've looked at the relevant [usage guide](https://next-intl.dev/docs/usage/extraction).
- [x] I've searched for [existing issues and discussions](https://github.com/amannn/next-intl/issues) on GitHub and [Stack Overflow](https://stackoverflow.com/search?q=next-intl).
- [x] I've compared my setup to a [working example](https://next-intl.dev/examples).

This issue is not reproducible in a code sandbox — it manifests in Google Search Console when Googlebot crawls a production site using `useExtracted` with auto-generated keys. The auto-generated keys (e.g. ones starting with `/`) appear in the page source and Googlebot interprets them as relative URL paths, then reports 404 errors.

### Reproduction description

1. Use `useExtracted` with auto-generated message keys in a Next.js app
2. Deploy to production and let Google crawl the site
3. Open Google Search Console → Pages → Not Found (404)
4. Observe 404 entries corresponding to auto-generated message keys that start with `/`

### Expected behaviour

Auto-generated message keys should not use characters that can be interpreted as URL paths by web crawlers. The character set should exclude URL-significant characters like `/`, `?`, `#`, `&`.

**Suggestion:**
- Use a character set for auto-generated keys that excludes URL-like characters, OR
- Allow users to customize the character set / prefix used for auto-generated keys

spokodev:
Auto-generated `useExtracted` message keys now use **URL-safe base64** (`-_`) instead of standard base64 (`+/`).

Standard base64 can produce keys starting with `/` (e.g. `/qP5M+`). When those keys appear in serialized page data, Googlebot treats `/`-prefixed values as relative URL paths and reports 404s in Search Console ([#2250](https://github.com/amannn/next-intl/issues/2250)).

The hash input is unchanged (still SHA-512, first 6 characters). Only the alphabet differs—keys no longer contain URL-significant `+` or `/` characters.

```diff
 // messages/en.json (example)
 {
-  "+YJVTi": "Hey!",
+  "-YJVTi": "Hey!",
   "NhX4DJ": "Hello"
 }
```

```diff
 // messages/en.po (example)
 #: src/components/Greeting.tsx
-msgid "+YJVTi"
+msgid "-YJVTi"
 msgstr "Hey!"
```

## Breaking change

Keys for messages whose hash previously encoded `+` or `/` in the first 6 characters get a new ID (e.g. `+YJVTi` → `-YJVTi`, `nm/7yQ` → `nm_7yQ`). Most keys are unaffected.

To migrate, in all locale catalogs, search & replace **message keys only** (JSON keys / `msgid` values—not `msgstr` content):

- `+` → `-`
- `/` → `_`

Only keys that previously contained `+` or `/` need updating, most keys are unchanged.

---

Fixes #2250


amannn:
Hey @spokodev, thanks for opening this PR!

Yes, I think this is the right approach and that we should go with this.

Let me see when we can release this, I'll get back to you …

spokodev:
Thanks for the quick look, @amannn — appreciated. Happy to rebase if anything drifts before the release window. Flipping out of draft.

PR #2327, opened 2026-05-13
docs: Migrate examples to `next/root-params`

amannn:
**TODO**
- [ ] Upgrade to Next.js 16.3 (removes `experimental.rootParams`): https://github.com/amannn/next-intl/pull/2354
- [x] Close https://github.com/amannn/next-intl/pull/2331

PR #2331, opened 2026-05-20
The setup page should instruct users to set the default locale before loading the 404 page

em-te:


amannn:
`setRequestLocale` is now deprecated in favor of `next-root-params`:

https://github.com/amannn/next-intl/releases/tag/v4.13.5

Examples are being updated as part of https://github.com/amannn/next-intl/pull/2327

PR #2333, opened 2026-05-21
feat: Expose `forcePrefix` on `Link` and `useRouter` (#2332)

### LINKED ISSUE #2332: Extend `forcePrefix` opt-out to `Link` and `useRouter` to fix locale switching under `output: 'export'`
## Summary

`redirect()` accepts a `forcePrefix` option since [#1845](https://github.com/amannn/next-intl/issues/1845) / [PR #1865](https://github.com/amannn/next-intl/pull/1865) (shipped in 4.1.0). The same option is not available on `Link` or `useRouter().push/replace/prefetch`, where `forcePrefix: true` is hardcoded whenever the `locale` prop is set. With `output: 'export'` + `localePrefix: 'as-needed'`, that produces 404s when switching to the default locale.

## Setup

```ts
// src/i18n/routing.ts
  locales: ['en', 'sv-se'],
  defaultLocale: 'en',
  localePrefix: 'as-needed',
  localeCookie: false
});
```

```js
// next.config.js
module.exports = { output: 'export' };
```

The project uses **static routes** (no `[locale]` dynamic segment, no middleware), because `output: 'export'` has no runtime to do locale rewrites:

```
src/app/
  (en)/
    layout.tsx          # <html lang="en">, setRequestLocale('en')
    page.tsx            # /
    about/page.tsx      # /about
  (sv-se)/
    layout.tsx          # <html lang="sv-se">, setRequestLocale('sv-se')
    sv-se/
      page.tsx          # /sv-se
      about/page.tsx    # /sv-se/about
```

`next build` emits exactly:

```
out/
  index.html          # /
  about.html          # /about
  sv-se.html          # /sv-se
  sv-se/about.html    # /sv-se/about
```

There is no `out/en/...` in the export — `localePrefix: 'as-needed'` means default-locale routes live at the unprefixed paths.

## Root cause

In `createSharedNavigationFns`, `locale` is translated to `forcePrefix: true`:

```js
const g = l ? y({ locale: t || h, href: ..., forcePrefix: null != t || void 0 }) : f;
```

Same pattern in the client `useRouter` wrappers for `push`/`replace`/`prefetch`. There's no way to override from userland.

`useRouter().push('/about', { locale: 'en' })` has the same problem; extra options like `forcePrefix: false` are destructured out and discarded.

## Why the existing guidance doesn't cover static export

The [docs for `Link` + `locale`](https://next-intl.dev/docs/routing/navigation#link-locale) and the discussion in [#1791](https://github.com/amannn/next-intl/issues/1791) both recommend the same workaround: drop the `locale` prop. That works because middleware rewrites `/en/...` → `/...` and the cookie sync happens server-side. With `output: 'export'`:

- No middleware exists to rewrite the prefixed URL.
- The prefixed default-locale path (`/en/about`) is not in the export at all.
- With `localeCookie: false`, there is no cookie to "defeat" in the first place.

So the forced prefix doesn't protect anything — it just breaks navigation.

## Proposed fix

Mirror what was done for `redirect` in [#1845](https://github.com/amannn/next-intl/issues/1845) / [PR #1865](https://github.com/amannn/next-intl/pull/1865): expose `forcePrefix` as a prop on `Link` and an option on `useRouter().push/replace/prefetch`. Backwards-compatible, opt-in, and consistent across the three navigation APIs.

Optionally, when `localeCookie: false` and `localePrefix: 'as-needed'` (where the cookie-defeating rationale doesn't apply), the implicit `forcePrefix: true` could be relaxed by default — but exposing the explicit opt-out alone is enough to unblock this case.

Happy to open a PR if the direction sounds reasonable.

## Workaround

Bypass `Link` for locale-switch links; render the precomputed alternate URL via `next/link` directly. Functional, but inconsistent with the rest of the app.

## Environment

- `next-intl` 4.12.0
- `next` 16.2.6, App Router, `output: 'export'`
- `localePrefix: 'as-needed'`, `localeCookie: false`

### Verifications

https://github.com/jarkkosyrjala/next-intl-repro-force-prefix-static-export

## Steps to reproduce:

1. Open <https://jarkkosyrjala.github.io/next-intl-repro-force-prefix-static-export/sv-se/about/>.
2. Click the "English" link in the locale switcher.
3. See error: GitHub Pages 404. The browser navigated to `…/en/about/`, which is not in the static export.

### Expected behaviour

The "English" locale switcher on `/sv-se/about/` should render `<a href="/about/">` and navigate to the existing default-locale page. Instead it renders `<a href="/en/about/">` — a path that does not exist in the export (because `localePrefix: 'as-needed'` keeps default-locale routes unprefixed) — producing a 404. There is currently no way to opt out of the forced prefix from `<Link>` or `useRouter`, only from `redirect()`.

jarkkosyrjala:
Fixes #2332.

## Summary

Mirrors the existing `forcePrefix` option on `redirect` for `Link` and `useRouter().push/replace/prefetch`. Today, whenever a `locale` prop is set on `Link` (or passed to `useRouter`), a prefix is implicitly forced so a locale cookie can be updated server-side before the request is rewritten to the unprefixed route. There's no userland opt-out.

With `output: 'export'`, this is broken by design: there is no middleware to perform the rewrite, the prefixed default-locale path (`/en/about`) isn't part of the static export, and with `localeCookie: false` there's no cookie to defeat in the first place. The forced prefix produces a 404 link.

This PR exposes `forcePrefix` so users can opt out:

```tsx
// With `localePrefix: 'as-needed'` and `en` as the default locale:
<Link href="/about" locale="en" forcePrefix={false}>About</Link>     // → /about
<Link href="/about" locale="de" forcePrefix={false}>Über uns</Link>  // → /de/about

router.push('/about', {locale: 'en', forcePrefix: false}); // → /about
router.push('/about', {locale: 'de', forcePrefix: false}); // → /de/about
```

## Semantics

| `forcePrefix` | Behavior                                                                             |
| ------------- | ------------------------------------------------------------------------------------ |
| omitted       | Unchanged — implicit prefix when `locale` is set (preserves the cookie-update flow). |
| `true`        | Force a prefix regardless of `localePrefix` setting (matches `redirect()`).          |
| `false`       | Opt out of the implicit force; defer to natural `localePrefix` behavior.             |

Note: for `Link`/`useRouter`, `forcePrefix={false}` is interpreted as "defer to `localePrefix`" rather than "never prefix" (the literal `redirect()` semantics). This is the behavior locale-switcher code actually wants and avoids a footgun where secondary-locale links would lose their prefix.

## Test plan

- [x] Unit tests added for `Link` and each `useRouter` method, covering both default-locale opt-out (no prefix) and secondary-locale opt-out (natural prefix retained). Full suite passes (902 tests).
- [x] Manually verified with the issue's reproduction repo under `output: 'export'` + `localePrefix: 'as-needed'`:
  - From `/about/`: English → `/about/`, Swedish → `/sv-se/about/` ✓
  - From `/sv-se/about/`: English → `/about/`, Swedish → `/sv-se/about/` ✓
- [x] `eslint`, `prettier`, `tsc --noEmit` clean.

## Docs

Updated the existing `<Details id="link-locale">` callout in `docs/.../navigation.mdx` to document the opt-out and the deferred-to-natural-prefixing semantics, with an example covering both the default-locale and secondary-locale outcomes.

## Not included / follow-up

The issue also floated two related improvements that I've left out of this PR to keep it focused:

1. A `forcePrefix` option on `defineRouting` so the opt-out can be set routing-wide instead of per call.
2. Automatically relaxing the implicit force when `localeCookie: false` + `localePrefix: 'as-needed'` (no new API — the cookie-defeating rationale doesn't apply in that config).

Which direction would you prefer for the follow-up? My read is that (2) is the more ergonomic default for the static-export case and doesn't expand the API surface, while (1) is more explicit. Happy to open a follow-up PR for whichever you'd like.

### DIFF
--- a/docs/src/pages/docs/routing/navigation.mdx
+++ b/docs/src/pages/docs/routing/navigation.mdx
@@ -169,6 +169,22 @@ The reason for this is that a potential [cookie](/docs/routing/configuration#loc
 
 If you'd like to avoid this behavior, you can instead use [`useRouter`](#userouter) to switch the locale, which can rely on updating the cookie on the client side before navigating to the target page.
 
+Alternatively, you can opt out of the forced prefix by passing `forcePrefix={false}`. The link will then follow your [`localePrefix`](/docs/routing/configuration#localeprefix) setting — switching to the default locale produces an unprefixed pathname, switching to a secondary locale still produces a prefix. This is useful when there's no cookie to defeat in the first place, e.g. when using [`output: 'export'`](https://nextjs.org/docs/app/guides/static-exports) where there's no middleware to perform the rewrite:
+
+```tsx
+// With `localePrefix: 'as-needed'` and `en` as the default locale:
+// Links to `/about`
+<Link href="/about" locale="en" forcePrefix={false}>
+  About
+</Link>
+// Links to `/de/about`
+<Link href="/about" locale="de" forcePrefix={false}>
+  Über uns
+</Link>
+```
+
+The same option is available on [`useRouter`](#userouter) (`push`, `replace`, `prefetch`).
+
 </Details>
 
 <Details id="link-unknown-routes">
--- a/packages/next-intl/src/navigation/createNavigation.test.tsx
+++ b/packages/next-intl/src/navigation/createNavigation.test.tsx
@@ -794,6 +794,25 @@ describe.each([
         expect(markup).toContain('href="/en/about"');
       });
 
+      it('can opt out of the forced prefix when linking to the default locale', () => {
+        mockCurrentLocale('de');
+        const markup = renderToString(
+          <Link forcePrefix={false} href="/about" locale="en">
+            About
+          </Link>
+        );
+        expect(markup).toContain('href="/about"');
+      });
+
+      it('still applies a natural prefix for a secondary locale when `forcePrefix={false}`', () => {
+        const markup = renderToString(
+          <Link forcePrefix={false} href="/about" locale="de">
+            Über uns
+          </Link>
+        );
+        expect(markup).toContain('href="/de/about"');
+      });
+
       it('renders an object href', () => {
         render(
           <Link href={{pathname: '/about', query: {foo: 'bar'}}}>About</Link>
--- a/packages/next-intl/src/navigation/react-client/createNavigation.test.tsx
+++ b/packages/next-intl/src/navigation/react-client/createNavigation.test.tsx
@@ -533,6 +533,20 @@ describe("localePrefix: 'as-needed'", () => {
         expect(useNextRouter()[method]).toHaveBeenCalledWith('/en/about');
       });
 
+      it('can opt out of the forced prefix when switching to the default locale', () => {
+        invokeRouter((router) =>
+          router[method]('/about', {locale: 'en', forcePrefix: false})
+        );
+        expect(useNextRouter()[method]).toHaveBeenCalledWith('/about');
+      });
+
+      it('still applies a natural prefix for a secondary locale when `forcePrefix: false`', () => {
+        invokeRouter((router) =>
+          router[method]('/about', {locale: 'de', forcePrefix: false})
+        );
+        expect(useNextRouter()[method]).toHaveBeenCalledWith('/de/about');
+      });
+
       it('sets a cookie when switching to the default locale', () => {
         global.document.cookie = 'NEXT_LOCALE=de';
         mockCurrentLocale('de');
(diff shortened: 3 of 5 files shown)


amannn:
https://github.com/amannn/next-intl/issues/2332#issuecomment-4516671932

jarkkosyrjala:
Closing this PR in favour of the the alternate approach that will possibly be introduced in the next major version, see https://github.com/amannn/next-intl/issues/2332#issuecomment-4564264651 and https://github.com/amannn/next-intl/issues/2047

> - When a user explicitly sets both `localeCookie: false` and `localePrefix: 'as-needed'`, there's no need to add a locale prefix to `Link` ([ref](https://github.com/amannn/next-intl/issues/2332#issuecomment-4520769401))

PR #2338, opened 2026-05-26
fix: declaration file extension in compileDeclaration

bennajah:
fix: declaration file extension in compileDeclaration


amannn:
> In TypeScript 5.0, when an import path ends in an extension that isn’t a known JavaScript or TypeScript file extension, the compiler will look for a declaration file for that path in the form of `{file basename}.d.{extension}.ts`

[source](https://www.typescriptlang.org/tsconfig/#allowArbitraryExtensions)

PR #2344, opened 2026-06-05
feat: Extract `useTranslations`/`getTranslations` usages in SWC plugin

amannn:
## Summary

Extends the SWC extractor plugin to also collect statically-analyzable `useTranslations` / `getTranslations` key usages, alongside the existing inline-message extraction (`useExtracted` / `getExtracted`).

The plugin's `results` payload is a **tagged union** (`type: "extracted" | "translation"`):

- **`extracted`** — an inline message from `useExtracted` / `getExtracted`: `{ type: "extracted", id, message, description, reference }`
- **`translation`** — a `useTranslations` / `getTranslations` key reference: `{ type: "translation", id, reference }`, where `id` is the fully-qualified `namespace.key`

### `translation` extraction

- `const t = useTranslations('ns'); t('key')` → `id: "ns.key"`
- root namespace `useTranslations(); t('key')` → `id: "key"`
- `rich` / `markup` / `has` / `raw` member calls are recognized
- the id is resolved as far as is statically known — a dynamic key under a namespace records the namespace (`id: "ns"`); a dynamic key in the global namespace is skipped
- `await getTranslations(...)` (server) is supported

## Consumer

next-intl's `MessageExtractor` consumes the `extracted` variant; the catalog `SourceMessage` type is renamed to `SourceExtractedMessage` to make explicit that the pipeline only handles extracted messages.

## Tests

Covered by the `tests/fixture/**` snapshots — each fixture's `output.json` carries the tagged union (incl. `use-translations`, `use-translations-dynamic-key`, `get-translations`, `use-translations-with-extracted`).


amannn:
@claude this repo currently uses a separate workflow for stable releases and prereleases. for trusted publishing, i can only specify a single workflow file. you need to combine both into a single file while making sure only canary publishing or stable publishing is triggered based on the correct branch name

PR #2350, opened 2026-06-23
fix: Deprecate `toJSONString` of custom formats (prints a deprecation warning)

amannn:
## Why

`toJSONString` was only ever called in one place: the `!precompile` branch of the catalog loader (`plugin/catalog/catalogLoader.tsx`). The `precompile` branch already loads catalogs through `decode`. Since `precompile` is set to become the default, `toJSONString` is redundant—`decode` carries all the information needed to load messages into the app.

## What changed

- **Loader no longer calls `toJSONString`.** Both branches now `decode` the source. The `!precompile` branch builds the nested messages object from the decoded messages (`buildMessages`), mirroring `precompileMessages` but keeping the raw strings. The string/array/object validation is shared via `assertStringMessage`. Output is identical after `JSON.parse` for the built-in JSON/PO codecs.
- **Built-in codecs drop `toJSONString`** (`JSONCodec`, `POCodec`); they're loaded via `decode` like everything else.
- **The type marks it `@deprecated` and optional** (`toJSONString?`), so existing custom codecs keep type-checking.
- **A one-time deprecation warning** fires if a resolved codec still provides `toJSONString`, pointing the author to remove it.
- **Docs and fixtures updated** to no longer reference `toJSONString` (the custom-format example in `plugin.mdx`, plus the `JSONCodecStructured` / `POCodecSourceMessageKey` test fixtures).

## Soundness note

For the built-in JSON/PO codecs this is behavior-preserving. For **custom** codecs whose old `toJSONString` produced a non-nested shape (e.g. flat dotted keys), the runtime output now matches the `precompile` path (nested via `setNestedProperty`). This unifies the two branches on the shape next-intl actually expects at runtime, and the deprecation warning prompts authors to remove the now-unused method.

## Verification

- `vitest run src/extractor` passes (53 tests), covering codec decode/encode and the structured/custom fixtures.
- ESLint + Prettier pass on all changed files.
- `tsc --noEmit` introduces no new errors in the changed files (the only remaining message is the pre-existing `icu-minify/compile` resolution quirk that requires a built `icu-minify` package).

---
_Generated by [Claude Code](https://claude.ai/code/session_01KV7kCx7pyhC2FrMw6fGzsj)_

### DIFF
--- a/packages/next-intl/src/plugin/catalog/catalogLoader.tsx
+++ b/packages/next-intl/src/plugin/catalog/catalogLoader.tsx
@@ -16,6 +16,8 @@ import type {TurbopackLoaderContext} from '../types.js';
 // create multiple loader instances so don't expect a singleton.
 let cachedCodec: ExtractorCodec | null = null;
 
+let hasWarnedAboutToJSONString = false;
+
 type CompiledMessageCacheEntry = {
   compiledMessage: unknown;
   messageValue: string;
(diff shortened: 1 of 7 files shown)


packages/next-intl/src/plugin/catalog/catalogLoader.tsx
```diff
@@ -62,26 +64,66 @@ export default function catalogLoader(
...
       callback(null, result);
     })
     .catch(callback);
 }
 
+function warnIfToJSONStringProvided(codec: ExtractorCodec) {
+  if (hasWarnedAboutToJSONString || !codec.toJSONString) return;
+  hasWarnedAboutToJSONString = true;
+  console.warn(
+    '`toJSONString` is deprecated and no longer used—catalogs are now loaded via `decode`. Remove `toJSONString` from your codec.'
```
amannn:
```suggestion
    '`toJSONString` is deprecated and no longer used, please remove it.'
```

packages/next-intl/src/plugin/catalog/catalogLoader.tsx
```diff
@@ -62,26 +64,66 @@ export default function catalogLoader(
...
       callback(null, result);
     })
     .catch(callback);
 }
 
+function warnIfToJSONStringProvided(codec: ExtractorCodec) {
+  if (hasWarnedAboutToJSONString || !codec.toJSONString) return;
+  hasWarnedAboutToJSONString = true;
+  console.warn(
+    '`toJSONString` is deprecated and no longer used, please remove it.'
```
amannn:
```suggestion
    '`toJSONString` is deprecated and no longer needed, please remove it.'
```

PR #2355, opened 2026-07-08
fix: Next.js 16.3 compatibility preparation (only update locale cookie for document requests and not prefetches)

amannn:
## Summary

Fixes a regression that appears with the Next.js 16.3 client-side router: after switching **back to the default locale** via `<Link />`, the `NEXT_LOCALE` cookie is overwritten back to the previous locale by background requests of the client-side router. On 16.3 this made `main.spec.ts › keeps the locale cookie updated …` fail deterministically (0/8 local runs passed) and `main.spec.ts › can use Link to link to the root of another language` fail intermittently on CI.

All statements below are from instrumented runs (CDP network capture of the real click sequence in the playground, `NEXT_PUBLIC_USE_CASE=main`, production build), not assumptions.

## The exact HTTP requests

Scenario: start on `/` (en) → click "Switch to German" → on `/de` → click "Zu Englisch wechseln" (a locale switch back to the default locale, `localePrefix: 'as-needed'`).

**Next.js 16.2.2 (current `main`) — after the switch-back click, exactly one request:**

```
GET /            Cookie: NEXT_LOCALE=en   → 200   (no Set-Cookie)
```

Final state: URL `/`, cookie `en` ✅. The client-side cookie write in `syncLocaleCookie` (runs on link click) is the only writer, and nothing contradicts it.

**Next.js 16.3.0-canary.78 — the same click produces the navigation *plus* a burst of background requests:**

```
GET /                     Cookie: en   rsc:1                                        → 200  (no Set-Cookie)
GET /de                   Cookie: en   rsc:1  next-router-prefetch:1  …segment-prefetch:/_tree → 200  Set-Cookie: NEXT_LOCALE=de
GET /de/client            Cookie: en   rsc:1  next-router-prefetch:1  …                        → 200  Set-Cookie: NEXT_LOCALE=de
GET /de/about             Cookie: en   rsc:1  next-router-prefetch:1  …                        → 200  Set-Cookie: NEXT_LOCALE=de
GET /de/verschachtelt     Cookie: en   rsc:1  next-router-prefetch:1  …                        → 200  Set-Cookie: NEXT_LOCALE=de
(+ further /de/* requests once the cookie is already `de` again, then without Set-Cookie)
```

Final state: URL `/`, cookie `de` ❌.

The `/de/*` requests reuse the `_rsc` cache keys of the routes that were prefetched/visited while the user was on `/de` — the 16.3 router revalidates existing client-router cache entries after a navigation. Each of these runs the middleware, which sees "path locale `de` ≠ cookie `en`" in `syncCookie` and re-asserts `Set-Cookie: NEXT_LOCALE=de`, clobbering the `en` that the switch had just set. Verifiable in isolation:

```
$ curl -I --cookie "NEXT_LOCALE=en" -H "RSC: 1" http://localhost:3000/de
set-cookie: NEXT_LOCALE=de; Path=/; …
```

This also explains the intermittent URL failure of the `Link to the root of another language` test on CI: if a `/de` revalidation lands before the `/en → /` redirect reconciliation completes, the destination `/` is resolved with cookie `de` again and redirects back to `/de` (`GET /` with cookie `de` → `307 /de`, measured). Fast machines: reconciliation wins (passes locally 8/8); loaded CI runners: revalidation wins (fails).

## Why gating on `Next-Router-Prefetch` is not possible

The obvious fix — skip `syncCookie` for requests with prefetch headers — does not work: **Next.js strips its router headers before invoking the middleware, by design.** In `next/src/server/web/adapter.ts` (comment in source: *"Headers should only be stripped for middleware"*), all `FLIGHT_HEADERS` are deleted from the request before the middleware runs:

```
rsc, next-router-state-tree, next-router-prefetch, next-hmr-refresh, next-router-segment-prefetch
```

This code is identical in `16.2.2` and `16.3.0-canary.91` (see `dist/server/web/adapter.js` + `dist/client/components/app-router-headers.js` in the published packages; the only exception is the `skipMiddlewareUrlNormalize` escape hatch, which a library can't rely on). Also confirmed empirically twice:

- Dumping `request.headers` inside the deployed middleware during the real browser flow: every single RSC request (navigations and prefetches alike) arrives with only `accept, accept-encoding, accept-language, connection, cookie, host, next-url, referer, sec-ch-ua*, sec-fetch-dest, sec-fetch-mode, sec-fetch-site, user-agent, x-forwarded-*`.
- A probe request on `16.3.0-canary.91` sending `RSC: 1`, `Next-Router-Prefetch: 1`, `Next-Router-Segment-Prefetch`, `Next-Router-State-Tree` plus a control header `X-Custom-Probe`: the middleware observed `x-custom-probe`, but none of the router headers.

(The `_rsc` search param is likewise stripped for the middleware.)

## The fix

The one reliable, middleware-visible discriminator is **`Sec-Fetch-Dest`** (set by all modern browsers):

- Real document navigations: `sec-fetch-dest: document` (measured: the initial page load)
- All background router fetches: `sec-fetch-dest: empty` (measured: 47/47 during the flow)

`syncCookie` now only updates the cookie for document requests. Requests without a `Sec-Fetch-Dest` header (older browsers, non-browser clients) keep the previous behavior.

Why this is safe:
- **Hard navigations / first visits** are document requests → cookie sync unchanged. This includes users **without JavaScript** — covered by a new minimal e2e app (`e2e/no-js`) that runs the full switch flow with `javaScriptEnabled: false` and asserts the cookie after each document navigation. (Negative-tested: an over-aggressive gate that also blocks document requests makes this test fail.)
- **Locale switches via `Link` / `useRouter`** already write the cookie client-side (`syncLocaleCookie`) — this was the mechanism that made the switch work in the first place; the middleware write on soft navigations was redundant for these flows.
- **Prefetches/revalidations** become side-effect-free, which they should be regardless of this bug.

Trade-off (intentional): a client-side navigation to a prefixed pathname that does *not* go through the `locale` prop (e.g. a plain `next/link` pointing directly at `/de`) no longer updates the cookie server-side. There is no way to distinguish such a request from a prefetch in the middleware on 16.3.

## New e2e apps

1. **`e2e/no-js`** (runs on the workspace `next` version): Locale switching + cookie updates via pure document requests, with JavaScript disabled. Guards the `Sec-Fetch-Dest` gate from the other direction: document requests must continue to update the cookie.
2. **`e2e/partial-prefetching`** (intentionally pinned to `next@16.3.0-canary.91`, since the flags don't exist in stable yet): Verifies next-intl works with `cacheComponents: true` + `partialPrefetching: true` from [Next.js 16.3 Instant Navigations](https://nextjs.org/blog/next-16-3-instant-navigations). With `generateStaticParams` + `setRequestLocale`, localized pages build as **◐ Partial Prerender** routes (static shell + a `Suspense`-streamed dynamic hole that reads the locale cookie). The test verifies client-side navigation, locale switching, and that both shell and streamed content render correctly — all navigations stay client-side (asserted via a `window` marker that survives the whole flow). One quirk this surfaced: the 16.3 router keeps the DOM of previously visited routes in the document (hidden), so test locators must filter for visible elements.

## Verification

Commit structure (as discussed):
1. **The fix** (`syncCookie` + unit tests) — the change this PR ships.
2. **Temporary**: playground + `next` dev dependency bumped to `16.3.0-canary.78` so CI demonstrates the fix against 16.3 (this CI run is the "green on 16.3" evidence). Includes the required `bfcacheId` router-mock type adaptation.
3. **Revert of 2** — the playground and packages keep `next@^16.2.2` (only the `e2e/partial-prefetching` app added later pins a canary, see above).
4. **`e2e/no-js`** test suite.
5. ESLint config cleanup (review feedback).
6. **`e2e/partial-prefetching`** test suite.

Local results on `16.3.0-canary.78` with the fix (production builds):

| Suite | Result |
| --- | --- |
| main | 72 passed, 1 skipped (pre-existing Radix skip) — the two regression tests pass 15/15 repeats |
| locale-prefix-never | 1 passed |
| trailing-slash | 7 passed |
| base-path | 3 passed, 1 skipped (see below) |
| domains | 2 passed |
| domains-locale-prefix | 29 passed |
| locale-cookie-false | 1 passed |
| e2e/no-js | 1 passed (also on `16.2.2`) |
| e2e/partial-prefetching | 1 passed (on `16.3.0-canary.91`) |
| jest, unit tests (240), `tsc` | all pass |

Without the fix, on the same canary: the cookie test failed 8/8 runs.

**Re-verified on `16.3.0-canary.91`** (latest canary at the time of writing): with the fix the cookie test passes; with `main`'s `syncCookie` it still fails (cookie `de` instead of `en`). The fix is not superseded by any upstream change.

### Separate upstream issue (not addressed here)

On 16.3 canary, switching back to the default locale with a `basePath` configured ends with the address bar at `/` instead of `/base/path`: the server responds correctly (`GET /base/path/en` → `307`, `location: /base/path`), but the router's redirect reconciliation commits the URL without re-adding the `basePath`. This is unrelated to cookie handling and can't be fixed in next-intl. Isolated reproduction: [nextjs-16-3-bug-repro-basepath](https://github.com/amannn/nextjs-16-3-bug-repro-basepath) (confirmed still broken on `16.3.0-canary.91`); expected upstream fix: [vercel/next.js#95608](https://github.com/vercel/next.js/pull/95608). The affected assertion is skipped with a note in the temporary commit 2 only (the skip is removed again by the revert, so the PR's final state contains no skips).

---
_Generated by [Claude Code](https://claude.ai/code/session_01EwJKaviKWaV6cayRiRQnqC)_


e2e/no-js/eslint.config.mjs
```diff
@@ -0,0 +1,20 @@
...
+  ...nextVitals,
+  ...nextTs,
+  {
+    rules: {
+      '@typescript-eslint/no-unused-vars': [
+        'error',
+        {
+          argsIgnorePattern: '^_',
+          caughtErrorsIgnorePattern: '^_',
+          varsIgnorePattern: '^_'
```
amannn: is this needed? can we remove the override?

PR #2356, opened 2026-07-09
fix: Correctly format plural/select branches that are only `#` or empty

spokodev:
### Bug

In `icu-minify/format`, a `plural` or `select` branch that compiles to a falsy value is wrongly rejected as a missing branch and throws in development:

- A branch that is only `#` compiles to `0` (`TYPE_POUND`), e.g. `{count, plural, one {one} other {#}}`.
- An empty branch `{}` compiles to `""`, e.g. `item{count, plural, one {} other {s}}`.

Both are valid ICU branches, but the guard uses `!branch`, which is truthy for `0` and `""`:

```
No matching branch for plural "count" with category "other"
```

The `?? options.other` selection is correct (nullish coalescing preserves `0`/`""`); only the subsequent dev guard misfires. In production the guard is skipped, so this is a development-only crash on valid messages, surfacing as a formatting error and message fallback.

### Fix

Check `branch === undefined` instead of `!branch`, so genuinely missing branches still throw while valid falsy branches format correctly. Same change in `formatSelect` and `formatPlural`.

### Tests

Adds regression tests for a bare-`#` branch and empty branches in both `plural` and `select`. They throw on `main` and pass with the fix.

### DIFF
--- a/packages/icu-minify/src/format.tsx
+++ b/packages/icu-minify/src/format.tsx
@@ -234,7 +234,8 @@ function formatSelect<RichTextElement>(
   const value = String(getValue(values, name));
   const branch: CompiledNode | undefined = options[value] ?? options.other;
 
-  if (process.env.NODE_ENV !== 'production' && !branch) {
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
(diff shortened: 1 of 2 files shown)


amannn (approved):
Looks great, thanks a lot!

PR #2361, opened 2026-07-21
fix: Resolve precompile format-only alias when config is bundled (e.g. `bun --bun next dev`)

amannn:
> [!IMPORTANT]
> **Blocked by https://github.com/vercel/next.js/issues/88540.** Once Turbopack supports package subpath exports as alias targets, the plugin can simply alias to the bare specifier `use-intl/format-message/format-only` and this entire class of problems disappears. Until then, affected users should apply the workaround below.

## Problem

Reported in [#2209](https://github.com/amannn/next-intl/discussions/2209#discussioncomment-17687273).

With `experimental.messages.precompile` enabled, running `bun --bun next dev` in a monorepo where `next-intl/plugin` is invoked through a shared wrapper package fails to start the dev server:

```
⨯ Failed to load next.config.ts
error: Cannot find module 'use-intl/format-message/format-only' from ''
```

## Why this problem exists

1. **The plugin must resolve a file path at config time.** With `precompile` enabled, messages are compiled to ASTs at build time, so the runtime should only ship the lightweight `format-only` formatter. The plugin achieves this by aliasing `use-intl/format-message` to the `format-only` build. Since Turbopack doesn't apply bare package subpaths as alias targets (vercel/next.js#88540), the alias target must be a **file path** — hence the plugin calls `require.resolve('use-intl/format-message/format-only')`, anchored to its own location via `createRequire(import.meta.url)`.

2. **A wrapper package causes the plugin to be inlined into the compiled config.** Next.js compiles `next.config.ts` before evaluating it. Packages in `node_modules` remain external, but **workspace packages are inlined** — and a wrapper around `next-intl/plugin` transitively pulls the plugin code into the compiled config. Inlined code loses its file identity: `import.meta.url` becomes `''`. This is why the issue only occurs with a wrapper package, and not when importing `next-intl/plugin` directly (a direct import stays external and keeps a valid `import.meta.url`).

## Why this only applies to Bun

Node.js evaluates the compiled config in a context where module resolution still works — `createRequire(import.meta.url).resolve(…)` returns the correct path (verified).

Under `bun --bun`, the compiled config is evaluated in a context where `createRequire` **ignores its filename argument entirely**: resolution runs `from ''` even when `createRequire` receives a valid absolute path, and even when the real built-in is obtained via `process.getBuiltinModule('module')` to rule out bundler shims (verified). This looks like a Bun bug in the module-evaluation context that Next.js uses for the compiled config — no resolution base that the plugin passes is honored. Only Bun's proprietary `Bun.resolveSync(id, parent)` API respects a parent in that context.

## Workaround for affected users

Import `next-intl/plugin` **directly** in each app's `next.config.ts` and share plain config options through your monorepo package instead of wrapping the plugin call:

```ts
// next.config.ts (in each app)
import createNextIntlPlugin from 'next-intl/plugin';
import i18nConfig from '@repo/i18n/config';

const withNextIntl = createNextIntlPlugin(i18nConfig);

export default withNextIntl({
  // …
});
```

```ts
// @repo/i18n — export plain options instead of wrapping the plugin
export default {
  experimental: {
    messages: {
      // …
      precompile: true
    }
  }
};
```

Verified working with `bun --bun next dev` (no library change required): the direct import keeps `next-intl` external during config compilation, so resolution behaves exactly like in a single-package setup. Alternatively, running `next dev` on Node.js (i.e. `bun next dev` without `--bun`) avoids the issue entirely — Bun remains the package manager, and Node.js is the officially supported runtime for Next.js.

## What this PR contains

- A fallback in the plugin that recovers resolution via `Bun.resolveSync` (from the app directory, or via `next-intl` for strict `node_modules` layouts) when `require.resolve` has no valid base, with the bare specifier as a last resort. Only the failure path is affected; behavior is unchanged whenever `import.meta.url` is available.
- An e2e reproduction (`e2e/bun-precompile-app` + `e2e/bun-precompile-config`) mirroring the reporter's setup, with a Playwright test that runs `next dev` on Bun and asserts the formatted output of a `Hello {name}` message — guarding both config loading and that the alias is actually applied (without it, precompiled messages fail to format at runtime).

Verified in all three states: unfixed plugin → dev server fails with the reported error; config-load-only fix (bare-specifier alias) → message fails to format; full fix → `Hello Jane!` renders.

https://claude.ai/code/session_01XMJB2oaFtqYr5zSqrcN5nx

### DIFF
--- a/e2e/bun-precompile-app/src/app/page.tsx
+++ b/e2e/bun-precompile-app/src/app/page.tsx
@@ -2,5 +2,5 @@ import {useTranslations} from 'next-intl';
 
 export default function Page() {
   const t = useTranslations();
-  return <h1>{t('Hello')}</h1>;
+  return <h1>{t('Hello', {name: 'Jane'})}</h1>;
 }
--- a/e2e/bun-precompile-app/src/app/page.tsx
+++ b/e2e/bun-precompile-app/src/app/page.tsx
@@ -1,0 +1,6 @@
+import {useTranslations} from 'next-intl';
+
+export default function Page() {
+  const t = useTranslations();
+  return <h1>{t('Hello')}</h1>;
+}
--- a/e2e/bun-precompile-app/test.mjs
+++ b/e2e/bun-precompile-app/test.mjs
@@ -1,0 +1,64 @@
+// Regression test for https://github.com/amannn/next-intl/discussions/2209
+//
+// When `experimental.messages.precompile` is enabled and `next-intl/plugin` is
+// used through a workspace wrapper package (see `next.config.ts`), running
+// `bun --bun next …` previously failed to load `next.config` with:
+//
+//   Cannot find module 'use-intl/format-message/format-only' from ''
+//
+// The plugin resolves the `format-only` build via `require.resolve`
+// (`createRequire(import.meta.url)`). When Next bundles the config before
+// evaluating it, `import.meta.url` is unavailable, so the resolution base is
+// empty and `require.resolve` throws. The fix falls back to the bare specifier
+// in that case.
+//
+// This test asserts that `next.config` *loads* under `bun --bun` (i.e. the
+// plugin no longer throws while resolving `format-only`). It deliberately does
+// not require the full build to finish: under `bun --bun`, Turbopack spawns a
+// Node subprocess for message-catalog loaders that rejects the inherited
+// `--bun`/`--smol` `NODE_OPTIONS`, which is unrelated to this regression.
+
+import {spawnSync} from 'node:child_process';
+import {createRequire} from 'node:module';
+
+const require = createRequire(import.meta.url);
+const REPORTED_ERROR =
+  "Cannot find module 'use-intl/format-message/format-only'";
+const CONFIG_LOAD_ERROR = 'Failed to load next.config';
+
+// The bug only reproduces on Bun's runtime (`--bun`), so skip when Bun is
+// unavailable (e.g. a plain Node environment) rather than failing.
+if (spawnSync('bun', ['--version'], {encoding: 'utf8'}).status !== 0) {
+  console.log('Skipping: `bun` is not available in this environment.');
+  process.exit(0);
+}
+
+const nextBin = require.resolve('next/dist/bin/next');
+
+console.log('Running `bun --bun next build` …\n');
+const result = spawnSync('bun', ['--bun', nextBin, 'build'], {
+  encoding: 'utf8',
+  stdio: 'pipe',
+  timeout: 300_000
+});
+
+if (result.error) {
+  console.error(`\n❌ Failed to run \`bun --bun next build\`: ${result.error}`);
+  process.exit(1);
+}
+
+const output = `${result.stdout ?? ''}${result.stderr ?? ''}`;
+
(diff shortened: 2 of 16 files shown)


e2e/bun-precompile-app/src/app/page.tsx
```diff
@@ -0,0 +1,6 @@
+import {useTranslations} from 'next-intl';
+
+export default function Page() {
+  const t = useTranslations();
+  return <h1>{t('Hello')}</h1>;
```
amannn: lets use a basic variable like `Hello {name}` here to ensure precompile actually works

e2e/bun-precompile-app/test.mjs
```diff
@@ -0,0 +1,64 @@
+// Regression test for https://github.com/amannn/next-intl/discussions/2209
```
amannn:
```suggestion
// Regression test for https://github.com/amannn/next-intl/discussions/2209#discussioncomment-17687273
```

e2e/bun-precompile-app/test.mjs
```diff
@@ -0,0 +1,64 @@
...
+// used through a workspace wrapper package (see `next.config.ts`), running
+// `bun --bun next …` previously failed to load `next.config` with:
+//
+//   Cannot find module 'use-intl/format-message/format-only' from ''
+//
+// The plugin resolves the `format-only` build via `require.resolve`
+// (`createRequire(import.meta.url)`). When Next bundles the config before
+// evaluating it, `import.meta.url` is unavailable, so the resolution base is
+// empty and `require.resolve` throws. The fix falls back to the bare specifier
+// in that case.
```
amannn:
```suggestion
// empty and `require.resolve` throws.
```

e2e/bun-precompile-app/test.mjs
```diff
@@ -0,0 +1,64 @@
...
+//
+// This test asserts that `next.config` *loads* under `bun --bun` (i.e. the
+// plugin no longer throws while resolving `format-only`). It deliberately does
+// not require the full build to finish: under `bun --bun`, Turbopack spawns a
+// Node subprocess for message-catalog loaders that rejects the inherited
+// `--bun`/`--smol` `NODE_OPTIONS`, which is unrelated to this regression.
```
amannn: i dont really understand, we're running a full build below—no?

amannn: instead of the current dev suite, can we just run next dev with bun and use playwright to verify correct output is there?

amannn:
Addressed the review comments in dc5b6a4 — and the Playwright suggestion turned out to be important: asserting the formatted output revealed that the previous bare-specifier fallback only fixed config *loading*, while the alias itself wasn't applied by Turbopack (bare package subpaths don't work as alias targets, vercel/next.js#88540), so precompiled messages failed to format at runtime.

Digging deeper showed that in the bundled-config environment under `bun --bun`, `createRequire` ignores its resolution base entirely (even with an explicit valid path), so the fix now recovers via `Bun.resolveSync` instead — from the app directory, or via `next-intl` for strict `node_modules` layouts. The test now runs `next dev` on Bun and asserts `Hello Jane!` renders (verified failing for both the unfixed plugin and the config-load-only fix). PR description is updated accordingly.

---

PR #2362, opened 2026-07-22
fix: Avoid `usePathname()` reads for non-locale-switching links, avoiding unnecessary `<Suspense>` wrappers with partial prefetching

ceolinwill:
Avoid calling `usePathname` for regular and same-locale links by isolating locale-cookie synchronization in an internal locale-changing link component. This prevents ordinary links from requiring Suspense with Cache Components.

Adds an E2E regression and raises the navigation bundle size limits by 25 B client-side and 40 B server-side for the component split.


amannn:
The fix looks good to me, thanks!

I guess you had to work around the setup in the existing partial-prefetching e2e app a bit, due to usage of `setRequestLocale`. I've updated it now to next/root-params which helps to simplify a bit here.

My changes are pushed, mind having a look if this looks good to you?

e2e/partial-prefetching/tests/main.spec.ts
```diff
@@ -38,3 +38,10 @@ it('supports navigation and locale switching', async ({page}) => {
   );
   expect(isSpa).toBe(true);
 });
+
+it('supports a link in the shell of a dynamic route', async ({page}) => {
+  await page.goto('/dynamic/example');
+  await expect(page.getByRole('heading', {name: 'Dynamic page'})).toBeVisible();
+  await page.getByRole('link', {name: 'Go to home page'}).click();
+  await expect(page.getByRole('heading', {name: 'Home'})).toBeVisible();
+});
```
amannn, line 47: Image

ceolinwill:
> I guess you had to work around the setup in the existing partial-prefetching e2e app a bit, due to usage of `setRequestLocale`

yeah. I thought about updating the existing example initially but I was not sure if I should update the existing fixtures because I know you have a draft PR in place already for root params. but, yeah, using root params there makes more sense

> mind having a look if this looks good to you

looks good to me

amannn (approved):

amannn:
npm failed halfway during the release: https://github.com/amannn/next-intl/actions/runs/30009041125/job/89212136114. Just a second, looking into this …

amannn:
Ok, should be out now!

Issue #2278, opened 2026-03-06
Message extraction misses translations when `t` is passed as a prop or function argument


alenjoy7:
### Description

When using the message extraction workflow, some translation keys are not detected if the `t` function is passed as a prop or used as a function argument.

Only the messages used directly in the component where the hook is defined or inside another component calling the hook itself are extracted.

### Code

```tsx
const TestComponent = ({ t }) => {
  return <>{t("from child prop")}</>;
};

const TestComponent2 = () => {
  const t = useExtracted();
  return <>{t("from child2")}</>;
};

const testFunction = (t) => {
  return t("from child function");
};

const Page = () => {
  const t = useExtracted();

  return (
    <>
      <p>{t("from parent")}</p>
      <TestComponent t={t} />
      {testFunction(t)}
      <TestComponent2 />
    </>
  );
};

export default Page;
```

### Extracted `en.json`

```json
{
  "Bk+y9q": "from child2",
  "hRD0WY": "from parent"
}
```

## Problem

The following messages are **not extracted**:

* `from child prop`
* `from child function`

## Expected behavior

All messages should be extracted:

* `from parent`
* `from child prop`
* `from child function`
* `from child2`

## Actual behavior

Only messages used directly with the hook (`useExtracted`) are extracted.

## Environment

* **Next.js version:** 16.1.6 (Next.js)
* **next-intl version:** 4.8.3
* **Node.js version:** 24.x (Node.js)

## Question

Is this a known limitation of the extraction system when the `t` function is passed as a prop or argument?
If so, is there a recommended pattern to ensure these messages are detected?

### Verifications

- [x] I've verified that the problem I'm experiencing isn't covered in [the docs](https://next-intl.dev).
- [x] I've searched for similar, existing issues on [GitHub](https://github.com/amannn/next-intl/issues) and [Stack Overflow](https://stackoverflow.com/search?q=next-intl).
- [x] I've compared my app to [a working example](https://next-intl.dev/examples) to look for differences.

### Mandatory reproduction URL

https://github.com/alenjoy7/next-intl-bug-repro-app-router

### Reproduction description

Steps to reproduce:
1. Open reproduction
2. Click on …
3. See error: …

### Expected behaviour

{
  "Bk+y9q": "from child2",
  "hRD0WY": "from parent",
  "someId":"from child function",
  "someId2":"from child prop"
}

abin-sunny:
Same problem

amannn:
[...]

PR #2367, opened 2026-07-29
docs: Add guidance for messages defined outside of components

amannn:
Documents the patterns that work today for messages that are defined outside of a component—the recurring question behind [#2278](https://github.com/amannn/next-intl/issues/2278), [this comment](https://github.com/amannn/next-intl/discussions/2036#discussioncomment-15110402), [this one](https://github.com/amannn/next-intl/discussions/2036#discussioncomment-15233911) and most recently [this one](https://github.com/amannn/next-intl/discussions/2036#discussioncomment-17814108).

### `useExtracted` → Static analysis

Two new expandable sections:

- **Enum-like values**: an async function retrieving `t` via `getExtracted` on the server (noting the resulting object of strings can be passed to a Client Component), or a hook that returns a function receiving the value on the client. Also points at `select` as an alternative.
- **Zod**: short pointer to the existing section in the Server Actions docs.

The key clarification in both: no wrapper component is needed. `t` has to be retrieved and used within the same function body, but that body can just as well be an async function or a hook—and it may return a closure.

### Server Actions → Zod

The existing example stays as-is: it keeps the schema locale-agnostic and allows tailoring the wording per field, which is often necessary in languages where phrasing depends on grammatical gender.

Adds a second example below it for sharing error messages across schemas. Note that Zod calls the error map synchronously, so the helper awaits the translations and *returns* the error map rather than being async itself.

Closes with a callout that both approaches apply to `useExtracted` by using `getExtracted` in place of `getTranslations`.

### Verification

- All three code shapes were run through `MessageExtractor` to confirm they actually extract (hook returning a closure with a switch, async helper with `getExtracted`, hook returning a record).
- `pnpm lint` in `docs` passes, including `next-validate-link` for the new anchors.
- Both pages checked in `next dev`.

### DIFF
--- a/docs/src/pages/docs/environments/actions-metadata-route-handlers.mdx
+++ b/docs/src/pages/docs/environments/actions-metadata-route-handlers.mdx
@@ -98,6 +98,50 @@ async function loginAction(data: FormData) {
 }
 ```
 
+This approach keeps the schema locale-agnostic and allows you to tailor the wording to each field—which is often necessary, e.g. in languages where the phrasing depends on the grammatical gender of the field name.
+
+If you'd instead like to share error messages across multiple schemas, you can move the mapping into a reusable helper. Note that Zod calls the error map synchronously, therefore the helper awaits the translations and returns the error map:
+
+```tsx filename="getValidationErrorMap.ts"
+import {getTranslations} from 'next-intl/server';
+
+export default async function getValidationErrorMap() {
+  const t = await getTranslations('Validation');
+
+  return function errorMap(issue) {
+    switch (issue.code) {
+      case 'too_small':
+        return t('tooSmall', {minimum: Number(issue.minimum)});
+      case 'invalid_format':
+        return t('invalidFormat');
+      default:
+        return undefined;
+    }
+  };
+}
+```
+
+```tsx
+async function loginAction(data: FormData) {
+  'use server';
+
+  const values = Object.fromEntries(data);
+  const result = loginFormSchema.safeParse(values, {
+    error: await getValidationErrorMap()
+  });
+
+  // ...
+}
+```
+
+<Callout>
+  Note that both approaches also work with
+  [`useExtracted`](/docs/usage/extraction) by using `getExtracted` in place of
+  `getTranslations`. Since `t` is retrieved and called within the same function
+  body in either case, the messages remain [statically
+  analyzable](/docs/usage/extraction#static-analysis).
+</Callout>
+
 </Details>
 
 ### Open Graph images
--- a/docs/src/pages/docs/environments/actions-metadata-route-handlers.mdx
+++ b/docs/src/pages/docs/environments/actions-metadata-route-handlers.mdx
@@ -98,9 +98,9 @@ async function loginAction(data: FormData) {
 }
 ```
 
-This approach keeps the schema locale-agnostic and allows you to tailor the wording to each field—which is often necessary, e.g. in languages where the phrasing depends on the grammatical gender of the field name.
+This approach keeps the schema locale-agnostic and allows you to tailor the wording to each field.
 
-If you'd instead like to share error messages across multiple schemas, you can move the mapping into a reusable helper. Note that Zod calls the error map synchronously, therefore the helper awaits the translations and returns the error map:
+If you'd instead like to share error messages across multiple schemas, you can move the mapping into a reusable helper:
 
 ```tsx filename="getValidationErrorMap.ts"
 import {getTranslations} from 'next-intl/server';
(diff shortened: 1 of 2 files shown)


docs/src/pages/docs/usage/extraction.mdx
```diff
@@ -346,6 +346,81 @@ async function AsyncExample() {
...
+Since a string is returned, the label can be used anywhere—not only in JSX, but also for e.g. `aria-label`, sorting or a toast message.
+
+Alternatively, if you prefer a single message, you can use [`select`](/docs/usage/translations#selecting-enum-based-values).
+
+</Details>
+
+<Details id="zod">
+<summary>How can I localize validation errors from Zod?</summary>
+
+[Zod](https://zod.dev/) schemas are typically defined at module scope, where `t` is not available. The recommended approach is to keep your schema locale-agnostic and to turn the structured errors it produces into messages in a place where `t` is in scope—either via a per-parse error map, or via a shared async helper.
```
amannn: newline after "not available."

docs/src/pages/docs/usage/extraction.mdx
```diff
@@ -346,6 +346,81 @@ async function AsyncExample() {
...
+
+See [When using Zod for validation, how can I localize error messages?](/docs/environments/actions-metadata-route-handlers#server-actions-zod) for examples of both approaches—they apply to `useExtracted` in the same way, by using `getExtracted` in place of `getTranslations`.
```
amannn:
```suggestion
See: [When using Zod for validation, how can I localize error messages?](/docs/environments/actions-metadata-route-handlers#server-actions-zod)
```

docs/src/pages/docs/environments/actions-metadata-route-handlers.mdx
```diff
@@ -98,6 +98,50 @@ async function loginAction(data: FormData) {
 }
 ```
 
+This approach keeps the schema locale-agnostic and allows you to tailor the wording to each field—which is often necessary, e.g. in languages where the phrasing depends on the grammatical gender of the field name.
```
amannn:
```suggestion
This approach keeps the schema locale-agnostic and allows you to tailor the wording to each field.
```

docs/src/pages/docs/environments/actions-metadata-route-handlers.mdx
```diff
@@ -98,6 +98,50 @@ async function loginAction(data: FormData) {
...
+}
+```
+
+<Callout>
+  Note that both approaches also work with
+  [`useExtracted`](/docs/usage/extraction) by using `getExtracted` in place of
+  `getTranslations`. Since `t` is retrieved and called within the same function
+  body in either case, the messages remain [statically
+  analyzable](/docs/usage/extraction#static-analysis).
+</Callout>
```
amannn: remove callout

docs/src/pages/docs/environments/actions-metadata-route-handlers.mdx
```diff
@@ -98,6 +98,50 @@ async function loginAction(data: FormData) {
 }
 ```
 
+This approach keeps the schema locale-agnostic and allows you to tailor the wording to each field—which is often necessary, e.g. in languages where the phrasing depends on the grammatical gender of the field name.
+
+If you'd instead like to share error messages across multiple schemas, you can move the mapping into a reusable helper. Note that Zod calls the error map synchronously, therefore the helper awaits the translations and returns the error map:
```
amannn:
```suggestion
If you'd instead like to share error messages across multiple schemas, you can move the mapping into a reusable helper:
```

docs/src/pages/docs/usage/extraction.mdx
amannn: Split into two paragraphs after "not available."

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

docs/src/pages/docs/usage/extraction.mdx
amannn: Applied — trimmed to the plain "See:" pointer.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

docs/src/pages/docs/environments/actions-metadata-route-handlers.mdx
amannn: Applied — dropped the grammatical gender aside.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

docs/src/pages/docs/environments/actions-metadata-route-handlers.mdx
amannn: Removed. The `Callout` import is still used elsewhere in the file, so it stays.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

docs/src/pages/docs/environments/actions-metadata-route-handlers.mdx
amannn: Applied — dropped the sentence about the error map being called synchronously.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #2375, opened 2026-08-05
docs: Link to Next.js `bind` docs in root params blog post

amannn:
In the `next/root-params` blog post, the caveat section mentions passing an explicit `locale` param at relevant call sites, but only showed the callee side. This adds a pointer to the Next.js docs for `bind`, which is the mechanism for getting that param in from a Server Action call site.

Link target: https://nextjs.org/docs/app/guides/forms#passing-additional-arguments (the current unversioned docs page that documents `bind`; `/docs/app/guides/server-actions` does not carry that section).

The `&nbsp;` between `e.g.`, `via` and the link keeps that cluster unbreakable — without it the link wrapped alone onto its own line at some viewport widths. Verified by measuring the rendered line boxes across viewport widths from 360–1600px: every wrap now leaves at least three tokens on the last line.

Lint passes (`pnpm lint` in `docs`: ESLint, Prettier, link check).

---
_Generated by [Claude Code](https://claude.ai/code/session_01NqMYt5JC8zRb1TVSzXF5Nb)_

### DIFF
--- a/docs/src/pages/blog/nextjs-root-params.mdx
+++ b/docs/src/pages/blog/nextjs-root-params.mdx
@@ -188,7 +188,7 @@ That's it—with a single change to `i18n/request.ts` you can start using `next/
 
 **One caveat though**: `next/root-params` currently doesn't work in Route Handlers or Server Actions.
 
-You can work around that though by passing an explicit `locale` param at relevant call sites:
+You can work around that though by passing an explicit `locale` param at relevant call sites, e.g.&nbsp;via&nbsp;[`bind`](https://nextjs.org/docs/app/guides/forms#passing-additional-arguments):
 
 ```tsx {3}
 async function action(locale: string) {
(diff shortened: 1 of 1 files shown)


docs/src/pages/blog/nextjs-root-params.mdx
```diff
@@ -188,7 +188,7 @@ That's it—with a single change to `i18n/request.ts` you can start using `next/
 
 **One caveat though**: `next/root-params` currently doesn't work in Route Handlers or Server Actions.
 
-You can work around that though by passing an explicit `locale` param at relevant call sites:
+You can work around that though by passing an explicit `locale` param at relevant call sites, e.g.&nbsp;via&nbsp;[`bind`](https://nextjs.org/docs/app/guides/forms#passing-additional-arguments):
```
amannn:
```suggestion
You can work around that though by passing an explicit `locale` param at call sites, e.g. via [`bind`](https://nextjs.org/docs/app/guides/forms#passing-additional-arguments):
```

PR #2382, opened 2026-08-12
feat: Support destructuring from `Promise.all` for `getExtracted` and `getTranslations`

amannn:
Supports the semantics of `Promise.all` for `getExtracted` and `getTranslations` in the SWC extractor plugin:

```tsx
const [t, nav] = await Promise.all([
  getExtracted(),
  getTranslations('Navigation')
]);
```

## Motivation

Devs (and especially coding agents following common skills/patterns) parallelize async work in Server Components via `Promise.all` and naturally include next-intl APIs in the array. While that's unnecessary for next-intl (calls are request-cached and typically resolve in <1ms), the pattern previously didn't just skip extraction — it broke at runtime: the `getExtracted` import is unconditionally rewritten to `getTranslations as getTranslations$1`, so an unrecognized `getExtracted()` call site was left dangling, producing a `ReferenceError` in compiled output.

## Implementation

- `visit_mut_var_declarator` now also recognizes `const [a, b] = await Promise.all([...])` with an inline array literal, mapping pattern elements positionally onto array elements and binding each pair exactly like a direct `const t = await getX()` initializer (extracted refactored into a shared `bind_translator` helper).
- Safe bail-outs where the static mapping doesn't hold:
  - Spread elements in the array literal (`Promise.all([...calls, getExtracted()])`) skip the whole call, since positions no longer line up.
  - Holes, rest/nested patterns, and unrelated array elements are skipped individually.
- `Promise.all` is matched by name; a local binding shadowing the `Promise` global is not detected (consistent with how similar tools handle globals).
- Out of scope (unchanged behavior): `Promise.allSettled` (different result shape), passing a variable to `Promise.all`, or awaiting a stored promise later.

No runtime changes were needed — the async APIs already compose with `Promise.all` at runtime; only static analysis was missing.

## Tests

- New plugin fixture `async-promise-all` covering `getExtracted` + namespaced `getTranslations` + an unrelated call in one destructuring (`cargo test`: 34 passed).
- New end-to-end `MessageExtractor` test through the rebuilt wasm binary (vitest: 895 passed).

## Docs

`extraction.mdx`: moved the `Promise.all` example from the "not supported" list to a supported example, keeping the note that parallelizing these calls isn't necessary for performance.

---
_Generated by [Claude Code](https://claude.ai/code/session_01F5XVppeuLSbiabAHpEWsBf)_


docs/src/pages/docs/usage/extraction.mdx
```diff
@@ -316,6 +316,28 @@ function Example({name}) {
...
+    getTranslations('Navigation'),
+    getExtracted('Dashboard')
+  ]);
+
+  // Note: `getExtracted` is cached internally and even
+  // the first invocation typically takes less than 1ms.
+  // Therefore, there's no performance need to parallelize
+  // it—but it works if you prefer this pattern.
+}
+```
```
amannn:
```suggestion
Despite not being needed for performance reasons, `Promise.all` is also supported:

```tsx
import {getExtracted} from 'next-intl/server';
import {getPost} from '@/lib';

async function AsyncExample() {
  // ✅ Destructuring from `Promise.all` is supported
  const [t, post] = await Promise.all([
    getExtracted(),
    getPost()
  ]);

  // ...
}
```

```

(github will mess formatting up here, commit with correct syntax)

Issue #2087, opened 2025-11-10
Stabilize `useExtracted`


amannn:
With [`next-intl@4.5`](https://next-intl.dev/blog/use-extracted), there's now experimental support for `useExtracted`.

The next phase will be to collect [feedback](https://github.com/amannn/next-intl/discussions/2036), potentially incorporate changes and fill in the gaps for missing features and docs.

**TODO**
- [ ] Custom formatters
- [ ] Open-source PO parser so custom po parser is easy to build ([ref](https://github.com/amannn/next-intl/discussions/2036#discussioncomment-14905569))
- [ ] Rename "formatter" to something like "adapter" to avoid a naming conflict with `useFormatter`
- [ ] Docs/demo for monorepos ([ref](https://x.com/jamannnnnn/status/1987163883165860145))
  - [ ] Not shared messages
  - [ ] Shared messages (what if consumers build messages themselves after running extraction script?)
- [ ] Check / fix source maps (see also [magic-string](https://github.com/Rich-Harris/magic-string))
- [ ] More docs for array-esque data ([ref](https://bsky.app/profile/alexdln.com/post/3m575lopx4c2c))
- [ ] `../ui` should likely not look into `node_modules` (check what Tailwind does)
- [ ] Double check PPR works for .po loader

This list acts as my personal todo list that I want to share for transparency, but please be sure to comment here if you have feedback for the feature:

**→ https://github.com/amannn/next-intl/discussions/2036**

zigang93:
@amannn How does the new useExtracted API work with https://github.com/lokalise/i18n-ally?
I’m kind of missing the feature where we’re able to preview all locales or auto-translate directly from the VS Code plugin.

we love the useExtracted API and it improved our DX

amannn:
@zigang93 In the `useExtracted` docs for [`formats`](https://next-intl.dev/docs/usage/extraction#formats) you can find links to tools which might be helpful. i18n Ally can still be helpful if you use .json, but for .po you need another tool (e.g. Poedit).

That being said, I'm thinking about providing such a plugin in the future myself with extended support for common workflows with `next-intl`!

Side note: Let's keep the discussion over in https://github.com/amannn/next-intl/discussions/2036

kasp416h:
I might just be confused,
but normally when I use translations in SSR, I use getTranslations in the page.tsx and pass the t() function to the server components children.
But when I try this with getExtracted, it does not automatically extract the messages in the server components children? Same with useExtracted.
[...]

PR #2386, opened 2026-08-16
feat: Split extracted catalogs by namespace

martijn00:
Supports splitting extracted catalogs by namespace so large `useExtracted` apps can keep one file per feature instead of one monolith per locale:

```ts
extract: {
  split: 'namespace'
}
```

```
messages/
  en.po
  de.po
  ui/en.po
  ui/de.po
```

## Motivation

Extraction always writes one catalog per locale (`messages/en.po`). Namespaces from `useExtracted('ui')` are only a logical grouping inside that file (`msgctxt` / nested JSON). That works until the catalog grows to thousands of messages: diffs, merge conflicts, and translation tools all operate on a single huge file.

`messages.path` as an array already supports **per-package** catalogs in a monorepo, but not **per-namespace** files within one app. Manual multi-file loading is documented for `useTranslations` JSON, but the extractor will not maintain those files.

This is opt-in and format-agnostic (`.po` and `.json`). Default behavior is unchanged.

## Implementation

- `extract.split: 'namespace'` on the object form of `extract` (not on `extract: true`).
- Partition key is the first `.` segment of the **combined message id** (`login.saas-login.form.x` → `login/`). The extractor only sees the combined id, so it cannot recover the original `useExtracted('login')` argument vs a dotted explicit id.
- Full ids stay inside each file. The filename is only a partition — the catalog loader still infers locale from the basename (`en.po` → `en`), so `{namespace}/{locale}.po` needs no loader change. `{locale}/{namespace}.po` would break that.
- Unnamespaced ids, and unsafe first segments (`/`, `..`, `__proto__`, …), stay in the root `{locale}` file. That root file is always written (header-only if needed) so locale inference still has a locale marker when every message is namespaced.
- `CatalogPersister` reads root + `{ns}/{locale}` files (namespace files win on id overlap), writes the partition, and deletes stale namespace files. `getLastModified` uses the max mtime across those files so translator edits to a single namespace file are not missed.
- `CatalogLocales` infers locales from file basenames in the root **and** one level of `{ns}/*{ext}` (Crowdin can create `ui/fr.po` before `fr.po` exists) and watches recursively when split is on.
- Enabling split against an existing monolith: the next save moves namespaced entries out and preserves translations. Runtime loading is still the app's job — merge the root file with each namespace file in `getRequestConfig` (shallow spread; each file owns a distinct top-level key).

Out of scope: tree-shaking / client bundle size, `{locale}/{namespace}.po`, stripping the namespace from ids because the filename is the namespace, a `loadCatalogs()` helper, and configurable split depth.

## Tests

- `getCatalogNamespace` unit tests for first-segment partition and unsafe fallbacks.
- `ExtractionCompiler` coverage for JSON and PO: write layout, target-locale translations, monolith migration, stale namespace delete, inferring a locale from only `ui/fr.po`, and the default single-file path when `split` is unset.
- New `e2e/extracted-po-split` app: root + `ui` catalogs, merge in `getRequestConfig`, and extraction of a new namespaced message into `ui/en.po`.

## Docs

`plugin.mdx`: `extract.split`. `extraction.mdx`: file layout and `getRequestConfig` merge example.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## Unreleased
+
+### Features
+
+* Split extracted catalogs by namespace (`extract.split: 'namespace'`)
+
 ## 4.13.6 (2026-08-10)
 
 ### Bug Fixes
--- a/docs/src/pages/docs/usage/extraction.mdx
+++ b/docs/src/pages/docs/usage/extraction.mdx
@@ -189,7 +189,7 @@ If you want to use an explicit ID instead of the auto-generated one, you can opt
 
 This can be useful when you have a label that is used in multiple places, but should have different translations in other languages. This is an escape hatch that should rarely be necessary.
 
-### Namespaces
+### Namespaces [#namespaces]
 
 If you want to organize your messages under a specific namespace, you can pass it to `useExtracted`:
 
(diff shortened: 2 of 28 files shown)


amannn:
Hey, thank you for proposing this feature!

I think this feature will come at some point, yes—but likely with a slightly different config.

I've recently worked on [eloqnt/cli](https://studio.eloqnt.dev/docs), which is a CLI companion for `next-intl`. It uses the same configuration that also `next-intl` uses for `useExtracted`, but as of recently also supports placeholders like `{locale}` and `{namespace}` in [`messages.path`](https://studio.eloqnt.dev/docs/configuration#messages-path).

I'm currently using eloqnt/cli as a testing ground to future-proof this configuration (as long as we're still pre 1.0). At some stage, I think we'll be able to support this for `next-intl` too, but I'd like to wait for now while we're gaining confidence.

> That works until the catalog grows to thousands of messages: diffs, merge conflicts, and translation tools all operate on a single huge file.

`bluesky-social/social-app` is an example of a single-file, huge catalog btw.: https://github.com/bluesky-social/social-app/blob/main/src/locale/locales/en/messages.po

---

I made a note of this in https://github.com/amannn/next-intl/issues/2087

PR #2387, opened 2026-08-16
feat: Allow defining extracted messages outside of components

### LINKED ISSUE #2368: Defining messages outside of components
## Summary

A recurring request for [`useExtracted`](https://next-intl.dev/docs/usage/extraction) is a way to _define_ a message outside of a component and _resolve_ it later—in the spirit of Lingui's [`msg`](https://lingui.dev/ref/macro#definemessage) or FormatJS's [`defineMessage`](https://formatjs.github.io/docs/react-intl/api#definemessages):

```tsx
// Module scope: compiles to a descriptor instead of being formatted in place
const label = msg('Pending');

// Later, in a component: values are supplied at render time
t(label);
```

**Current stance: this is not supported by design.** The reasoning is below, along with the patterns that work today.

That said, I'd like to use this issue to collect further input. If you have a use case that isn't covered by the existing patterns, please share it in the comments.

## Use cases

Two variations of the same underlying need have come up so far:

1. **Enum-like label maps** — mapping a status, role or category to a label. These naturally want to be module-scope constants.
2. **Validation schemas** — [Zod](https://zod.dev/) schemas define messages at schema-definition time. The message travels through the validation result as _data_ and is rendered later in a shared component.

## Why it's not supported

The reason is tree-shaking, tied to [#1](https://github.com/amannn/next-intl/issues/1).

Today, a message is only ever referenced from the function body that retrieves `t`. That's what makes it possible to statically determine which module graph a given message belongs to—and therefore which messages a given route actually needs.

If messages can be defined in one module and resolved in another, that link is broken. A descriptor can be exported, re-exported, stored in a map, or passed through a third-party API, and there's no reliable way to trace which routes end up depending on it. The result is that we can't safely drop unused messages from a bundle.

One possible mitigation would be to de-opt when a namespace is used, i.e. include all messages of that namespace wherever it's referenced. But depending on how namespaces are structured, this could accidentally drag in a large number of unrelated messages, so it's a partial mitigation at best.

See also: [RFC 001 — Macro for defining messages](https://github.com/amannn/next-intl/blob/main/rfcs/001-message-extraction.md#macro-for-defining-messages)

## What works today

The important clarification is that the rule is **not** "`t` must be used in a component". The rule is that `t` must be retrieved and used within the same function body ([static analysis](https://next-intl.dev/docs/usage/extraction#static-analysis)).

That function body can be:

- an **async function** on the server, retrieving `t` via `getExtracted`
- a **hook** on the client, which may return a closure that receives the value to translate

Both are documented here:

- [How can I translate enum-like values?](https://next-intl.dev/docs/usage/extraction#enum-values)
- [When using Zod for validation, how can I localize error messages?](https://next-intl.dev/docs/environments/actions-metadata-route-handlers#server-actions-zod)

And of course, [`useTranslations`](https://next-intl.dev/docs/usage/translations) with a hand-maintained catalog remains fully supported and can be used alongside `useExtracted` for cases where it fits better.

## Acknowledgement

Finding abstractions that compose well within this paradigm does take some work, and I recognize this as a genuine limitation of the current approach. The patterns above cover the cases reported so far, but if you hit one they don't, please add it below.

## Previous discussions

- [Passing `t` into a Zod schema factory](https://github.com/amannn/next-intl/discussions/2036#discussioncomment-15110402) ([answer](https://github.com/amannn/next-intl/discussions/2036#discussioncomment-15123660))
- [Enum-like translations with typed constants](https://github.com/amannn/next-intl/discussions/2036#discussioncomment-15233911) ([answer](https://github.com/amannn/next-intl/discussions/2036#discussioncomment-15239754))
- [#2278 — Message extraction misses translations when `t` is passed as a prop or function argument](https://github.com/amannn/next-intl/issues/2278)
- [Passing `t` through function arguments breaks static analysis](https://github.com/amannn/next-intl/discussions/2036#discussioncomment-16466940)
- [#2246 — How to use it in non-page settings?](https://github.com/amannn/next-intl/discussions/2246) (see also [#2295](https://github.com/amannn/next-intl/pull/2295#issuecomment-4334474931))
- [Messages outside components — a `msg()`-style marker?](https://github.com/amannn/next-intl/discussions/2036#discussioncomment-17814108)
- https://github.com/amannn/next-intl/pull/2387

martijn00:
Closes #2368

Adds `msg()` / `defineMessage()` so `useExtracted` messages can be defined at module scope and resolved later with `t()` — the Lingui/FormatJS pattern requested for enum-like label maps and Zod schemas.

```tsx
import {msg, useExtracted} from 'next-intl';

const pending = msg('Pending');
const passwordMin = msg(
  'Password must be at least {length, plural, one {# character} other {# characters}} long'
);

function Status({name}: {name: string}) {
  const t = useExtracted();
  return (
    <>
      {t(pending)}
      {t(passwordMin, {length: 8})}
    </>
  );
}
```

## Motivation

`t` must stay in the same function body as `useExtracted` / `getExtracted` so extraction (and later per-route splitting, #1) can see every inline message. That rule makes two common cases awkward:

1. **Enum-like label maps** want to be module-scope constants, reusable as plain strings (`aria-label`, toasts, sorting) — not only as a label-only component whose job is holding `t`.
2. **Zod schemas** define field-specific messages at construction time. The string travels through the validation result as data and is rendered later. A shared `issue.code` map is not enough when gendered languages need per-field wording, and ICU values like `{length}` only exist at render time.

Passing `t` into a schema factory looks like it works via the uncompiled fallback and silently omits the message from the catalog (#2278).

## Implementation

- `msg` / `defineMessage` in `use-intl` core return a branded string. Uncompiled (tests/Storybook) that string is the source message; after the SWC plugin it is the hashed catalog key (`eKEL_g`, or `validation.Seanpx` when `namespace` is set). Zod can store it as `issue.message: string`.
- Values are supplied only at `t` / `t.rich` / `t.markup` — not on `msg()`. Tagged templates were rejected (RFC 001 + the Zod ICU case).
- The SWC plugin extracts `msg(...)` at any scope, rewrites the call to the full key literal, and no longer errors when `t` receives an identifier, member, optional chain, or `msg()` call. Concatenation, interpolated templates, and `t(getMessage())` still fail static analysis.
- Inline `t('…')` is unchanged: still same-function-body, still literal-only.
- Discovery: `MessageExtractor` and the Turbopack `content` regex also match `msg(` / `defineMessage`.
- Namespaced `msg({message, namespace})` keys must be resolved with root `useExtracted()` so `createBaseTranslator` can walk the dotted path. File-level `reference.path` is the tree-shaking attribution for #1 (unused `msg()` in the same file stay in the catalog).

## Tests

- New SWC fixtures: `msg-basic`, `msg-object`, `msg-define-message`, `msg-alias`, `msg-resolve`, `msg-inline-t` (`cargo test --test fixture`: 34 passed).
- `MessageExtractor`: `msg()`-only files are not skipped; `t(pending)` does not re-extract; plain `t(message)` is pass-through.
- `ExtractionCompiler`: `msg()` messages persist like `t()` messages.
- `useExtracted` / `msg` unit tests and a `getExtracted` type test for ICU args on descriptors.
- `e2e/extracted-json`: temp `msg()` labels + `t(descriptor, {length})` assert catalog + rendered output.

## Docs

`extraction.mdx`: `msg()` section, updated static-analysis rules, enum and Zod details. `actions-metadata-route-handlers.mdx`: field-specific Zod example. RFC 001: tagged templates stay rejected; function-form `msg()` is the supported de-opt.

Made with [Cursor](https://cursor.com)


sss-Mihail-sss:
@amannn Can you look at this please? :innocent:

amannn:
Hey, thank you for looking into this!

As you might have seen in https://github.com/amannn/next-intl/issues/2368, this is currently not supported by design. The docs contain patterns that help to address the common use cases with the current API surface. Increasing the public API is something I only do with great consideration, I hope you can understand!

If you have a use case that is not supported yet, please consider commenting on the open issue—thanks!

sss-Mihail-sss:
any updates?

PR #2393, opened 2026-08-26
feat: Switch `msgid` of built-in `.po` codec to source string for `useExtracted`

amannn:
## What has changed

PO catalogs now follow the gettext convention: the source text is stored in `msgid`, while the message key moved to `msgctxt`. Translations in target catalogs appear right next to their source text.

```diff
 #. Advance to the next slide
 #: src/components/Carousel.tsx
-msgid "5VpL9Z"
-msgstr "Rechts"
+msgctxt "5VpL9Z"
+msgid "Right"
+msgstr "Rechts"
```

This makes catalogs readable everywhere the PO ecosystem expects the source text in `msgid`: diffs on GitHub can be reviewed without looking up keys, and PO editors and translation management systems work out of the box.

## How to migrate

Migrating existing PO catalogs is mandatory: catalogs in the previous layout stop the build with an error that links to this page.

Your options:

1. **Migrate to the new structure** (recommended). You can use an agent with the following prompt:

   <details>
   <summary>Migration prompt</summary>

   > In this project, convert every `.po` messages catalog from the layout where `msgid` holds a message key and `msgstr` holds the text to the gettext-style layout. For each entry in each catalog: set `msgctxt` to the full key (if the file's entries have a `msgctxt` namespace, join it with the `msgid` using a dot), set `msgid` to the entry's source text (look it up by key in the source-locale catalog's `msgstr`; in the source-locale catalog itself it's the entry's own `msgstr`), and keep `msgstr` as it is. In the header block of each file, remove the `"X-Crowdin-SourceKey: msgstr\n"` line and add `"X-Message-Key: msgctxt\n"`. Preserve entry order, comments, references and flags. Verify afterwards by running a build: extraction must not modify the converted files further. Details on the layout change: `https://github.com/amannn/next-intl/pull/2393`

   </details>

   If you sync your catalogs with a translation platform, update the sources there after migrating.

2. **Keep the previous structure** by configuring [`POCodecLegacy`](https://github.com/amannn/next-intl/blob/f96607320dfae1a01e778297a18a906c8d6e2ccc/packages/next-intl/src/extractor/format/codecs/fixtures/POCodecLegacy.tsx) as a [custom format](https://next-intl.dev/docs/usage/plugin#formats-custom)

## Other changes

- Custom codecs can now be referenced by the name of an installed package (e.g. `codec: '@eloqnt/format-po'`), not only by file path.
- JSON catalogs: `null` values no longer crash extraction, boolean and number values are preserved instead of being silently dropped, and keys containing literal dots are no longer re-nested.


packages/next-intl/src/extractor/format/codecs/fixtures/POCodecLegacy.tsx
```diff
@@ -0,0 +1,35 @@
+import createPoCodec from '@eloqnt/format-po';
+import POParser from 'po-parser';
+import {defineCodec} from '../../ExtractorCodec.js';
```
amannn: we should import from eloqnt/config here

packages/next-intl/src/extractor/format/codecs/BuiltInPoCodec.tsx
```diff
@@ -0,0 +1,26 @@
+import createPoCodec from '@eloqnt/format-po';
+import {defineCodec} from '../ExtractorCodec.js';
+
+const LEGACY_HEADER = 'X-Crowdin-SourceKey: msgstr';
+
+const MIGRATION_ERROR = `This catalog uses the previous PO layout of next-intl, which stored message keys in \`msgid\`. The PO format now follows the gettext convention: the source text in \`msgid\` and the message key in \`msgctxt\`. Please migrate your catalogs, or use a custom codec to keep the previous layout: https://github.com/amannn/next-intl/pull/2393`;
```
amannn: these are only used once, inline below

packages/next-intl/src/extractor/ExtractionCompiler.test.tsx
```diff
@@ -1253,26 +1338,22 @@ msgstr ""
...
       "Language: de\\n"
-      "Content-Type: text/plain; charset=utf-8\\n"
-      "Content-Transfer-Encoding: 8bit\\n"
-      "X-Generator: next-intl\\n"
-      "X-Crowdin-SourceKey: msgstr\\n"
 
       #: src/Greeting.tsx
       #, no-wrap
-      msgid "-YJVTi"
+      msgid "Hey!"
```
amannn, line 1352: whats going on here in this test? isn't there a msgctxt missing with the id? am i missing smth?

packages/next-intl/src/extractor/ExtractionCompiler.test.tsx
```diff
@@ -1699,10 +1762,6 @@ msgstr "Hallo!"`
           "msgid ""
         msgstr ""
         "Language: en\\n"
-        "Content-Type: text/plain; charset=utf-8\\n"
-        "Content-Transfer-Encoding: 8bit\\n"
-        "X-Generator: next-intl\\n"
-        "X-Crowdin-SourceKey: msgstr\\n"
```
amannn, line 1705: why is this header gone?

packages/next-intl/package.json
amannn: since typescript is no longer a peer dependency, i guess this entry can be removed?

PR #2395, opened 2026-08-27
fix: Unbreak `experimental.messages` without a `messages.sourceLocale`

### LINKED ISSUE #2394: Turbopack build fails: catalogLoader options not serializable when sourceLocale is unset
### Description

Running `next build` with Turbopack fails when `experimental.messages` is configured in the `next-intl` plugin config *without* explicitly setting `sourceLocale`:

```
Error: loader next-intl/extractor/catalogLoader for match "*.json" does not have serializable options. Ensure that options passed are plain JavaScript objects and values.
```

**Root cause:** In [`getCatalogLoaderConfig()`](https://github.com/amannn/next-intl/blob/main/packages/next-intl/src/plugin/getNextConfig.tsx#L112-L132), `sourceLocale` is always spread into the loader's `options.messages` object, even when it resolves to `undefined`:

```ts
const sourceLocale = (messages.sourceLocale ??
  (typeof experimental.extract === 'object'
    ? experimental.extract.sourceLocale
    : undefined))!;
return {
  loader: 'next-intl/extractor/catalogLoader',
  options: {
    messages: {
      format: messages.format,
      ...(messages.precompile !== undefined && {
        precompile: messages.precompile
      }),
      sourceLocale // <- always included, even if undefined
    }
  } satisfies CatalogLoaderConfig as TurbopackLoaderOptions
};
```

Next.js validates Turbopack loader options with a JSON round-trip deep-equal check (`isDeepStrictEqual(options, JSON.parse(JSON.stringify(options)))`). Since `JSON.stringify` drops object keys whose value is `undefined`, an object containing `{ sourceLocale: undefined }` never equals its own JSON round-trip, so the check always fails — even though the options are otherwise a plain, serializable object.

`precompile` a few lines above is already handled correctly by conditionally spreading it only `!== undefined`. Applying the same pattern to `sourceLocale` fixes the issue:

```ts
options: {
  messages: {
    format: messages.format,
    ...(messages.precompile !== undefined && {
      precompile: messages.precompile
    }),
    ...(sourceLocale !== undefined && {
      sourceLocale
    })
  }
} satisfies CatalogLoaderConfig as TurbopackLoaderOptions
```

I verified this fix locally by patching `node_modules/next-intl` (via `bun patch`) across the `esm/development`, `esm/production`, and `cjs/development` builds, and it resolves the build error without any other behavior change.

### Verifications

https://github.com/melkir/next-intl-bug-repro-app-router

### Reproduction description

Steps to reproduce:
1. Clone the reproduction repo above (based on the official `next-intl-bug-repro-app-router` template, bumped to `next@16.3.3` / `next-intl@4.14.0`, with `experimental.messages` configured in `next.config.mjs` without a `sourceLocale`)
2. `bun install`
3. `bunx next build`
4. See error: `Error: loader next-intl/extractor/catalogLoader for match "*.json" does not have serializable options.`

### Expected behaviour

`next build` should succeed. `sourceLocale` should be optional in `experimental.messages` the same way `precompile` already is, without breaking Turbopack's loader-options serializability check.

amannn:
Fixes https://github.com/amannn/next-intl/issues/2394


packages/next-intl/src/plugin/catalog/catalogLoader.tsx
```diff
@@ -69,7 +69,10 @@ export default function catalogLoader(
       const locale = path.basename(this.resourcePath, extension);
       const decoded = codec.decode(source, {
         locale,
-        sourceLocale: options.messages.sourceLocale
+        // A source locale is only enforced when extraction is enabled, therefore
+        // this can be absent when catalogs are merely loaded. Codecs compare
+        // this to `locale`, hence an empty string matches no catalog.
+        sourceLocale: options.messages.sourceLocale ?? ''
```
amannn: this empty string looks bad. what exactly are the use cases for sourceLocale in the codec? should we make it optional at codec level? or should we just always require it now since it's mandatory for codecs?

amannn: Agreed, the empty string is gone — I went with requiring it (5ef21bb).

**What codecs use it for:** in `decode`, exactly one thing. `@eloqnt/format-po` computes `layout !== 'id-msgid' && msgstr === '' && locale === sourceLocale` to decide whether a message falls back to `msgid` (and marks `formatPoEmptySource` so `encode` writes the empty `msgstr` back). `@eloqnt/format-json` doesn't read it at all. Nothing else in our decode path touches it — `CatalogPersister` always has a real one from `normalizeExtractorConfig`, so `catalogLoader` was the only caller that could lack it.

**Why require rather than make it optional at codec level:** it isn't only cosmetic. For PO, a messages-only config without a source locale decodes untranslated entries in the source catalog as empty strings instead of falling back to `msgid` — silently wrong output rather than a missing niceness. Requiring it also keeps the codec contract simple (no third-party codec has to handle an absent source locale), and it's consistent with `messages.locales`, which is already required in the same object.

**What changed:**
- `initPlugin` validates a source locale whenever `experimental.messages` is configured, not just when `extract` is enabled. The `else if` keeps `normalizeExtractorConfig` as the single place that warns about the deprecated `extract.sourceLocale`.
- `CatalogLoaderConfig['messages']['sourceLocale']` is required again and `catalogLoader` passes it straight through.
- `pruneUndefined` stays — it guards the whole class of non-serializable loader options rather than this one key.
- Docs: the `messages` example was missing both `locales` and `sourceLocale`, and `messages.sourceLocale` now notes that it matters for loading too.

Verified against the config from #2394: it now throws ``[next-intl] `messages.sourceLocale` is required when loading messages.`` instead of the Turbopack serialization error, and with a source locale the options survive the JSON round-trip. Every `e2e`/`examples` config that uses `messages` already sets it, so nothing in the repo needs updating.

Worth noting this changes what the fix means for the reporter: their build won't start working on upgrade, they'll get the descriptive error and add one line. If you'd rather unbreak them without a config change, the alternative is the codec-level route — `sourceLocale?: Locale` in `@eloqnt/config` — which needs an eloqnt release and pushes the absent case onto every codec.

---

amannn: Following up on this after digging into where the behavior came from (eloqnt#329) and how Lingui handles it: the config-level requirement is reverted (41372ea), and this PR is back to just unblocking current users.

The empty string stays, but as a deliberate "unknown" sentinel rather than a shrug — it reads as "no catalog is the source one", which restores 4.13 behavior and is unambiguous for JSON. The real fix belongs in the PO codec, which can detect it and error when it encounters an empty `msgstr` it cannot attribute. That keeps JSON and other formats unaffected, since they never read the value, and it's tracked separately against eloqnt.

For reference, Lingui doesn't use a source locale for this on the read side at all: `@lingui/format-po`'s `parse(content)` ignores the ctx and keeps `message` (from `msgid`) and `translation` (from `msgstr`) as separate fields, resolving the fallback later at compile time. It only uses `sourceLocale` on serialize, for explicit IDs. eloqnt collapses both into one `message` field at decode time, which is why the decision has to happen there.

---

PR #2400, opened 2026-09-02
fix: improved the isPromise promise validation logic

vedanshshetti:
I have improved the existing validation logic for the isPromise utility from checking for a `.then` to use a safer, more modern approach via `Promise.resolve()`

### DIFF
--- a/packages/next-intl/src/shared/utils.tsx
+++ b/packages/next-intl/src/shared/utils.tsx
@@ -201,5 +201,5 @@ export function isPromise<Value>(
   value: Value | Promise<Value>
 ): value is Promise<Value> {
   // https://github.com/amannn/next-intl/issues/1711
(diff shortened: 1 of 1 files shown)


packages/next-intl/src/shared/utils.tsx
```diff
@@ -201,5 +201,5 @@ export function isPromise<Value>(
   value: Value | Promise<Value>
 ): value is Promise<Value> {
   // https://github.com/amannn/next-intl/issues/1711
-  return typeof (value as any).then === 'function';
+  return Promise.resolve(value as any) === value;
```
amannn, line 204: Hey, thanks for having a look here, but the way this works currently is correct and important to keep like this (see the linked issue in the comment above and https://github.com/amannn/next-intl/pull/1719)

PR #2401, opened 2026-09-02
fix: graceful error handling for invalid pathnames

### LINKED ISSUE #2399: Next-Intl silently fails when locale is an unsupported URI character
### Description

When locale is "en" or "fr" it works, as expected.
However, **when given a non-URL safe character**, it doesn't give a clear error for certain chars:
- a locale named "²" will silently throw a 404 without explaination as standard browsers and Embedded Chromium Instances always URI-encode "²" to "%C2%B2", which does not match expected locale of "²".
- same goes for backticks (the \` character), which gets encoded -> "%60", not matching "\`"
- and using "%" will return a 400 Bad Request error by Next.js Router, as % is used strictly as an escape character only and mandates two hexadecimal chars following it.
- and using an emoji is even worse, as it breaks ALL routes, not just itself.

I'm currently working on a fix, I'll open a PR in an hour or so.

> Note: I have tested this on a App-Router template, I did not test this yet on a Pages-Router application, as the router system does not matter for this specific bug.

### Verifications

https://github.com/vedanshshetti/next-i18n-bugconcern

### Reproduction description

Steps to reproduce:
1. Open reproduction
2. Open the terminal
3. Run `npm run dev`
4. Open `http://localhost:3000/` or `http://127.0.0.1:3000/` followed by one of the characters listed above
6. See error: 400 Bad Request (for /%) and a silent 404 (blank page) for the others

To test out the Emoji Problem, please first ADD an emoji to `locales: [...]` in `src/app/i18n/routing.js`

### Expected behaviour

I expected next-intl to throw either a clean, "Unsupported character detected in locale" or perhaps throw an error directly in the IDE.

> Quick Note: I'm currently working on a fix, I'll open a PR once I finish, just give me a few hours.

vedanshshetti:
# Fixes #2399.

Adds safe handling for malformed percent‑encoded pathnames in middleware.
When decodeURI fails, next‑intl now returns a clear 400 error page instead of forwarding to Next.js or crashing.

Also extracts the locale safely using decodeURIComponent with fallback.

> IMPORTANT NOTE: I haven't tested this yet, please test it yourself if you're unsure about the code.

### DIFF
--- a/packages/next-intl/src/middleware/middleware.tsx
+++ b/packages/next-intl/src/middleware/middleware.tsx
@@ -46,13 +46,35 @@ export default function createMiddleware<
 
   return function middleware(request: NextRequest) {
     let unsafeExternalPathname: string;
+    let extractedLocale: string;
+
+    try {
+      extractedLocale = decodeURIComponent(request.nextUrl.pathname.split("/")[1] as string);
+    } catch {
+      extractedLocale = request.nextUrl.pathname.split("/")[1] as string;
+    }
     try {
       // Resolve potential foreign symbols (e.g. /ja/%E7%B4%84 → /ja/約))
       unsafeExternalPathname = decodeURI(request.nextUrl.pathname);
     } catch {
-      // In case an invalid pathname is encountered, forward
-      // it to Next.js which in turn responds with a 400
-      return NextResponse.next();
+      // In case an invalid pathname is encountered, return a custom error message with detailed info alongside a 400 Status.
+
+      return new NextResponse(
+        `
+        <div style='display: flex; flex-direction: column; gap: 1rem;'>
+          <h1 style='font-size: 2.25rem; color: oklch(50.5% 0.213 27.518);'>Invalid locale</h1>
+          <p>Unsupported character detected; "${extractedLocale}" is not a valid locale, as it contains non URL-safe characters.</p>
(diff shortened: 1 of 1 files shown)


packages/next-intl/src/middleware/middleware.tsx
```diff
@@ -46,13 +46,35 @@ export default function createMiddleware<
...
-      return NextResponse.next();
+      // In case an invalid pathname is encountered, return a custom error message with detailed info alongside a 400 Status.
+
+      return new NextResponse(
+        `
+        <div style='display: flex; flex-direction: column; gap: 1rem;'>
+          <h1 style='font-size: 2.25rem; color: oklch(50.5% 0.213 27.518);'>Invalid locale</h1>
+          <p>Unsupported character detected; "${extractedLocale}" is not a valid locale, as it contains non URL-safe characters.</p>
+          <small style='font-family: "Lucida Console", "Courier New", monospace;'>This error was thrown by next-intl</small>
+        </div>
```
amannn, line 68: This hardcoded HTML is surely not something that we can return from the middleware. 

Separately, I think this opens up a XSS vulnerability:
1. `GET /%3Cimg src=x onerror=alert(1)%3E/%`
2. `extractedLocale` is literally ``, reflected into the HTML body

We forward with `NextResponse.next()` to Next.js so the error handling is delegated there.

If you want to add custom behavior, you can [compose the middleware](https://next-intl.dev/docs/routing/middleware#composing-other-middlewares).

PR #2406, opened 2026-09-10
fix: Keep `$` patterns in route param values literally

### LINKED ISSUE #2405: Route param values containing `$` patterns corrupt localized URLs (or throw)
### Description

When a route param value contains a `$` replacement pattern (`$&`, `$'`, `` $` ``, `$1`, …), the localized URL built by next-intl is corrupted — or building it throws outright.

### Reproduction

```ts
import {defineRouting} from 'next-intl/routing';

const routing = defineRouting({
  locales: ['en', 'de'],
  defaultLocale: 'en',
  pathnames: {'/about/[param]': {'en': '/about/[param]', 'de': '/ueber-uns/[param]'}}
});

const {getPathname} = createNavigation(routing);

// Throws: "Insufficient params provided for localized pathname.
// Template: /about/[param], Params: {\"param\":\"a$&b\"}"
getPathname({locale: 'en', href: {pathname: '/about/[param]', params: {param: 'a$&b'}}});
```

With other patterns (e.g. `$'`) it silently produces a wrong URL instead of throwing.

### Expected behaviour

Param values are inserted verbatim: `/about/a$&b`.

### Root cause

Param substitution uses `String.replace` with the param value as a *string* replacement, so `$` sequences are interpreted as replacement patterns:

- `packages/next-intl/src/navigation/shared/utils.tsx` (`compileLocalizedPathname`, affects `getPathname`, `Link`, `redirect`, `useRouter`)
- `packages/next-intl/src/middleware/utils.tsx` (`formatPathnameTemplate`, affects middleware localized-pathname redirects)

Passing a replacer function fixes it.

### Verifications

- [x] I've searched for similar, existing issues (found #607 and #2014, both about encoding — different from this).

soroush5:
Fixes #2405.

Route param values containing `$` replacement patterns (`$&`, `$'`, `` $` ``, `$1`) were interpreted by `String.replace`, corrupting localized URLs from the navigation APIs (`getPathname`, `Link`, `redirect`, `useRouter`) and from middleware localized-pathname redirects. With `$&` the placeholder was re-inserted and `compileLocalizedPathname` even threw `Insufficient params provided`.

Both substitution sites now pass a replacer function so values are inserted verbatim. Regression tests added in `navigation/shared/utils.test.tsx` and `middleware/utils.test.tsx`.

Verification: full `next-intl` suite 899/899 green, `tsc --noEmit`, ESLint and Prettier clean.

### DIFF
--- a/packages/next-intl/.size-limit.ts
+++ b/packages/next-intl/.size-limit.ts
@@ -27,7 +27,7 @@ const config: SizeLimitConfig = [
     name: "import {createNavigation} from 'next-intl/navigation' (react-server)",
     path: 'dist/esm/production/navigation.react-server.js',
     import: '{createNavigation}',
-    limit: '3.155 KB'
+    limit: '3.16 KB'
   },
   {
     name: "import * from 'next-intl/server' (react-client)",
--- a/packages/next-intl/src/middleware/utils.test.tsx
+++ b/packages/next-intl/src/middleware/utils.test.tsx
@@ -177,6 +177,15 @@ describe('formatPathname', () => {
       '/users/23/42'
     );
   });
+
+  it('keeps `$` patterns in parameter values literally', () => {
+    expect(formatPathnameTemplate('/users/[userId]', {userId: 'a$&b'})).toBe(
+      '/users/a$&b'
+    );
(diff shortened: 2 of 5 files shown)


amannn (approved):
Thanks a lot!

Issue #607, opened 2023-11-08
Symbols/foreign letters not working with localized pathnames


Entropy-10:
### Description

When using localized pathnames, if you currently try to use symbols or foreign letters it causes the url to generate incorrectly and cause a 404 not found error.

`navigation.ts`
```ts
export const pathnames = {
    '/': '/',
    '/test': {
        en: '/test',
        ja: '/テスト'
    }
} satisfies Pathnames<typeof locales>
```

### Mandatory reproduction URL (CodeSandbox or GitHub repository)

[![Edit next-intl-bug-template (app) (forked)](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/p/sandbox/next-intl-bug-template-app-forked-xskh6r?file=%2Fsrc%2Fnavigation.ts&embed=1)

### Reproduction description

Steps to reproduce:
1. Open reproduction
2. Click on `Go to Japanese Test Page`
3. See incorrect url and 404 not found error

### Expected behaviour

Expected: `http://localhost:3000/ja/テスト` 
Actual Result: `http://localhost:3000/ja/%E3%83%86%E3%82%B9%E3%83%88`.

amannn:
Thanks for the report!

Also the rewrite in the middleware doesn't seem to handle this URL correctly: https://xskh6r-3000.csb.app/ja/テスト

AdamZaczek:
Localized pathnames are super important for any platform relying on content. We've been trying to get around this, with no success. Looking forward to the fix!

amannn:
Contributions are welcome here in case someone wants to look into it.

I think this can be tackled as following:
1. Add a failing test in [the middleware test suite](https://github.com/amannn/next-intl/blob/main/packages/next-intl/test/middleware/middleware.test.tsx)
2. Adapt [the middleware](https://github.com/amannn/next-intl/blob/main/packages/next-intl/src/middleware/middleware.tsx) accordingly
3. Add a failing test in [the localized navigation test suite](https://github.com/amannn/next-intl/blob/main/packages/next-intl/test/navigation/createLocalizedPathnamesNavigation.test.tsx)
[...]

PR #2408, opened 2026-09-10
fix: Keep `[` and `]` in route param values literally

### LINKED ISSUE #2407: Route param values containing `[` or `]` are re-interpreted as template syntax
### Description

Route param values that contain `[` or `]` get re-interpreted as template syntax during param substitution. Depending on the value, this either throws a misleading error, silently produces a wrong URL, or drops the segment entirely.

Same family as #2405 (`$` patterns in param values, fixed in #2406), but a distinct root cause. #2406 stopped values from being read as *replacement patterns*; here the problem is that values are read back as *template placeholders* by the passes that run after substitution.

### Affected call sites

- `packages/next-intl/src/navigation/shared/utils.tsx` — `compileLocalizedPathname`, affecting `getPathname`, `Link`, `redirect` and `useRouter`
- `packages/next-intl/src/middleware/utils.tsx` — `formatPathnameTemplate`, affecting middleware redirects to localized pathnames

### Observed behaviour

With `pathnames: {'/about/[param]': …, '/x/[a]/[b]': …}`, on `main` (including #2406):

| Params | Actual | Expected |
| --- | --- | --- |
| `{param: 'a[b'}` | throws `Insufficient params provided for localized pathname` | value inserted verbatim |
| `{a: '[b]', b: 'B'}` | `/x/B/B` | `[b]` kept as the value of `a`, giving `/x/[b]/B` |
| `{param: '[[...x]]'}` | `/about` | value inserted verbatim, giving `/about/[[...x]]` |
| `{param: 'a]b'}` | `/about/a]b` | correct today, but only by luck: the guard happens to check `[` only |

Middleware is affected too, and there the params come straight from the incoming request path, so a plain request can reach it:

```ts
formatPathnameTemplate('/x/[a]/[b]', {a: '[b]', b: 'B'}); // '/x/B/[b]'
```

Whether `[` and `]` should additionally be percent-encoded in the output is the separate pre-existing encoding topic (#607, #2014). `new URL().pathname`, which `encodePathname` relies on, leaves both untouched today.

### Root cause

Substitution runs one param at a time against the running result, and the passes that follow treat leftover brackets as template syntax:

1. A value containing `[…]` becomes a substitution target for a *later* param in the same loop.
2. The optional-catch-all cleanup `compiled.replace(/\[\[\.\.\..+\]\]/g, '')` strips bracket text that came from a value.
3. The `compiled.includes('[')` guard then reads any surviving bracket as a missing param and throws.

A fix would make substitution single-pass over the template, so values are never re-scanned, and check for unresolved params against the template rather than against the compiled result.

### Related: duplicate params in middleware

Noticed while investigating. `formatPathnameTemplate` replaces without the `g` flag, so a template that uses the same param twice keeps the second placeholder:

```ts
formatPathnameTemplate('/x/[a]/y/[a]', {a: '1'}); // '/x/1/y/[a]'
```

`compileLocalizedPathname` handles this correctly and has a test for it (`compiles a pathname where a param appears twice`). Happy to split this out if it's better tracked separately.

### Verifications

- [x] I've searched for similar, existing issues (#2405 and #2406 cover `$` patterns, #607 and #2014 cover encoding).

soroush5:
Fixes #2407.

Route param values containing `[` or `]` were re-interpreted as template syntax during substitution, either throwing a misleading `Insufficient params` error, producing a wrong URL, or dropping the segment entirely:

| Params | Before | After |
| --- | --- | --- |
| `{param: 'a[b'}` | throws `Insufficient params…` | `/about/a[b` |
| `{one: '[two]', two: '2'}` | `/test/2/2` | `/test/[two]/2` |
| `{param: '[[...x]]'}` | `/about` | `/about/[[...x]]` |

**Root cause:** substitution ran one param at a time against the running result, so later passes treated leftover brackets as template syntax: values became substitution targets for later params, the optional-catch-all cleanup stripped bracket text that came from a value, and the `compiled.includes('[')` guard threw on surviving value brackets. In middleware, `formatPathnameTemplate` additionally replaced only the first occurrence, so a template using the same param twice kept the second placeholder (`/x/[a]/y/[a]` + `{a: '1'}` → `/x/1/y/[a]`).

**Fix:** both `compileLocalizedPathname` (`packages/next-intl/src/navigation/shared/utils.tsx`) and `formatPathnameTemplate` (`packages/next-intl/src/middleware/utils.tsx`) now substitute all placeholders in a single pass over the template via a replacer function, so inserted values are never re-scanned. Unresolved params are tracked against the template (missing required params still throw the same `Insufficient params` error in dev; unresolved optional catch-alls are still removed; extra params are still ignored). The replacer function also keeps `$` patterns in values literally.

**Relationship to #2406:** this subsumes the middleware half of #2406 (`() => value` replacer) and the same `$` class in `compileLocalizedPathname`, since single-pass function replacement handles both. Happy to rebase here or there — whichever you merge first, I will rebase the other.

**Verification:**
- New RED-first tests in `navigation/shared/utils.test.tsx` (4 failing cases from the issue table) and `middleware/utils.test.tsx` (bracket values + duplicate-param case) — all pass after the fix.
- Full `packages/next-intl` suite: 30 files / 900 tests green; `tsc --noEmit`, ESLint and Prettier pass on all touched files.
- Dup search: no prior issue/PR covers bracket re-interpretation (#607/#2014 are the separate encoding topic, as noted in the issue).

Note: CI flagged the size-limit budget (3 entries over by 3–18 B brotlied). Instead of bumping limits I slimmed the implementation (counter instead of tracking array, key captured without dots, `Object.hasOwn` in server-only middleware) — verified locally with `pnpm build && pnpm size`: all entries pass.

### DIFF
--- a/packages/next-intl/src/middleware/utils.test.tsx
+++ b/packages/next-intl/src/middleware/utils.test.tsx
@@ -195,11 +195,9 @@ describe('formatPathname', () => {
     expect(formatPathnameTemplate('/x/[a]/[b]', {a: '[b]', b: 'B'})).toBe(
       '/x/[b]/B'
     );
-  });
-
-  // https://github.com/amannn/next-intl/issues/2407
-  it('replaces a parameter that appears twice', () => {
-    expect(formatPathnameTemplate('/x/[a]/y/[a]', {a: '1'})).toBe('/x/1/y/1');
+    expect(
+      formatPathnameTemplate('/about/[param]', {param: '[[...x]]'})
+    ).toBe('/about/[[...x]]');
   });
 });
 
--- a/packages/next-intl/src/middleware/utils.test.tsx
+++ b/packages/next-intl/src/middleware/utils.test.tsx
@@ -186,6 +186,21 @@ describe('formatPathname', () => {
       "/users/$'`$1"
     );
   });
+
+  // https://github.com/amannn/next-intl/issues/2407
+  it('keeps `[` and `]` in parameter values literally', () => {
+    expect(formatPathnameTemplate('/users/[userId]', {userId: 'a[b'})).toBe(
+      '/users/a[b'
+    );
+    expect(formatPathnameTemplate('/x/[a]/[b]', {a: '[b]', b: 'B'})).toBe(
+      '/x/[b]/B'
+    );
+  });
+
+  // https://github.com/amannn/next-intl/issues/2407
+  it('replaces a parameter that appears twice', () => {
+    expect(formatPathnameTemplate('/x/[a]/y/[a]', {a: '1'})).toBe('/x/1/y/1');
+  });
 });
 
 describe('getInternalTemplate', () => {
--- a/packages/next-intl/src/middleware/utils.tsx
+++ b/packages/next-intl/src/middleware/utils.tsx
@@ -251,7 +251,7 @@ export function formatPathnameTemplate(template: string, params?: object) {
   // A replacer function is used so that `$` patterns in values
   // are kept literally.
   return template.replace(/\[([^\][]+)\]/g, (match, key) =>
-    Object.prototype.hasOwnProperty.call(params, key)
+    params && Object.hasOwn(params, key)
       ? String((params as Record<string, unknown>)[key])
       : match
   );
--- a/packages/next-intl/src/middleware/utils.tsx
+++ b/packages/next-intl/src/middleware/utils.tsx
@@ -251,7 +251,7 @@ export function formatPathnameTemplate(template: string, params?: object) {
   // A replacer function is used so that `$` patterns in values
   // are kept literally.
   return template.replace(/\[([^\][]+)\]/g, (match, key) =>
-    params && Object.hasOwn(params, key)
+    Object.hasOwn(params, key)
       ? String((params as Record<string, unknown>)[key])
       : match
   );
--- a/packages/next-intl/src/middleware/utils.tsx
+++ b/packages/next-intl/src/middleware/utils.tsx
@@ -245,12 +245,16 @@ export function formatPathnameTemplate(template: string, params?: object) {
   // we can replace the value with simple interpolation
   template = template.replace(/\[\[/g, '[').replace(/\]\]/g, ']');
 
-  let result = template;
-  Object.entries(params).forEach(([key, value]) => {
-    result = result.replace(`[${key}]`, () => value);
-  });
-
-  return result;
+  // Substitute all placeholders in a single pass so that values
+  // containing `[` or `]` are never re-interpreted as template
+  // syntax (see https://github.com/amannn/next-intl/issues/2407).
+  // A replacer function is used so that `$` patterns in values
+  // are kept literally.
+  return template.replace(/\[([^\][]+)\]/g, (match, key) =>
+    Object.prototype.hasOwnProperty.call(params, key)
+      ? String((params as Record<string, unknown>)[key])
+      : match
+  );
 }
 
 export function formatPathname(
--- a/packages/next-intl/src/navigation/shared/utils.test.tsx
+++ b/packages/next-intl/src/navigation/shared/utils.test.tsx
@@ -136,7 +136,37 @@ describe('compileLocalizedPathname', () => {
         pathnames
       })
     ).toBe('/about/a$&b');
+  });
 
+  // https://github.com/amannn/next-intl/issues/2407
+  it('keeps `[` and `]` in param values literally', () => {
+    expect(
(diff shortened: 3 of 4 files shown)


amannn:
Thanks for picking this up, I discovered this coincidentally during the review of your other PR.

Can you rebase on main?

soroush5:
Quick heads-up: the `main` workflow doesn't seem to trigger for my latest pushes (`7cba2f2` and `ef3f06e`) — only `lint-pr` ran for both. The 07:33 run on `eb6a93c` worked fine, and I verified locally with `pnpm build && pnpm size` that all size entries pass after the slim-down.

Not sure if it's an approval gate or something on my side — could you take a look when you get a chance? Happy to rebase or adjust whatever's needed.

amannn:
Yes, workflows need to be manually triggered by a maintainer.

Can you take care of the conflicts?

image

Then we can trigger workflows again.

soroush5:
Conflicts are resolved — rebased onto `v4.14.3`. I kept the merged `$` handling from #2406 and folded the bracket fix on top (single-pass substitution covers both, as noted in the description), and kept both sides' tests: 46 passed across the two suites, plus `tsc --noEmit` and `pnpm size` clean locally.

Ready whenever you want to trigger the workflows again.

soroush5:
The 3 remaining failures are all Vercel preview deployments asking for team authorization ("A member of the Team first needs to authorize it"). As an external contributor I can't authorize deploys to the next-intl Vercel team — that click has to come from your side. They're preview-only, so feel free to ignore them if you prefer.

packages/next-intl/src/middleware/utils.test.tsx
```diff
@@ -186,6 +186,21 @@ describe('formatPathname', () => {
...
+      '/users/a[b'
+    );
+    expect(formatPathnameTemplate('/x/[a]/[b]', {a: '[b]', b: 'B'})).toBe(
+      '/x/[b]/B'
+    );
+  });
+
+  // https://github.com/amannn/next-intl/issues/2407
+  it('replaces a parameter that appears twice', () => {
+    expect(formatPathnameTemplate('/x/[a]/y/[a]', {a: '1'})).toBe('/x/1/y/1');
```
amannn: Next.js doesn't support this:

> Error: You cannot have the same slug name "a" repeat within a single dynamic path

packages/next-intl/src/middleware/utils.test.tsx
```diff
@@ -186,6 +186,21 @@ describe('formatPathname', () => {
...
     );
   });
+
+  // https://github.com/amannn/next-intl/issues/2407
+  it('keeps `[` and `]` in parameter values literally', () => {
+    expect(formatPathnameTemplate('/users/[userId]', {userId: 'a[b'})).toBe(
+      '/users/a[b'
+    );
+    expect(formatPathnameTemplate('/x/[a]/[b]', {a: '[b]', b: 'B'})).toBe(
+      '/x/[b]/B'
```
amannn, line 196: Also this would be a problem if you pass it to a Next.js' link, right?

Did you encounter this in a real app or is this a theoretical fix?

packages/next-intl/src/middleware/utils.test.tsx
soroush5: You're right, removed. A route with the same slug twice is rejected by Next.js itself, so pinning behavior for it in this PR was wrong. Per your note in #2407 I left that middleware sub-case out entirely — happy to split it into a separate issue if you want it tracked.

packages/next-intl/src/middleware/utils.test.tsx
soroush5: Fair question. That expectation comes straight from your table in #2407 (row 2: "`[b]` kept as the value of `a`, giving `/x/[b]/B`") — I didn't invent the shape. Whether `[` should additionally be percent-encoded in the output is the #607/#2014 topic you carved out there, so I left that alone.

On real vs theoretical: I found this by code inspection while working through the #2405 family, not from a production app. But the middleware path takes params straight from the incoming request path (your own words in #2407: "a plain request can reach it"), so any URL carrying brackets — user-generated slugs, array-style segments — can hit it. The navigation path needs a bracket-containing value, which I'll admit is rarer. I also added your row-3 case (`[[...x]]` as a value) to the test since the old cleanup pass would have eaten it.

amannn:
Ok, since this is purely theoretical then, I'll close this for now. If we discover this in a real app, let's open an issue first.

soroush5:
Sounds good, thanks for reviewing it. If I run into this in a real app I'll open an issue first.

PR #2409, opened 2026-09-10
fix: Strip locale prefix on cross-domain redirects to unprefixed domains

### LINKED ISSUE #2369: Cross-domain redirect keeps locale prefix when target domain's defaultLocale is unprefixed (as-needed)
### Description

With `domains` configured and `localePrefix: { mode: 'as-needed', prefixes: {...} }`, a cross-domain redirect keeps the locale prefix even when the target domain's `defaultLocale` is that domain's un-prefixed root. The resulting URL 404s.

`redirect()` in `middleware.js` already contains exactly the normalization needed, but it is gated behind `!redirectDomain` — and the cross-domain call site always passes a domain, so the normalization never runs for the case it appears to be written for.

**`redirect()`** — normalization is inside `if (... && !redirectDomain && domain)`:

```js
function redirect(url, redirectDomain) {
  const urlObj = new URL(url, request.url);
  urlObj.pathname = normalizeTrailingSlash(urlObj.pathname);
  if (domainsConfig.length > 0 && !redirectDomain && domain) {   // <-- gate
    const bestMatchingDomain = getBestMatchingDomain(domain, locale, domainsConfig);
    if (bestMatchingDomain) {
      redirectDomain = bestMatchingDomain.domain;
      const redirectDomainMode = bestMatchingDomain.localePrefix || resolvedRouting.localePrefix.mode;
      if (bestMatchingDomain.defaultLocale === locale && redirectDomainMode === 'as-needed') {
        urlObj.pathname = getNormalizedPathname(...);             // <-- the fix we want
      }
    }
  }
```

**The cross-domain call site** passes `pathDomain?.domain`, so the gate is always false there:

```js
const pathDomain = getBestMatchingDomain(
  domain,
  pathnameMatch.locale,
  domainsConfig,
);
if (domain?.domain !== pathDomain?.domain && !hasUnknownHost) {
  response = redirect(externalHref, pathDomain?.domain); // <-- skips normalization
}
```

Because `getLocalePrefix` is `localePrefix.prefixes?.[locale] || '/' + locale`, a locale intentionally omitted from `prefixes` (i.e. one meant to be served unprefixed at its domain root) still gets `/{locale}` synthesized into `externalHref`. There is no way to declare a prefix-less locale via config — `''` is falsy — so this cannot be worked around in userland routing config.

### Reproduction

Two hosts are required, so this needs `/etc/hosts` entries rather than a sandbox URL:

```
127.0.0.1 example-eu.localhost example-de.localhost
```

```ts
  locales: ["en", "nl", "de"],
  defaultLocale: "en",
  // `nl` is served at a sub-path; `en` and `de` are their domains' roots,
  // so they are deliberately omitted from `prefixes`.
  localePrefix: { mode: "as-needed", prefixes: { nl: "/nl" } },
  domains: [
    {
      domain: "example-eu.localhost:3000",
      defaultLocale: "en",
      locales: ["en", "nl"],
    },
    {
      domain: "example-de.localhost:3000",
      defaultLocale: "de",
      locales: ["de"],
    },
  ],
});
```

```bash
curl -sI 'http://example-eu.localhost:3000/de' | grep -i '^location'
```

### Expected behaviour

`de` is `example-de.localhost:3000`'s `defaultLocale` and the mode is `as-needed`, so it is served unprefixed at that domain's root:

```
location: http://example-de.localhost:3000/
```

### Actual behaviour

The prefix from the originating domain is carried over, producing a URL that 404s (nothing is served at `/de` on that host, since `de` is unprefixed there):

```
location: http://example-de.localhost:3000/de
```

Same for nested paths: `/de/contact` → `example-de.localhost:3000/de/contact` instead of `example-de.localhost:3000/contact`.

### Suggested fix

Resolve the target domain config even when the caller supplied `redirectDomain`, so the existing `defaultLocale === locale && mode === 'as-needed'` normalization applies in both paths — e.g.:

```js
if (domainsConfig.length > 0 && domain) {
  const bestMatchingDomain = redirectDomain
    ? domainsConfig.find((cur) => cur.domain === redirectDomain)
    : getBestMatchingDomain(domain, locale, domainsConfig);
  ...
```

### Version

`next-intl@4.12.0`; verified still present in `4.13.4` (current latest) — the gate and call site are unchanged there.

soroush5:
Fixes #2369.

With `domains` and `localePrefix: 'as-needed'`, requesting e.g. `/de/about` on a domain that doesn't serve `de` redirects to the right domain but keeps the prefix (`de.example.com/de/about`), even though `de` is that domain's unprefixed default locale. That URL redirects a second time to `/about` (or 404s for clients that don't follow redirects).

The `redirect()` helper already strips the prefix in this situation, but only when it picks the domain itself (`!redirectDomain`). The cross-domain call site always passes a domain, so the normalization never ran there. This change looks up the explicitly passed domain and applies the same normalization.

Repro (middleware, no hosts setup needed):

```ts
createMiddleware({
  defaultLocale: 'en',
  locales: ['en', 'nl', 'de'],
  localePrefix: {mode: 'as-needed', prefixes: {nl: '/nl'}},
  domains: [
    {defaultLocale: 'en', domain: 'eu.example.com', locales: ['en', 'nl']},
    {defaultLocale: 'de', domain: 'de.example.com', locales: ['de']}
  ]
});
// GET http://eu.example.com/de/about
// before: Location: http://de.example.com/de/about
// after:  Location: http://de.example.com/about
```

Verification: new test in `middleware.test.tsx` failed before (`expected 'http://de.example.com/de/about' to be 'http://de.example.com/about'`) and passes after. Three existing cross-domain expectations that asserted the prefixed intermediate URL now assert the canonical URL. Full `next-intl` suite: 900/900 pass; eslint and prettier clean on changed files.

Note: CI flagged the middleware bundle 40 B over budget, so I slimmed the normalization (inlined one single-use const, no behavior change) — verified locally with `pnpm build && pnpm size`: entry back at exactly the 10.12 kB limit.


amannn:
Hey, did you see my comment at https://github.com/amannn/next-intl/issues/2369#issuecomment-5203614014? If you found a way to reproduce this, you can add a PR that targets https://github.com/amannn/next-intl/pull/2377 with the minimal changes to get this failing. Let's wait with a fix until we have a clear understanding of the issue.

soroush5:
Thanks — and yes, I'd missed your #2369 comment, sorry about that. To answer it directly: I never saw a real 404 either. My repro was at the middleware unit level (the `Location` header carries the prefix over to the target domain); your e2e chain confirms the intermediate URL just 307s again instead of 404ing. So the honest framing is your "against" case: nothing broken, it's a round-trip optimization where the normalization already exists but never runs for the cross-domain path.

Done as you suggested: #2411 targets #2377 with only the hop assertions flipped to single-redirect (verified locally: 6/6 baseline green, exactly the three cross-domain tests fail with the change, nothing else touched). Closing this one — the fix branch is kept and I'm happy to rebase it wherever you want if you decide to go ahead.

PR #2410, opened 2026-09-11
fix: Improve base path detection for setting a cookie

soroush5:
I ran into this with a base path that contains the current pathname as a substring.

Repro: app mounted under `/dashboard`, user on `/dashboard/dash`, so `getBasePath('/dash', '/dashboard/dash')` runs. The old code did a first-occurrence replace and returned `/board/dash`, which is garbage and breaks navigation under that base path.

This change only strips the pathname when it's actually the trailing part of the window path, and leaves the window path untouched otherwise. I added cases for `/dash` under `/dashboard` and `/doc` under `/docs`.

Tests: the `getBasePath` suite in `utils.test.tsx` passes (18/18).

### DIFF
--- a/packages/next-intl/src/navigation/shared/utils.tsx
+++ b/packages/next-intl/src/navigation/shared/utils.tsx
@@ -250,8 +250,13 @@ export function getBasePath(
 ) {
   if (pathname === '/') {
     return windowPathname;
+  } else if (windowPathname.endsWith(pathname)) {
+    // Only strip the trailing occurrence: the pathname can also
+    // appear inside the base path itself (e.g. pathname `/dash`
+    // with a base path of `/dashboard`).
+    return windowPathname.slice(0, -pathname.length);
   } else {
-    return windowPathname.replace(pathname, '');
+    return windowPathname;
   }
 }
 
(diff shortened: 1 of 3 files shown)


packages/next-intl/src/navigation/shared/utils.tsx
```diff
@@ -250,8 +250,17 @@ export function getBasePath(
 ) {
   if (pathname === '/') {
     return windowPathname;
+  } else if (windowPathname.endsWith(pathname)) {
+    // Only strip the trailing occurrence: the pathname can also
+    // appear inside the base path itself (e.g. pathname `/dash`
+    // with a base path of `/dashboard`).
```
amannn:
```suggestion
    // Only strip the trailing occurrence: the pathname can also
    // appear inside the base path itself (e.g. `/dashboard/dash`)
```

packages/next-intl/src/navigation/shared/utils.tsx
```diff
@@ -250,8 +250,17 @@ export function getBasePath(
...
+    return windowPathname.slice(0, -pathname.length);
   } else {
-    return windowPathname.replace(pathname, '');
+    // Unexpected, since the window pathname should always end with
+    // the pathname. Assuming no base path is safer than returning
+    // the window pathname, as the latter would scope the locale
+    // cookie to the current page instead of the app root.
```
amannn:
```suggestion
    // Unexpected, since the window pathname should always end with
    // the pathname. Assuming no base path is safe though.
```

amannn:
Added some minor changes, looks good to me now.

PR #2413, opened 2026-09-11
fix(navigation): read the pathname in useRouter only when the locale cookie is enabled

### LINKED ISSUE #2412: useRouter subscribes to usePathname on every render and blocks route shells under Cache Components
### Description

`createNavigation(...).useRouter()` calls `usePathname()` at hook level. The value is only used to pass the current pathname into `syncLocaleCookie` inside `push`/`replace`/`prefetch`.

With Next.js 16 Cache Components that read blocks the route's App Shell. Any shared component that calls `useRouter` — a header, a bottom nav, a link wrapper — turns into `Next.js encountered URL data 'usePathname()' in a Client Component outside of <Suspense>` while the shell for a route with dynamic params is generated, so the whole shell is lost. It happens even with `localeCookie: false`, where the pathname is never needed.

### Suggested fix

Read the pathname where it is used instead of on every render: inside the returned `push`/`replace`/`prefetch` callbacks. They run in the browser, so `window.location.pathname` gives the same value the cookie sync needs. Skipping the read entirely when `localeCookie` is disabled would cover the common case too.

### Current workaround

Our own `useRouter` over `next/navigation` plus the exported `getPathname` for localization — same behaviour, no pathname subscription.

### Environment

next-intl 4.14.4, next 16.3.4, `localePrefix: 'as-needed'`, `localeCookie: false`, `cacheComponents: true`.

AdzerKI:
Closes #2412.

`useRouter` called `usePathname()` on every render, although the value is only passed to `syncLocaleCookie`. With Next.js Cache Components that read blocks the App Shell of every route with dynamic params, so any shared component using the router — header, bottom nav, link wrapper — takes the shell down with it, even when `localeCookie: false`.

The hook is now picked once per navigation instance: with a locale cookie it stays `usePathname`, without one it returns `null` and nothing subscribes to URL data. Hook order stays stable because the config is fixed when `createNavigation` runs.

Apps with a locale cookie still read the pathname — the base path is derived from it, so dropping the read there needs a different source.

`pnpm --filter next-intl vitest run src/navigation/react-client` — 93 passed, including the added case.

### DIFF
--- a/packages/next-intl/src/navigation/react-client/createNavigation.test.tsx
+++ b/packages/next-intl/src/navigation/react-client/createNavigation.test.tsx
@@ -313,6 +313,28 @@ describe("localePrefix: 'always', with `localeCookie`", () => {
   });
 });
 
+describe("localePrefix: 'always', with `localeCookie: false`", () => {
+  const {useRouter} = createNavigation({
+    locales,
+    defaultLocale,
+    localePrefix: 'always',
+    localeCookie: false
+  });
+
+  describe('useRouter', () => {
+    const invokeRouter = getInvokeRouter(useRouter);
+
+    it('navigates without reading the pathname', () => {
+      vi.mocked(useNextPathname).mockClear();
+
+      invokeRouter((router) => router.push('/about', {locale: 'de'}));
+
+      expect(useNextRouter().push).toHaveBeenCalledWith('/de/about');
+      expect(useNextPathname).not.toHaveBeenCalled();
+    });
+  });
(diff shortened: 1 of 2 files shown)


amannn:
Hey, thanks for opening this PR @AdzerKI! I just dug into this and it seems like this can be solved in a more generic way like this: https://github.com/amannn/next-intl/pull/2415

I'm currently finishing that PR and should have a release out soon.

Thanks again!

PR #2415, opened 2026-09-14
fix: Avoid reading the pathname in `useRouter` and `Link` to work better with Cache Components

### LINKED ISSUE #2412: useRouter subscribes to usePathname on every render and blocks route shells under Cache Components

amannn:
Fixes #2412. Alternative to #2413.

`useRouter` called `usePathname()` on every render and `Link` read it whenever the `locale` prop was set. With Cache Components, reading URL data blocks the App Shell of routes with dynamic params, so any shared component using the router (header, bottom nav, link wrapper) takes the shell down with it.

The pathname was only needed to _detect_ a `basePath`: `syncLocaleCookie` subtracted the Next.js pathname from `window.location.pathname` to derive the cookie `path`. Instead, the base path is now forwarded from `next.config.ts` by the plugin via the `env` setting—the same mechanism that's already used for `trailingSlash` (#1190).

Apart from removing the pathname read, this also matches the value that the middleware uses when it sets the cookie (`request.nextUrl.basePath`), rather than inferring it from the browser location. The heuristic could disagree with the middleware—e.g. behind a proxy prefix that isn't part of `basePath`—which resulted in two `NEXT_LOCALE` cookies on different paths.

## Changes

- `plugin`: Forward `basePath` as `_next_intl_base_path`, alongside `_next_intl_trailing_slash` (no `undefined` values, since Turbopack rejects those).
- `syncLocaleCookie`: Take the base path from that value instead of a `pathname` argument.
- `useRouter`: No `usePathname()` call, and `nextPathname` is gone from the `useMemo` deps.
- `BaseLink`: No `usePathname()` call in `LocaleChangingLink`—locale-switching links no longer require a `<Suspense>` boundary either.
- `getBasePath`: The string-subtraction heuristic (incl. the `/dashboard/dash` special case from #2410) is replaced by reading the forwarded value.
- `.size-limit.ts`: `createNavigation` shrinks by ~50 B in both environments.

## Behavior change

A base path is now only considered for the cookie `path` if the `next-intl` plugin is set up, since that's what forwards the setting. Without it, the cookie falls back to `/`. This matches how `trailingSlash` already behaves; I've added a note to the `next.config.ts` section of the docs.

## Tests

- Unit: `does not read the pathname` regression test for `useRouter`, updated `getBasePath` tests, base-path suites now provide the value the way the plugin does. 576 tests pass (5 files fail to resolve `next-intl/config` in this environment, on `main` as well).
- E2E (playground, Next.js 16.3.1 with Turbopack): `base-path` (4 passed, incl. a new case covering a locale switch via `router.replace`), `main` (72), `trailing-slash` (7), `locale-cookie-false` (1). Verified in the built client bundle that the value is inlined by Turbopack: `function P(){try{return"/base/path"}catch{return""}}`.

https://claude.ai/code/session_01VVhgSCEcmfASkiyTqcvTY5

---
_Generated by [Claude Code](https://claude.ai/code/session_01VVhgSCEcmfASkiyTqcvTY5)_


docs/src/pages/docs/routing/configuration.mdx
```diff
@@ -551,7 +551,7 @@ export default async function middleware(request: NextRequest) {
 
 ## `next.config.ts` [#next-config]
 
-Apart from your routing configuration, `next-intl` will also incorporate settings from [`next.config.ts`](https://nextjs.org/docs/pages/api-reference/config/next-config-js).
+Apart from your routing configuration, `next-intl` will also incorporate settings from [`next.config.ts`](https://nextjs.org/docs/pages/api-reference/config/next-config-js). Note that these settings are forwarded by the [`next-intl` plugin](/docs/getting-started/app-router#next-config), therefore make sure it's set up in your app.
```
amannn:
```suggestion
Apart from your routing configuration, `next-intl` will also incorporate settings from [`next.config.ts`](https://nextjs.org/docs/pages/api-reference/config/next-config-js).
```

packages/next-intl/src/navigation/react-client/createNavigation.test.tsx
```diff
@@ -310,6 +310,16 @@ describe("localePrefix: 'always', with `localeCookie`", () => {
...
     });
+
+    it('does not read the pathname', () => {
+      // Reading URL data would require a `<Suspense>` boundary
+      // in the calling component when Cache Components are used
+      vi.mocked(useNextPathname).mockClear();
+
+      invokeRouter((router) => router.push('/about', {locale: 'de'}));
+
+      expect(useNextPathname).not.toHaveBeenCalled();
```
amannn:
```suggestion
      expect(useNextPathname).not.toHaveBeenCalled();
```

packages/next-intl/src/plugin/getNextConfig.tsx
```diff
@@ -349,11 +349,16 @@ export default function getNextConfig(
   }
 
   // Forward config
+  // (note that `undefined` values are not supported by Turbopack)
```
amannn: move this comment to the object.keys section below (without brackets)

packages/next-intl/src/navigation/shared/utils.tsx
```diff
@@ -244,19 +244,11 @@ export function getRoute<AppLocales extends Locales>(
...
-    // Only strip the trailing occurrence: the pathname can also
-    // appear inside the base path itself (e.g. `/dashboard/dash`)
-    return windowPathname.slice(0, -pathname.length);
-  } else {
-    // Unexpected, since the window pathname should always end with
-    // the pathname. Assuming no base path is safe though.
+export function getBasePath() {
+  try {
+    // Provided via `env` setting in `next.config.js` via the plugin
+    return process.env._next_intl_base_path || '';
```
amannn, line 250: is reading from process.env safe here? do we already have this pattern in the codebase or could it break consumers? 

answer the question, dont change anything

amannn: Yes — and it's the same pattern we already use for `trailingSlash`.

**Existing precedent:** `hasTrailingSlash()` in `src/shared/utils.tsx` is identical, down to the `try`/`catch`:

```ts
function hasTrailingSlash() {
  try {
    // Provided via `env` setting in `next.config.js` via the plugin
    return process.env._next_intl_trailing_slash === 'true';
  } catch {
    return false;
  }
}
```

That's been shipping in client bundles since #1190 (Jul 2024), and it's reached from `normalizeTrailingSlash`, which the navigation APIs call on every `getPathname`.

**Why it can't break consumers.** I checked the actual playground output (Next.js 16.3.1, Turbopack). Two cases:

With `basePath: '/base/path'` configured, the plugin adds the `env` entry and the value is inlined at build time:

```js
function P(){try{return"/base/path"}catch{return""}}
```

Without a `basePath` the plugin adds no entry, so nothing is replaced — and the read still resolves, because Turbopack rewrites `process.env` to its `process` shim module:

```js
function P(){try{return o.default.env._next_intl_base_path||""}catch{return""}}
```

where that module is:

```js
t.exports = (e.g.process?.env && typeof e.g.process?.env === 'object')
  ? e.g.process
  : e.r(30239) // polyfill
```

So an unknown key is just `undefined` → `''`, no `ReferenceError`. Webpack behaves the same way (its own `process/browser` shim). The `try`/`catch` is belt-and-braces for exotic bundlers that neither define nor shim `process` — same reason it's there in `hasTrailingSlash`.

The one real consequence is the one in the PR description: without the plugin the entry is never written, so the cookie `path` falls back to `/` instead of the base path. Identical to how `trailingSlash` already behaves without the plugin.

---

amannn:
@ceolinwill This is closely related to https://github.com/amannn/next-intl/pull/2362. I noticed we can address this in a general way that also works for locale-switching links (in case you had to add suspense boundaries for these).

ceolinwill:
@amannn this seems a good direction to go. Thanks for letting me know

PR #2417, opened 2026-09-16
fix: Preserve brackets in localized route params

### LINKED ISSUE #2407: Route param values containing `[` or `]` are re-interpreted as template syntax

tiffanychou0316:
Fixes #2407.

## What changed

- Replace pathname parameters in a single pass so inserted values are never interpreted as template syntax.
- Share the substitution logic between navigation and middleware.
- Preserve literal brackets in regular and catch-all parameter values.
- Replace duplicate middleware placeholders consistently.
- Keep the existing optional catch-all and unresolved-parameter behavior.

## Verification

- Added regression coverage for bracket-containing values and cross-parameter substitution.
- Added middleware coverage for duplicate and catch-all placeholders.
- 43 targeted tests pass.
- ESLint and Prettier pass.
- `next-intl`, `use-intl`, and `icu-minify` build successfully.
- TypeScript `--noEmit` passes.


amannn:
The issue is closed since it's purely theoretical: https://github.com/amannn/next-intl/issues/2407#issuecomment-5696570522

PR #2418, opened 2026-09-19
test: await locale cookie checks in base-path navigation

Yaqoubj:
The base-path locale-switching test calls the async `assertLocaleCookieValue` helper without awaiting it after switching to German and back to English. This lets navigation continue, or the test callback finish, before the browser cookie read and assertions complete.

Await both calls so their assertions remain part of the test's awaited control flow.

Validation on Windows with Node 24.13.0:
- Built `icu-minify`, `use-intl`, and `next-intl` from source.
- Built the playground in its `base-path` configuration, including TypeScript checks.
- All 4 tests in `base-path.spec.ts` passed in Chromium.
- All 2 playground Jest tests passed.
- ESLint and Prettier passed for the changed file.
- A local controlled-delay experiment using the original test callback and helper showed that the original callback completes before pending cookie checks; the patched callback waits. Browser navigation is stubbed in this experiment, so it supplements the real browser tests above.


amannn:
Great point, thanks!

